PR #323, opened 2026-03-22
fix(runtime): cap stream lines buffer to prevent unbounded growth

slepp:
## Why

The `LinesStream` adapter (`hew_stream_lines`) buffered upstream data without any size limit. If upstream never sent a newline — binary data, a malicious sender, or a misconfigured stream — the buffer grew unboundedly until memory was exhausted.

## What

Added a `MAX_LINE_BUFFER_SIZE` constant (1 MiB). When the buffer reaches this limit without finding a newline, the adapter drains exactly `MAX_LINE_BUFFER_SIZE` bytes as a truncated line and keeps leftover bytes for the next call. This is consistent with the existing flush-on-EOF behaviour.

A `skip_next_delimiter` flag ensures the trailing delimiter (`\n` or `\r\n`) for the oversized line is consumed — even when split across chunk boundaries — so no spurious empty line is yielded.

## Tests

- Normal line splitting, CRLF handling, and final-line-without-newline still work (existing tests)
- Data exactly at the limit with a trailing newline yields one complete line (no premature truncation)
- Chunked delivery past the limit yields a truncated line at exactly the limit, with leftovers in the next item
- A single oversized chunk (larger than the limit) is capped at the limit, not returned whole
- `\n` immediately after a forced flush is consumed (no spurious empty line)
- `\r\n` immediately after a forced flush is consumed
- `\r\n` split across the flush boundary is handled correctly
- Data after the delimiter continues to split normally

Fixes #22

### DIFF
--- a/hew-runtime/src/stream.rs
+++ b/hew-runtime/src/stream.rs
@@ -262,6 +262,11 @@ impl SinkBacking for FileWriteSink {
 
 // ── Lines adapter backing ─────────────────────────────────────────────────────
 
+/// Maximum bytes the lines adapter will buffer before yielding a truncated
+/// line.  Prevents unbounded memory growth when the upstream never sends a
+/// newline (e.g. binary data or a malicious sender).
+const MAX_LINE_BUFFER_SIZE: usize = 1024 * 1024; // 1 MiB
+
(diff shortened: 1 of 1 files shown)


slepp:
Consolidated into #328 (fix/stream-buffering-hardening)

PR #625, opened 2026-04-04
fix(observe): replace unwrap() with if-let/map_or in refresh() trace path

slepp:
## Problem

`refresh()` in `hew-observe/src/app.rs` contained two bare `.unwrap()` calls that assumed `self.cluster` was always `Some` when the Messages or Timeline tab was active:

```rust
// line ~522
let cluster_ref = self.cluster.as_mut().unwrap();

// line ~538
let cluster_status = self.cluster.as_ref().unwrap().status();
```

These were only protected by an early `return` at line 465. Because `try_rediscover()` runs earlier in the same function and can replace (or, in future refactors, clear) `self.cluster` during a disconnect/reconnect transition, any code path that reaches lines 522/538 with `cluster = None` would **panic the TUI** instead of staying alive in waiting/reconnect mode.

## Fix

Two surgical changes in `hew-observe/src/app.rs`:

1. **Trace-fetch block** — wrap in `if let Some(cluster_ref) = self.cluster.as_mut()` so a `None` cluster yields empty traces rather than panicking.
2. **Status line** — use `.map_or(ConnectionStatus::Disconnected, ClusterClient::status)` so a missing cluster falls back to `Disconnected` rather than panicking.

No behaviour change in normal connected flows.

## Tests added (3 new → 8 total)

| Test | What it covers |
|---|---|
| `refresh_no_cluster_does_not_panic` | All 7 tabs with `cluster = None` complete without panic |
| `refresh_cluster_cleared_messages_tab_does_not_panic` | Messages tab, `cluster` explicitly set to `None` |
| `refresh_cluster_cleared_timeline_tab_does_not_panic` | Timeline tab, `cluster` explicitly set to `None` |

## Validation

```
cargo test -p hew-observe
test result: ok. 8 passed; 0 failed
```

Pre-commit hooks (clippy + rustfmt) clean.

### DIFF
--- a/hew-observe/src/app.rs
+++ b/hew-observe/src/app.rs
@@ -519,23 +519,29 @@ impl App {
         let traces = if (self.active_tab == Tab::Messages && !self.trace_paused)
             || self.active_tab == Tab::Timeline
         {
-            let cluster_ref = self.cluster.as_mut().unwrap();
-            let mut all_traces = Vec::new();
-            for node in &mut cluster_ref.nodes {
-                if let Some(mut t) = node.client.fetch_traces() {
-                    all_traces.append(&mut t);
+            if let Some(cluster_ref) = self.cluster.as_mut() {
+                let mut all_traces = Vec::new();
+                for node in &mut cluster_ref.nodes {
+                    if let Some(mut t) = node.client.fetch_traces() {
+                        all_traces.append(&mut t);
+                    }
                 }
-            }
-            if all_traces.is_empty() {
-                None
+                if all_traces.is_empty() {
+                    None
+                } else {
+                    Some(all_traces)
+                }
             } else {
-                Some(all_traces)
+                None
             }
         } else {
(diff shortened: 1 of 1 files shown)


slepp:
Closing after overlap review — the review is correct.

**There is no real panic path.** Exact control flow on current main:

- `try_rediscover()` (line 462) has a single write to `self.cluster` in the entire codebase (line 625), always `Some(...)`, never `None`.
- The `let Some(cluster) = &mut self.cluster else { return }` guard at line 465 makes lines 522 and 538 unreachable when `cluster` is `None`.
- Nothing between line 465 and lines 522/538 modifies `self.cluster`.

The added `if let`/`map_or` guards are dead code. The three new tests all return at line 465 and never execute the guarded lines, so they provide no coverage of the new code.

The earlier overlap review overestimated the risk. Closing rather than merging misleading history.

PR #667, opened 2026-04-05
fix(types,lsp): unify builtin method resolution

slepp:
## Summary
- centralize builtin and named method resolution in `hew-types`
- reuse the shared resolver in analysis/LSP completions, hover, and signature help
- fix imported builtin handle precedence, fallback receiver type substitution, and invalid method-call signature help fallback

## Validation
- `cargo test -p hew-types --quiet`
- `cargo test -p hew-analysis --quiet`
- `cargo test -p hew-lsp --quiet`
- `cargo test -p hew-types --test e2e_typecheck --quiet`


slepp:
Superseded by #669, which carries the repaired resolution follow-on branch after the late signature-help blocker fix.

PR #948, opened 2026-04-11
fix(types): validate method_call_receiver_kinds output contract

slepp:
## Summary
- add the missing checker-side output-contract validator for `method_call_receiver_kinds`
- prune orphaned, `Ty::Error`, and unresolved receiver-kind entries after expr output pruning
- add focused hew-types regressions for the fail-closed prune paths

## Validation
- cargo test -p hew-types


slepp:
Closing as superseded: the local restack onto current origin/main produced an empty diff because the method_call_receiver_kinds output-contract implementation is already present upstream. The only conflict in `hew-types/src/check/admissibility.rs` resolved entirely to the upstream version, and `cargo test -p hew-types` passed on the restacked state.

PR #985, opened 2026-04-11
fix(codegen): use out-ptr ABI for hew_vec_pop_generic in VecPopOpLowering

slepp:
## Bug

`VecPopOpLowering::matchAndRewrite` had no `_generic` branch. When the element type was a struct or tuple (`vecElemSuffix` returns `"_generic"`), the lowering fell through to the primitive path and generated:

```
hew_vec_pop_generic(vec_ptr) -> <struct_type>  // WRONG: direct return
```

But the actual runtime signature is an out-pointer ABI:

```c
hew_vec_pop_generic(v: *mut HewVec, out: *mut c_void) -> i32  // i32 = 1 (popped) or 0 (empty)
```

This ABI mismatch would produce incorrect MLIR IR for any `Vec<StructType>.pop()` call.

## Fix

Add the `_generic` branch to `VecPopOpLowering`, following the same pattern already used in `VecPushOpLowering` and `VecGetOpLowering`:

1. Allocate a stack slot (`llvm.alloca`) of the result struct type
2. Declare `hew_vec_pop_generic` with signature `(ptr, ptr) -> i32`
3. Call it, passing the vec pointer and the alloca as arguments
4. Load the struct back from the alloca to produce the SSA result

The i32 return (success/empty flag) is correctly captured but not used — callers that need empty-checking use `vec.is_empty()` before popping.

## Test

Adds `vec_struct_pop.hew` — a `Vec<Point>` push/pop e2e regression test exercising the `_generic` lowering path end-to-end through code generation and execution.


slepp:
Superseded by #1037, which landed the same core generic Vec.pop fail-closed repair on main while preserving the raw FFI contract. The only remaining value here is extracting the residual tuple-pop/runtime-unit coverage into a fresh bounded follow-up.

PR #993, opened 2026-04-13
fix(types): non-root module body typecheck parity

slepp:
## Summary

Brings non-root module bodies to full typecheck parity with the root module. All imported module bodies are now typechecked with correct pre-registration ordering, privacy/call-gating, and local-type scoping — matching the guarantees already in place for the root module.

### Core changes (hew-types)
- Pre-register non-root module types and scope `local_type_defs` for body checking
- Typecheck non-root `module_graph` bodies
- Refresh non-root module parity in registration pass
- Resolve short handle method lookups after preregistration
- Cover private non-root local types with tests
- Keep private module functions uncallable (call-gating)
- Preserve type static methods in call-gating
- Registration ordering fixes

### Supporting changes
- **hew-serialize/enrich.rs**: CLI plumbing for module-body enrichment path
- **hew-cli/compile.rs**: surface enriched diagnostics from non-root module bodies
- **std/**: minor parity fixes in `vec.hew` and `datetime.hew`

### Incidental: close/drop bitcast cleanup (hew-codegen)
Two small, identical hunks in `MLIRGenStmt.cpp` fix a latent type-mismatch bug in the close/drop alloca paths: when the value being stored is not already an `LLVMPointerType`, a `BitcastOp` is now inserted before the `StoreOp`. This was discovered while validating drop paths during the module-body work. It is a correctness fix (prevents potential miscompilation/crash for non-pointer close targets) and is low-risk, but it is explicitly disclosed here to keep scope visible.

### Validation
- hew-types: 415/415 tests pass
- hew-serialize: 117/117 tests pass
- Worktree clean at head `df889e55`

### DIFF
--- a/hew-cli/src/compile.rs
+++ b/hew-cli/src/compile.rs
@@ -460,6 +460,11 @@ fn enrich_program_ast(
     source: &str,
     input: &str,
 ) -> Result<Vec<hew_serialize::ExprTypeEntry>, String> {
+    let root_module_item_count = program
+        .module_graph
+        .as_ref()
+        .and_then(|mg| mg.modules.get(&mg.root).map(|module| module.items.len()));
+
     // Flatten resolved_items from module imports into top-level items.
     // Flattening must happen before enrichment so that normalize_all_types
     // and enrich_fn_decl process the imported functions too.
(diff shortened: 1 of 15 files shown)


slepp:
All content from this branch was already squash-merged to main as #756 (955d4df6), including the non-root module body parity work and the incidental MLIRGenStmt.cpp close/drop bitcast fix. This branch is now stale/conflicting against current main, so closing as superseded rather than rebasing duplicate changes.

PR #1003, opened 2026-04-14
fix(runtime): replace poison-skipping if-let-Ok with lock_or_recover on TOP_LEVEL_SUPERVISORS

slepp:
## Problem

Four production call sites in `shutdown.rs` used `if let Ok(...) = mutex.lock()` on `TOP_LEVEL_SUPERVISORS`, which silently dropped work when the mutex was poisoned:

- `hew_shutdown_register_supervisor` — new registrations silently lost
- `hew_shutdown_unregister_supervisor` — unregistrations silently lost
- `free_registered_supervisors` — supervisor memory leaks on poisoned state
- `shutdown_orchestrate` drain path — **fell back to an empty `Vec::new()`, skipping all supervisor stops entirely during graceful shutdown**

The last case is the most serious: a poisoned mutex causes the next shutdown to silently skip stopping every registered supervisor.

This was inconsistent with the rest of the file — `is_supervisor_registered_for_test`, `registered_supervisors_snapshot`, and the test reset path all already used `lock_or_recover()`.

## Fix

Replace all four `if let Ok(...)` sites with `MutexExt::lock_or_recover()`, which recovers poisoned state via `PoisonError::into_inner` and proceeds. The recovered data is still valid, so recovering and continuing is correct.

The `shutdown_orchestrate` drain path no longer falls back to an empty `Vec` — it now always drains the real supervisor list.

## Tests

Four focused regression tests poison `TOP_LEVEL_SUPERVISORS` via `std::panic::catch_unwind` and verify each fixed call site still performs its work:

- `register_supervisor_recovers_from_poisoned_mutex`
- `unregister_supervisor_recovers_from_poisoned_mutex`
- `free_registered_supervisors_recovers_from_poisoned_mutex`
- `shutdown_orchestrate_recovers_from_poisoned_supervisor_mutex`

All 1109 `hew-runtime` tests pass.

## Context

The timer test flake that was also present on `fix/shutdown-supervisor-lock-poison` was already fixed in #1001. This PR contains only the `TOP_LEVEL_SUPERVISORS` poison-recovery fix, cleanly cherry-picked onto current main.

### DIFF
--- a/hew-runtime/src/shutdown.rs
+++ b/hew-runtime/src/shutdown.rs
@@ -100,10 +100,9 @@ pub unsafe extern "C" fn hew_shutdown_register_supervisor(
     if sup.is_null() {
         return;
     }
-    if let Ok(mut sups) = TOP_LEVEL_SUPERVISORS.lock() {
-        if !sups.iter().any(|s| s.0 == sup) {
-            sups.push(SupervisorPtr(sup));
-        }
+    let mut sups = TOP_LEVEL_SUPERVISORS.lock_or_recover();
+    if !sups.iter().any(|s| s.0 == sup) {
+        sups.push(SupervisorPtr(sup));
     }
 }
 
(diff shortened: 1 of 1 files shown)


slepp:
Superseded by #1005 — same core fix rebased onto `931846d2` (includes merged #1001) with improved drain-path test using `drain_supervisors_for_test()` helper to avoid scheduler/ticker coupling.

PR #1017, opened 2026-04-14
fix(codegen): isolate lambda actor receive drop scope

slepp:
## Summary
- give lambda actor receive parameters their own drop scope instead of leaking enclosing function-level exclusion state
- preserve the surrounding function-generation state after the receive body finishes
- add focused MLIR/codegen regressions for owned param cleanup and exclusion-state restoration

## Validation
- make codegen-test PATTERN='^mlirgen$'
- make test-codegen

### DIFF
--- a/hew-codegen/src/mlir/MLIRGenActor.cpp
+++ b/hew-codegen/src/mlir/MLIRGenActor.cpp
@@ -1307,14 +1307,55 @@ mlir::Value MLIRGen::generateSpawnLambdaActorExpr(const ast::ExprSpawnLambdaActo
     FunctionGenerationScope funcScope(*this, recvFuncOp);
     auto prevFuncLevelDropScopeBase = funcLevelDropScopeBase;
     funcLevelDropScopeBase = dropScopes.size();
+    auto prevFuncLevelDropExcludeVars = std::move(funcLevelDropExcludeVars);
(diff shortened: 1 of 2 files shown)


slepp (commented):
**READY** — independent final-gate review

**Core fix** — lambda actor receive drop scope isolation is correct:
- All 8 `funcLevel*` fields are saved/cleared/restored; the set matches the MLIRGen header exactly.
- `pushDropScope`/`popDropScope` bracket matches the named-actor receive handler pattern.
- Param binding correctly adapts to `LambdaParam` (which lacks `is_mutable`, so the `mutableVars.lookup` fallback is the right check; `bindParam(const Param&)` cannot be reused due to different AST type).
- No early-return paths between save and restore — follows existing manual save/restore convention.
- Both regression tests target the right invariants.

**Non-blocking follow-up** — the bundled removal of `panicOnScopeSpawnFailure` leaves an inconsistency: `hew_scope_spawn` in `hew-runtime/src/scope.rs:163-166` still returns `-1` when `HEW_SCOPE_MAX_ACTORS` (64) is exceeded, but the codegen now silently discards the result. An actor that fails registration will not be tracked by `hew_scope_wait_all`, quietly breaking the structured-concurrency guarantee for that scope. Consider either removing the runtime-side limit (switching `actors` to a `Vec`) or making `hew_scope_spawn` panic internally to keep the contract fail-closed.

PR #1030, opened 2026-04-14
fix(stdlib/http): align string helpers to Option<String>

slepp:
## Summary
- make request_string, get_string, and post_string return Option<String> to match the existing null-as-None ABI
- update the Hew HTTP surface docs and examples to use Option semantics
- add surface coverage for the Option-return contract

## Validation
- cargo test -p hew-std-net-http

### DIFF
--- a/std/net/http/http_client.hew
+++ b/std/net/http/http_client.hew
@@ -9,6 +9,8 @@
 //!
 //! Response accessors: `status()`, `body()`, `header(name)`, `content_type()`,
 //! `headers()` (all headers as `Vec<(String, String)>`), and `free()`.
+//! The `_string` convenience helpers (`request_string`, `get_string`,
+//! `post_string`) return `Option<String>` — `None` on transport failure.
 //!
 //! # Examples
 //!
@@ -16,8 +18,10 @@
 //! import std::net::http::http_client;
 //!
 //! fn main() {
-//!     let body = http_client.get_string("https://example.com");
-//!     println(body);
+//!     match http_client.get_string("https://example.com") {
+//!         Some(body) => println(body),
+//!         None => println("request failed"),
+//!     }
 //!
 //!     let headers: Vec<(String, String)> = Vec::new();
 //!     headers.push(("Accept", "application/json"));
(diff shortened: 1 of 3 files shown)


slepp (commented):
## BLOCKED — ABI mismatch at the FFI boundary

The public API change (`request_string`, `get_string`, `post_string` → `Option<String>`) is the right call, but the **extern "C" binding declaration** was also changed to `-> Option<String>`, which creates a real ABI mismatch.

### Evidence

1. **C/Rust side** (`std/net/http/src/client.rs:682-691`): `hew_http_request_string_hew` returns `*mut c_char` — a raw pointer, `NULL` on transport failure.


3. **No FFI bridge** (`hew-codegen/src/mlir/MLIRGen.cpp:1658-1698`): extern `"C"` lowering is generic — `Option` is not special-cased. There is no nullable-pointer-to-Option conversion at FFI boundaries.

4. **No precedent**: no existing `extern "C"` block in the repo uses `Option` in a return type. The ad-hoc Option wrapping for channels/streams (`MLIRGenExpr.cpp:1564-1597`) is hardcoded call-site lowering, not generic.

### Impact

At runtime, the codegen will expect an `(i32, String)` struct return from the C function, but the C side returns a bare pointer. This is an ABI-level type confusion — likely a crash or garbage data.

### Suggested fix

Keep the extern binding as `-> String` (matching the actual C ABI), and perform the null-to-Option conversion in the Hew wrapper, following the pattern used for `hew_channel_try_recv`. Alternatively, add the `hew_http_request_string_hew` call site to the codegen's ad-hoc nullable-pointer wrapping list alongside the channel hooks.

slepp (commented):
## READY — prior ABI blocker resolved, gate passes

Verified at head `b65c1f1eaafba3c6fc8341856af4fecadaf17325`.

### Gate checklist

| Check | Result |
|---|---|
| **Extern block ABI-safe** | ✅ `hew_http_request_string_hew` removed from extern block. No `Option` in any extern return type. |
| **Option conversion in Hew** | ✅ `request_string` calls `hew_http_request_hew` → checks `status < 0` → returns `None` or `Some(body)`. Pure Hew logic, no ABI crossing. |
| **Drop safety** | ✅ `hew_http_response_free(resp)` called in both the `None` (status < 0) and `Some` paths. No resource leak. |
| **Docs / examples coherent** | ✅ Module doc, `request_string`, `get_string`, `post_string` doc examples all updated to `match Some/None`. |
| **Surface test** | ✅ `http_request_surface_test.hew` exercises the `Option<String>` return contract at the Hew level. |
| **Rust primitive-chain test** | ✅ New `loopback_primitive_chain_some_path` covers the exact `request → status → body → free` primitive chain end-to-end. |
| **No unintended breakage** | ✅ `get_string`/`post_string` delegate to `request_string`; no other callers in the codebase. |

Sound to auto-squash once CI clears.

PR #1050, opened 2026-04-14
feat(runtime): WASM mailbox parity — try_push / sys_len, Fail→-1 mapping, DropOld OOM safety, coalesce+reply certification

slepp:
## Summary

Completes WASM mailbox parity for the two API surfaces the native mailbox exposed but the WASM counterpart lacked, and fixes two correctness bugs found during review.

### New API surfaces (`hew-runtime/src/mailbox_wasm.rs`)

- **`hew_mailbox_try_push`** — policy-aware push with fine-grained status codes mirroring the native contract:
  - `0` = Enqueued, `1` = DropNew, `2` = DroppedOld, `3` = Coalesced, `-1` = Fail (distinct from DropNew)
- **`hew_mailbox_sys_len`** — returns the count of system messages in the mailbox, matching `hew_mailbox_sys_len` on native.

### `SendOutcome` refactor

`send_user_message` is split into `send_user_message_inner` (returns `SendOutcome`) shared by both `send` and `try_push`, plus a thin mapping layer. This lets `Fail` and `DropNew` produce distinct `try_push` return codes (`-1` vs `1`) while both still map to `ErrMailboxFull` through `hew_mailbox_send`.

### Bug fixes

1. **Fail → `-1` (was `1`)** — `Fail` policy now correctly returns `-1` from `hew_mailbox_try_push`. Same fix applied to the coalesce-fallback `Fail` path.
2. **DropOld OOM safety** — the replacement node is now allocated *before* evicting the oldest queued message. If allocation fails, the queue is left unchanged (no message loss). Previously, OOM after eviction would permanently discard the old message. Fixed in both the top-level DropOld path and the coalesce → DropOld fallback.

### 18 certification tests

| Category | Tests |
|---|---|
| `try_push` return codes | enqueued, dropped, dropped_old, coalesced, fallback_dropnew, closed, null_data |
| Fail mapping | `try_push_fail_policy_returns_negative`, `try_push_coalesce_fallback_fail_returns_negative` |
| DropOld OOM safety | `try_push_drop_old_oom_preserves_queue`, `try_push_coalesce_fallback_drop_old_oom_preserves_queue` |
| Coalesce + reply-channel | `coalesce_reply_channel_preserved_on_merge`, `coalesce_no_reply_channel_does_not_retire`, `coalesce_same_reply_channel_not_retired`, `coalesce_drop_old_fallback_retires_evicted_reply_channel` |
| `sys_len` | tracks system messages, zero on fresh, unaffected by close, independent of user queue |

### Scope

Single file: `hew-runtime/src/mailbox_wasm.rs`. No protocol or serialization changes. No other crates touched.

---

Part of v0.3.2 bounded WASM-native parity work.

### DIFF
--- a/hew-runtime/src/mailbox_wasm.rs
+++ b/hew-runtime/src/mailbox_wasm.rs
@@ -381,21 +381,42 @@ unsafe fn retire_reply_channel(reply_channel: *mut c_void) {
     }
 }
 
-unsafe fn send_user_message(
+/// Outcome of an overflow-policy-aware send into the user queue.
+///
+/// Mirrors the native `SendOutcome` in `mailbox.rs`. FFI entry points map
+/// these variants to their own return conventions.
+#[derive(Clone, Copy)]
+enum SendOutcome {
+    /// Message was successfully enqueued.
+    Enqueued,
+    /// Mailbox is closed — message was not sent.
+    Closed,
+    /// Message intentionally dropped (`DropNew` / `Fail` / `Block` on WASM).
+    Dropped,
+    /// Oldest message was evicted to make room (`DropOld` policy).
+    DroppedOld,
+    /// Message payload was merged with an existing queued message
+    /// (`Coalesce` policy).
+    Coalesced,
+    /// Memory allocation failed.
+    Oom,
+}
+
+unsafe fn send_user_message_inner(
     mb: &mut HewMailboxWasm,
     msg_type: i32,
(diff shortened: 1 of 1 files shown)


slepp:
Closing as duplicate already landed on main via #969. Final gate found this branch predates later mailbox instrumentation from #1020 and would regress scheduler recording if merged, so this lane is being retired as stale merged residue.

PR #1071, opened 2026-04-15
codegen: fail closed method receiver dispatch

slepp:
## Summary
- fail close Duration receiver classification in `MLIRGenExpr.cpp` by requiring resolved receiver types before duration-only dispatch
- fail close Rc receiver classification when `ExprMethodCall` dispatch depends on receiver `expr_types` and no `method_call_receiver_kinds` authority exists
- add MLIRGen regressions covering positive Duration/Rc lowering and hard errors when receiver type metadata is missing

## Scope
Included in this PR:
- Duration receiver detection in `hew-codegen/src/mlir/MLIRGenExpr.cpp`
- Rc receiver detection in `hew-codegen/src/mlir/MLIRGenExpr.cpp`
- Regression coverage for well-typed programs and missing receiver `expr_types` metadata

Deferred follow-up:
- trait receiver fallback at the `resolvedTypeOf` seam
- struct/enum name-resolution fallback at the `resolvedTypeOf` seam

## Validation
- `cargo clippy --workspace --quiet`
- `cargo fmt --check`
- `make codegen-test PATTERN='^mlirgen$'`

### DIFF
--- a/hew-codegen/src/mlir/MLIRGenExpr.cpp
+++ b/hew-codegen/src/mlir/MLIRGenExpr.cpp
@@ -4289,11 +4289,19 @@ std::optional<mlir::Value> MLIRGen::generateBuiltinMethodCall(const ast::ExprMet
   }
 
   // --- Duration methods (inline arith ops on i64 nanosecond values) ---
-  // Duration maps to i64 at the MLIR level; use resolvedTypeOf to distinguish
-  // from plain i64 integer methods.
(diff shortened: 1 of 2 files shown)


slepp:
Superseded by #1072, which is the chosen v0.3 compiler closeout slice. Keeping this branch parked as a possible later follow-on if it still matters after the coercion-proof tranche lands.

PR #1171, opened 2026-04-16
fix(types): suppress struct-pattern cascade over error scrutinees

slepp:
Summary
- bind struct-pattern fields when the scrutinee has already degraded to Ty::Error or remains Ty::Var
- add focused regressions for plain struct patterns, struct variants, and renamed field subpatterns
- keep the follow-up tightly scoped to checker-boundary cascade suppression in hew-types

Context
PR #1164 merged before this amendment could be pushed, so this branch now carries the minimal follow-up delta on top of main.

Validation
- cargo fmt --all --check
- cargo test -p hew-types typecheck_error_scrutinee_ -- --nocapture
- cargo test -p hew-types typecheck_struct_pattern_unknown_field_errors -- --exact
- cargo clippy -p hew-types --tests -- -D warnings

### DIFF
--- a/hew-types/src/check/patterns.rs
+++ b/hew-types/src/check/patterns.rs
@@ -155,6 +155,20 @@ impl Checker {
                             }
                         }
                     }
+                } else if matches!(ty, Ty::Var(_) | Ty::Error) {
+                    for pf in fields {
+                        if let Some((pat, ps)) = &pf.pattern {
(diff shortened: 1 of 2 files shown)


slepp:
Superseded by #1172, which is the same narrow struct-pattern cascade fix re-cut from fresh main after #1164 merged.

PR #1189, opened 2026-04-16
runtime: guard HEW_TRANSPORT with ENV_LOCK

slepp:
## Summary
- guard HEW_TRANSPORT reads in `hew_node_start` with `ENV_LOCK.read_or_recover()`
- guard `hew_node_api_set_transport` writes with `ENV_LOCK.write_or_recover()`
- expose the existing runtime env lock to `hew_node.rs` without changing behavior

## Validation
- `cargo clippy -q -p hew-runtime --lib -- -D warnings`
- `cargo test -q -p hew-runtime --lib`


slepp:
Superseded by #1225 (PoisonSafe wrapper sweep). The HEW_TRANSPORT env access now goes through ENV_LOCK.read_access / ENV_LOCK.access on the new closure-only API in hew-runtime/src/lifetime/poison_safe.rs, so the raw RwLock plus voluntary _or_recover helper pattern is replaced throughout.

Issue #1230, opened 2026-04-17
runtime: typed, quiescence-aware drain-actors primitive for JIT sessions


slepp:
Parent: #1226

## Capability

A public, typed, quiescence-aware primitive for draining a set of actors with a caller-supplied deadline and an explicit typed outcome.

## Why this is needed regardless of redefinition policy

The persistent-JIT-session work in #1226 (milestone 3) needs this primitive no matter which redefinition policy lands:

- **Shadow-and-retain** needs it to decide when an old `JITDylib` / `ResourceTracker` has no live referrers and can be released.
- **Gated-on-quiescence** needs it as the precondition check before swapping code.
- **Fail-closed** needs it to answer "is any actor still using the old definition?" without racing.

Milestone 2's `:reset` soak test ("N repeated evaluations show flat memory and no orphaned actors") also consumes the same surface.

## What exists today

- `hew_actor_stop` (`hew-runtime/src/actor.rs:1496-1535`) is cooperative — it closes the mailbox and enqueues one shutdown system message. Return value is `void`; callers cannot tell whether the actor has actually drained.
- `hew_actor_free` (`hew-runtime/src/actor.rs:1548-1583`) embeds an internal 2-second spin-wait on quiescence and returns `-2` on timeout. The deadline is hard-coded, it only handles one actor at a time, and the waiting happens inside the free path rather than as a separable query.
- Cleanup ordering (timers → links → monitors → named registrations → terminate → free) in `hew_actor_free` at `:1585-1615` is the invariant the new primitive must preserve on the successful-drain path. `cleanup-all-exits` applies directly.

There is no existing way for a session-layer caller to ask "have these N actors drained yet?" and get back a typed answer in bounded time.

## Proposed shape (to be refined in design)

A session-facing entry point, roughly:

```
enum DrainOutcome { Drained, StillLive(Vec<ActorId>), Crashed(Vec<ActorId>) }
fn drain_actors(ids: &[ActorId], deadline: Instant) -> DrainOutcome;
```

Implementation notes:

- Sends `hew_actor_stop` to each actor up front, then polls quiescence for the set with a shared deadline (not N independent 2-second spins).
- Preserves the `hew_actor_free` cleanup ordering for actors that do drain; leaves the rest in place with their typed outcome so the caller decides escalation (force-free with `-2`, fail-closed, or just report).
- Fail-closed on deadline: `StillLive` is a first-class outcome, not a panic.
[...]

Issue #1226, opened 2026-04-17
JIT execution path for Hew (native, opt-in)


slepp:
## Summary

Hew today ships code to users through an AOT pipeline: parse, typecheck, lower to MLIR, lower to LLVM IR, emit an object, link with `clang`, and run. That is the right default for production binaries but is a poor fit for interactive workloads — REPL evaluation, one-shot expression eval, hosted playground cells, and live tooling — where the link-and-spawn round trip dominates perceived latency. A just-in-time execution path would let a Hew developer type an expression in `hew eval` or `hew` (REPL) and see it run without paying for temp object files, external linkers, or a subprocess boundary. The goal is a native, opt-in JIT that reuses the existing frontend and codegen, targets realistic Hew programs (actors, channels, timers, runtime services), and stays competitive with established JITted dynamic-language VMs for typical interactive workloads.

## Current state

- AOT-only execution. `hew-codegen/src/codegen.cpp` builds an `llvm::Module` via `Codegen::buildLLVMModule` (declared in `hew-codegen/include/hew/codegen.h`), writes an object, and the CLI links it with `clang` and spawns a child (see `hew-cli/src/link.rs`, `hew-cli/src/process.rs`).
- The C++ codegen already links LLVM ORCv2 (`hew-codegen/src/CMakeLists.txt` pulls in `OrcJIT`), but nothing currently drives it — there is no runtime codegen entry point and no JIT session type.
- The runtime (`hew-runtime/`) is built around a process-global scheduler, actor registry, timer wheel, and channel machinery. Generated `main` is responsible for scheduler init/shutdown today; there is no re-entrant "run one module, then reset" surface.
- The REPL (`hew-cli/src/eval/`) already synthesizes a module per evaluation and hands it through the full AOT pipeline. It would be the most direct beneficiary of a JIT path.
- WASM output (`hew-wasm/`) is a separate target and is AOT by construction.

## Gap analysis

What is missing to turn the existing AOT backend into a JIT-capable execution path, framed as capabilities rather than tasks:

- **Runtime codegen driver.** A way to take the same LLVM IR that AOT emits and hand it to an in-process execution engine (LLJIT / ORCv2) instead of an object file. The insertion point already exists — between `buildLLVMModule` and object emission — but there is no engine wired up behind it.
[...]

PR #1271, opened 2026-04-17
feat(runtime): session reset-hook registry with shutdown wiring and dispatch-registry clear

slepp:
Delivers Stages 1–3 of the JIT session lifecycle plan (Milestone 2 of #1226). Stage 4 (`hew_session_reset` C ABI entry point) remains blocked on #1230.

## What this does

**Commit 1 — `hew-runtime/src/session.rs` (new, unconditional)**

Introduces a centralised reset-hook registry shared by both schedulers. `register_reset_hook(fn())` + `session_reset()` backed by `Mutex<Vec<ResetHook>>`. Hooks snapshot-and-release before firing to avoid deadlock if a hook re-registers. `tracing::register_trace_reset_hook()` (Once-guarded) enlists `hew_trace_reset` from `hew_sched_init`. SHIM markers placed in `bridge_shutdown()` and native `hew_runtime_cleanup()` where `handler_names.clear()` belongs once `feat/message-handler-name-side-table` rebases onto this module.

Exposes `pub(crate) SESSION_TEST_LOCK` + `reset_hooks_for_test()` for cross-module test isolation (pattern from #1257).

**Commit 2 — Wire `session_reset()` into both `hew_sched_shutdown` paths**

- `scheduler_wasm.rs`: call after `bridge_shutdown()`, before scheduler statics are zeroed.
- `scheduler.rs`: call after all workers are joined AND after `maybe_write_on_exit()` has snapshot the live dispatch-type registry. Ordering invariant: profile snapshot sees live metadata; reset then clears it for the next session.

Closes the native-wasm parity gap: `hew_trace_reset` was registered but never called. Includes `session_reset_clears_trace_state_via_hook` test in `tracing::tests` that acquires `SESSION_TEST_LOCK` first then `TEST_LOCK` (consistent acquisition order) to prevent races with the session tests that clear `RESET_HOOKS`.

**Commit 3 — `clear_dispatch_registry` + profiler reset hook**

- `actor_registry::clear_dispatch_registry()`: empties `DISPATCH_TYPE_REGISTRY` so stale dispatch-pointer-to-type-name mappings cannot bleed across JIT reload cycles.
- `profiler::register_reset_hooks()`: Once-guarded, called unconditionally from `hew_sched_init` (not from `maybe_start` which is HEW_PPROF-gated and would leave the registry un-cleared when the feature is active but the profiler is not started).
- `actor.rs` `hew_actor_register_type` SHIM extended with JIT leak-risk note referencing #1226 M3.
- `dispatch_registry_cleared_after_clear` test in `actor_registry::tests`.

**Commit 4 — LESSONS.md updates**

Strengthens `cleanup-all-exits`, `native-wasm-parity`, and `global-test-isolation` rows with the invariants this work enforces.

## Stage 4 is BLOCKED

`hew_session_reset()` (the public C ABI entry point) requires the `drain_actors` primitive from #1230. Do NOT attempt Stage 4 until #1230 lands.

## Unblocks

Stage 1 resolves the shared-module gap that blocked `feat/message-handler-name-side-table` Commits 2–4. That branch should rebase onto this one and add `state.handler_names.clear()` at the SHIM marker in `bridge_shutdown()`.

## Testing

`make ci-preflight` green locally (full fallback lane: fmt, lint, playground-check, full test suite including C++ unit tests).


slepp:
### Code review

Found 1 issue:

1. Three functions remain `pub` when `pub(crate)` is the correct minimum visibility — all callers are within `hew-runtime`. `session.rs` (`ResetHook`, `register_reset_hook`, `session_reset`) is already narrowed, but:

   - `register_trace_reset_hook` — called only from `scheduler.rs` and `scheduler_wasm.rs`
   - `clear_dispatch_registry` — called only from `profiler/mod.rs` and test code
   - `register_reset_hooks` — called only from `scheduler.rs`

https://github.com/hew-lang/hew/blob/cc3fce37f80fe0ec9bbb1fbf325da6b73f50a109/hew-runtime/src/tracing.rs#L459-L462

https://github.com/hew-lang/hew/blob/cc3fce37f80fe0ec9bbb1fbf325da6b73f50a109/hew-runtime/src/profiler/actor_registry.rs#L109-L112

https://github.com/hew-lang/hew/blob/cc3fce37f80fe0ec9bbb1fbf325da6b73f50a109/hew-runtime/src/profiler/mod.rs#L74-L77

These create accidental semver surface for a crate-internal API. Changing `pub` → `pub(crate)` on all three tightens the visibility.

PR #1292, opened 2026-04-18
fix(net): guard hew_http_server_close against double-free (closes #1281)

### LINKED ISSUE #1281: HTTP Server double-free: `close()` + `Drop` both invoke `hew_http_server_close`
## Summary

`std/net/http/http.hew` declares both:

- `impl ServerMethods for Server { fn close(srv) { unsafe { hew_http_server_close(srv) } } }`
- `impl Drop for Server { fn drop(srv) { unsafe { hew_http_server_close(srv) } } }`

Both call the same FFI function. Any Hew program that calls `server.close()` explicitly will hit a double-free (SIGSEGV) when the `Drop` impl fires at the end of the variable's scope.

## How it was found

The `e2e_http_server/http_try_respond_stream` e2e test (added in PR #1266) originally called `server.close()` inside a `scope |s| { s.spawn { ... server.close() } }` block. The test crashed with exit code 139 (SIGSEGV). `lldb` backtrace confirmed `hew_http_server_close` was called twice — once from the explicit `server.close()` in the spawn task, and once from the Drop impl when `server` went out of scope at the end of `main`.

The test was fixed by removing the explicit `close()` call, relying solely on `Drop`. But the public API still exposes `close()`, and users who call it will hit the same crash.

## Contrast with QUIC

`std/net/quic/quic.hew` has the opposite contract: `impl QUICEndpointMethods for QUICEndpoint { fn close(...) { ... } }` with **no** `impl Drop for QUICEndpoint`. The endpoint must be closed explicitly. HTTP went the other way (Drop-based cleanup) but kept the explicit `close()` method, creating an inconsistency.

## Fix options

**Option A** — Remove `impl Drop for Server` from `http.hew` and require explicit `close()`, matching QUIC's pattern. Any existing code that relies on implicit cleanup would need to add `server.close()`.

**Option B** — Remove `fn close(srv: Server)` from `ServerMethods` and `impl ServerMethods for Server`, leaving `Drop` as the sole release mechanism. The doc comment on `Server` would need updating.

**Option C** — Make `hew_http_server_close` idempotent: replace `Box::from_raw(srv)` with an atomic-take pattern (e.g., swap the pointer to null before freeing, so a second call is a no-op). This is safe but adds indirection.

Option B is likely the cleanest — the Drop-based model is already wired and the only HTTP server test never needed explicit `close()`. Option A would require auditing all call sites. Option C is defensive and hides a logic error.

## Files

- `std/net/http/http.hew` — both `impl ServerMethods` and `impl Drop for Server`
- `std/net/http/src/server.rs` — `hew_http_server_close` (not idempotent)

slepp:
## Status: Parked — Hew-layer fix blocked by language gap (#1295)

The preferred fix (add `closed: bool` to `Server`, guard Drop) cannot be expressed in Hew today. Struct methods receive the receiver by value; mutations inside `close()` are not visible to the Drop that fires at the caller's scope exit. See #1295 for the full analysis and what language changes would unblock the clean fix.

## What this PR does instead

Makes `hew_http_server_close` idempotent at the FFI layer by wrapping the server handle in `Option<Box<...>>`. The first call extracts and frees the server via `.take()`; subsequent calls detect `None` and no-op safely.

This is a sound workaround until #1295 is resolved. The outer `Box` (the `HewHttpServer` allocation) is not freed by `close()` — it is freed only when the caller's Drop fires. That means one small allocation outlives the listening socket, which is acceptable for a server handle.

## Why the FFI-layer fix is kept

- The alternative (revert to single-consume destructor, no Hew-side guard) leaves the double-free in place for every user who calls `close()` then lets the binding fall out of scope.
- The FFI-level `Option::take()` pattern is already used in this codebase (`hew-cabi/src/sink.rs`).
- Thread-local tracking (rev1, reverted) was unsound at FFI boundaries.

## Not changing

- Public C FFI signatures
- Behaviour under normal (single close per handle) usage
- Thread safety model

## Validation

- `make ci-preflight` PASS (785 tests, 5 C++ tests)
- 4 new unit tests: same-thread idempotence, cross-thread safety, drop-after-close, drop-without-close

## Follow-up

- #1295 tracks the language gap. When `consume`/`&mut self`/linear-type semantics land, this PR's FFI workaround can be replaced with a clean Hew-side guard.
- `std/net/tls/tls.hew` (`TlsStream`) and `std/text/regex/regex.hew` (`Pattern`) have the same double-Drop structure and should be audited after #1295 is addressed.


slepp:
Marking as draft. Gate review () correctly identifies that this fix makes `hew_http_server_close` idempotent on the inner resource but never frees the outer `Box<HewHttpServer>` that `hew_http_server_new` allocated — every `server.close()` leaks the wrapper.

The real bug is on the Hew side: `http.hew` has **both** `fn close(srv: Server)` (line 51) **and** `impl Drop for Server` (line 156), so the FFI destructor is called twice. Canonical pattern is `std/net/smtp/smtp.hew`, which has `fn close(conn)` **only** — no `impl Drop`, so the destructor is called exactly once.

Options for a proper fix:
1. Remove `impl Drop for Server` from `http.hew` and rely on explicit `server.close()`. Matches smtp. Breaking if any caller depends on auto-close.
2. Keep `impl Drop` and make `fn close()` consuming + suppress Drop. Requires Hew-side ownership semantics (`mem::forget`-equivalent).
3. Restructure via a scoped `with_server` pattern (matches the `CloseOnDrop` helper in `std/net/smtp/src/lib.rs:32-55`).

Will return with a plan before continuing.

slepp:
**Design change summary (per Discord direction):**

The direction was to revert the FFI-level idempotence hack and instead fix at the Hew layer: add `closed: bool` to `Server`, set it in `close()`, and guard Drop against double-free.

After probing the Hew compiler:

1. `var f: Foo` in struct method receivers is rejected — the compiler says "struct methods receive self by value" and `var self` has no effect.
2. `fn close(f: Foo)` passes a copy. Mutations to `f.closed` inside `close()` are not visible to the caller's binding. Drop fires twice: once inside `close()` when `f` goes out of scope, and again when the caller's binding goes out of scope. This is the exact double-free path.
3. A return-modified-copy approach (`fn close(f: Foo) -> Foo`) does not help: the caller's original still drops (with `closed=false`), then the returned copy drops (with `closed=true`) — still double-drop.

This is a language gap, not a logic gap. Filed #1295 to track the `&mut self` / consume / linear-type feature needed to express the clean fix.

The current rev2 FFI workaround (`Option<Box<...>>` with `take()`) is the correct approach while #1295 is open. PR updated to reflect this and marked ready for review.

slepp:
Closing as superseded by the explicit-release fix for #1281 on branch `fix/1281-http-double-free` (commit 9522df3e). This replacement removes the duplicate Hew-layer destructor entries so `http.Server.close()` and `regex.Pattern.free()` are the single canonical release paths, making the parked FFI-level idempotence workaround unnecessary.

Issue #1276, opened 2026-04-17
wirecodec: revisit char BMP bounds and Duration u64::MAX in public descriptor API


slepp:
## Status

**Duration side resolved** by PR #1256 via `duration_field_carries_integer_bounds` test at `hew-wirecodec/src/json_desc.rs:298-316`, pinning `bounds.max == u64::try_from(i64::MAX).unwrap()` — signed i64 nanosecond contract. The Duration bounds are correct; `u64::MAX` was a misread in earlier reviews.

**Char side — accepted as intentional BMP-only**, deferred to a dedicated bounds-refactor. Rationale:

- `IntegerBounds::for_kind(Char)` at `hew-wirecodec/src/plan.rs:54` maps `Char` to U16 bounds (0..=0xFFFF).
- This matches the existing `MLIRGenWire.cpp:127,147-149` C++ path which already emits `WireJsonKind::Integer` and uses BMP-only encoding for msgpack parity.
- Lifting bounds to full Unicode scalar (0..=0x10FFFF) requires audit + changes to: `hew-serialize/src/msgpack.rs` (Char-specific encode/decode path), the C++ consumer's handling of SMP codepoints, and any wire-tag/index assumptions about BMP-only storage.
- The existing `SetChar` SHIM comment in `json_desc.rs:51-58` explicitly cites this follow-up as deferred.

## Action

Before a descriptor-driven C++ consumer lands (Phase M of the canonical Value plan), audit the BMP assumption against `hew-serialize/src/msgpack.rs`:
- If BMP is a deliberate wire-format invariant, document it in `plan.rs:54` with a formal `SHIM:` block citing this issue and close this.
- If BMP is incidental, update `IntegerBounds::for_kind(Char)` to `{0, 0x10_FFFF}` and update the C++ consumer.

No action needed to unblock #1256 — the bounds design is frozen as the contract the current runtime already implements. Descriptor-API fidelity is maintained (descriptor matches runtime, both BMP).

Labels: `wirecodec`, `design`, `deferred-by-design`

PR #1294, opened 2026-04-18
feat(wirecodec): wire SetChar/SetDuration emission and unify struct/enum traversal (closes #1272, #1274)

### LINKED ISSUE #1272: wirecodec: wire SetChar/SetDuration ops into runtime + MLIRGenWire consumer
## Summary

The `JsonOp` descriptor exports `SetChar` and `SetDuration` operations (and `YamlOp` aliases them), but the current C++ consumer has no runtime entry points or descriptor-driven code path for these kinds.

## Current gap

In **MLIRGenWire.cpp:147-149**, both `Char` and `Duration` are routed through `WireJsonKind::Integer`:

```cpp
case hew::WireKind::Char:
case hew::WireKind::Duration:
    emitJsonOp(field, WireJsonKind::Integer);
```

This means:
- No `hew_json_object_set_char` or `hew_json_object_set_duration` runtime calls exist yet.
- The descriptor-driven path (descriptor ops → runtime calls) is not wired up for these kinds.
- The C++ consumer reads `jsonKindOf` directly instead of reading the descriptor's op stream.

## Forward path

Implement dedicated runtime entry points and update the MLIRGenWire consumer to:
1. Read `SetChar` and `SetDuration` from the descriptor's op stream.
2. Emit `hew_json_object_set_char` and `hew_json_object_set_duration` calls instead of routing through Integer.
3. Remove the routing heuristic at MLIRGenWire.cpp:147-149.

This is part of the larger descriptor-driven shrink of `MLIRGenWire` (see Canonical Value plan, Phase M scope).

## Reference

- PR #1256 (wirecodec JSON+YAML descriptor infrastructure).
- SHIM markers added in commit 5e13b911 document the removal condition.

### LINKED ISSUE #1274: wirecodec: decouple YAML descriptor from JSON layer and unify struct/enum traversal
## Scope

This is a pre-existing pattern expansion in the wirecodec codebase, discovered during the JSON/YAML codec unification PR (#1256). Both items are out of scope for the current lane but block future descriptor-driven architecture improvements.

## Finding 1: YAML descriptor depends on JSON layer

`yaml_desc.rs::from_plan` reuses `json_op_for_kind` from the `json_desc` module. This creates a layer inversion:

- The codebase pattern is to keep shared logic in neutral modules (`kind.rs`, `plan.rs`), and let format-specific modules (JSON, msgpack) consume them.
- `yaml_desc` imports and depends on `json_desc::{json_op_for_kind, JsonOp}`, making it harder to evolve YAML independently if required (e.g., if YAML later adds format-specific ops like block vs flow style).

**Proposal:** Extract `json_op_for_kind` to a new neutral module (e.g., `op_codec.rs` or fold into `kind.rs`) so both `json_desc` and `yaml_desc` consume the same neutral source. Update the alias contract in `yaml_desc.rs::YamlOp` to reference the neutral location.

**Evidence:** PR #1256 (commit 49851cc7), lines 20-30 in `yaml_desc.rs`.

## Finding 2: Struct/enum field traversal is duplicated

Three `from_plan` implementations duplicate the struct/enum shape-match and field iteration:

- `json_desc.rs::JsonCodecDesc::from_plan` (lines 120-139)
- `yaml_desc.rs::YamlCodecDesc::from_plan` (lines 67-86)
- `msgpack_desc.rs::MsgpackCodecDesc::from_plan` (similar pattern in existing code)

Each re-implements the same traversal:
```rust
match &plan.shape {
    WireShape::Struct { fields } => (
        fields.iter()
            .filter(|f| !f.modifiers.is_reserved)
            .map(field_op_from_plan)
            .collect(),
        Vec::new(),
    ),
    WireShape::Enum { variants } => (
        Vec::new(),
        variants.iter().map(|v| v.name.clone()).collect(),
    ),
}
```

**Proposal:** Extract a generic traversal helper in `plan.rs` (e.g., `WireCodecPlan::fold_for_codec`) that takes format-specific closures for field and variant handling. All three descriptors call the same traversal, reducing duplication and ensuring consistent filtering logic.

**Evidence:** PR #1256 and broader `hew-wirecodec` codebase.

## Phase M (Canonical Value)

Both findings feed into the Canonical Value plan's Phase M scope: consolidating descriptor logic and preparing for C++ consumer replacement of `jsonKindOf`.

---

**Related:** PR #1256 (foundations/wirecodec-json-yaml), Issue #1272 (MLIRGenWire shrink).

slepp:
Wires the `SetChar` / `SetDuration` wire ops through the JSON, YAML, and MLIRGenWire layers so generated code uses dedicated codec entry points instead of routing `char` / `Duration` fields through the generic `Integer` path. Unifies struct and enum `fold_shape` traversal through a shared neutral `op_codec` module. Adds matching runtime C ABI entry points (`hew_*_object_set_char`, `hew_*_object_set_duration`, `hew_*_get_char`, `hew_*_get_duration`) for both JSON and YAML. Restores a full-Unicode bounds check on the new dedicated Char decode paths to preserve the pre-existing fail-closed behaviour for invalid codepoints.

## What this PR does

- Dedicated MLIR decode path for `char` in enum-variant and struct body contexts, backed by full-Unicode `[0, 0x10FFFF]` bounds check (matches pre-PR Integer-path behaviour).
- Dedicated MLIR encode/decode path for `Duration` that no longer routes through generic Integer, and drops two dead `payload.getType() == i32Type` branches (Duration maps unconditionally to i64 per `wireTypeToMLIR:177`).
- Refactor extracts `op_for_kind` + `fold_shape` to a neutral `op_codec` module; JSON- and YAML-descriptor emitters share the traversal.
- New C ABI set/get entry points for `char` and `Duration` on JSON and YAML runtime libraries.
- Corrected SHIM comments in `json_desc.rs` (removes stale "no runtime entry point exists today" lines that the new wiring invalidates).
- Corrected `u32` → `i64` parameter types in the `set_char` doc comments.
- New enum-variant serializer test: `test_wire_enum_char_duration_variant_serializers`.

## What it does NOT do

- Does not change `plan.rs` `IntegerBounds::for_kind(Char)` — still `u16::MAX` at the plan layer. The plan / MLIRGen asymmetry is pre-existing (predates this PR) and tracked as part of #1276.
- Does not reject surrogate codepoints (U+D800..U+DFFF). The bounds check validates Unicode-scalar-range `[0, 0x10FFFF]` which admits surrogates. Stricter Unicode-scalar validation is a separate question; tracked as part of #1276.
- Does not remove the old `json_op_for_kind` forwarder. It remains as a thin alias while callers migrate.

## Breaking changes

None. The Char decode path still panics on out-of-range inputs (same as before); the set of rejected inputs is unchanged (`> 0x10FFFF`). Valid codepoints including emoji continue to round-trip.

## Scope claims

- Closes #1272 (wire `SetChar`/`SetDuration` ops into runtime + MLIRGenWire consumer).
- Closes #1274 (decouple YAML descriptor from JSON layer; unify struct/enum traversal).
- **Does not close** #1276 (BMP-only vs full-Unicode `char` bounds policy + msgpack parity migration) — remains open as a design discussion.

## Validation

- `make ci-preflight` green locally: 352 Rust tests + 5 C++ tests, 0 failures.
- Gate: cross-ecosystem `review_worktree --models gpt-5.4` returns CONDITIONAL (3 phases, 3 advisories). Remaining findings are either tracked by #1276 (surrogate handling, plan.rs asymmetry) or minor (hardcoded "JSON" in a shared error string, stale runtime-lib docs mentioning "BMP-only"). None block correctness of the advertised scope.

### DIFF
--- a/hew-codegen/src/mlir/MLIRGenWire.cpp
+++ b/hew-codegen/src/mlir/MLIRGenWire.cpp
@@ -122,7 +122,18 @@ static bool isVarintType(const std::string &ty) {
 
 /// Classifies a wire type for JSON/YAML serialization.
 /// Resolves the semantic kind from the type name string.
-enum class WireJsonKind { Bool, Float32, Float64, String, Bytes, Integer };
+///
+/// Char and Duration are distinct from Integer so the consumer can emit
+/// dedicated `hew_{format}_object_set_char` / `hew_{format}_object_set_duration`
+/// calls rather than routing through `set_int`.  The binary-wire path
+/// (wireKindOf / encodeFunc) keeps them grouped as varint because the on-wire
+/// representation is identical; only the C ABI name changes.
+///
+/// WHY: Issue #1272 — descriptor op stream (JsonOp::SetChar / SetDuration) must
+/// match the emitted C symbol so a future descriptor-driven Phase-M consumer
+/// can consume the descriptor without re-plumbing the ops.
+/// WHEN obsolete: when the C++ consumer is fully replaced by the descriptor.
+enum class WireJsonKind { Bool, Float32, Float64, String, Bytes, Integer, Char, Duration };
 
 static WireJsonKind jsonKindOf(const std::string &ty) {
   switch (primitiveTypeKind(ty)) {
@@ -136,6 +147,10 @@ static WireJsonKind jsonKindOf(const std::string &ty) {
     return WireJsonKind::String;
   case PrimitiveTypeKind::Bytes:
     return WireJsonKind::Bytes;
+  case PrimitiveTypeKind::Char:
+    return WireJsonKind::Char;
+  case PrimitiveTypeKind::Duration:
+    return WireJsonKind::Duration;
   case PrimitiveTypeKind::I8:
   case PrimitiveTypeKind::I16:
   case PrimitiveTypeKind::I32:
@@ -144,8 +159,6 @@ static WireJsonKind jsonKindOf(const std::string &ty) {
   case PrimitiveTypeKind::U16:
   case PrimitiveTypeKind::U32:
   case PrimitiveTypeKind::U64:
-  case PrimitiveTypeKind::Char:
-  case PrimitiveTypeKind::Duration:
     return WireJsonKind::Integer;
   case PrimitiveTypeKind::Unknown:
     llvm_unreachable(
@@ -1361,6 +1374,8 @@ void MLIRGen::generateWireToSerial(
     std::string rtSetString = "hew_" + format.str() + "_object_set_string";
     std::string rtSetBytes = "hew_" + format.str() + "_object_set_bytes";
     std::string rtSetInt = "hew_" + format.str() + "_object_set_int";
+    std::string rtSetChar = "hew_" + format.str() + "_object_set_char";
+    std::string rtSetDuration = "hew_" + format.str() + "_object_set_duration";
     std::string rtSetNull = "hew_" + format.str() + "_object_set_null";
     std::string rtStringify = "hew_" + format.str() + "_stringify";
     std::string rtFree = "hew_" + format.str() + "_free";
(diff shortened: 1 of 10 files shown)


slepp:
Scope reduced to close #1272 + #1274 only. #1276 remains open for the BMP vs full-Unicode char bounds decision and msgpack-parity migration discussion. Preflight green; cross-ecosystem review surfaces two advisories that trace to #1276 (surrogate handling, plan.rs `char` bounds asymmetry) and a few minor polish items (hardcoded "JSON" in a shared error string, stale BMP-only runtime-doc snippets) worth a small follow-up.

Issue #1290, opened 2026-04-18
lsp-rename: disk walk recurses into symlinked directories without cycle detection; potential DoS on hostile repos


slepp:
The new disk-walk functions in hew-lsp/src/server/navigation.rs:

- `scan_dir_for_conflicts` (lines 1270-1354)
- `collect_unopened_importers_in_dir` (lines 1372-1459)

recurse into directories via `path.is_dir()` without cycle detection. A symlink loop under the workspace would send the rename into unbounded recursion/stack overflow.

The canonical workspace walker at hew-lsp/src/server/workspace.rs:229-245 uses `std::fs::symlink_metadata` to detect symlinks; the new disk walk should do the same or maintain a visited set.

This is a quality issue, not a feature gap: rename should fail safely rather than hang or crash on hostile filesystem layouts.

slepp:
Closed via #1299 (symlink-cycle detection + error propagation in workspace walker).

PR #1299, opened 2026-04-18
fix(lsp-rename): unify disk-scan with workspace walker, surface I/O errors, detect symlink cycles (closes #1287, #1288, #1290)

### LINKED ISSUE #1287: lsp: unify rename disk-scan with workspace walker
The new rename disk-scan path (`scan_dir_for_conflicts` in navigation.rs:1064-1144) duplicates existing workspace-root and file-walk policy already defined in navigation.rs:49-60 (`compute_import_path`) and workspace.rs:225-258 (`collect_hew_files`).

This creates a second place for traversal rules to drift. The proper fix unifies both walkers onto the existing workspace helpers, replacing the custom `find_workspace_root` + `scan_dir_for_conflicts` logic.

**Affected areas:**
- `navigation.rs:1025` — `find_workspace_root` (duplicates navigation.rs:49-60)
- `navigation.rs:1064-1144` — `scan_dir_for_conflicts` (duplicates workspace.rs:225-258)
- Both functions only partially reuse `should_skip_workspace_dir` from workspace.rs:261-265

**Related findings:** consistency[gpt-5.4] from rev2 of fix/lsp-rename-followups (#1285 follow-up).

slepp:
## What

Closes #1287, #1288, #1290.

Three hardening fixes to the LSP rename disk-scan in `hew-lsp`.

---

### #1287 — Unify redundant disk-scan walkers

`navigation.rs` had three separate recursive walker functions
(`scan_dir_for_conflicts`, `collect_unopened_importers_in_dir`,
`find_workspace_root`) that duplicated the logic in `workspace.rs`.

**Fix:**
- Extracted `find_workspace_root_for_uri` and `for_each_hew_file` as
  `pub(super)` helpers in `workspace.rs`.
- `for_each_hew_file` is a generic iterative walker:
  `where E: From<(PathBuf, io::Error)>, F: FnMut(&Path) -> Result<(), E>`.
- Deleted all three ad-hoc functions from `navigation.rs`.
- `collect_hew_files` (workspace-symbol scan) retains its own inline
  best-effort walker — see note below.

**Note on `collect_hew_files`:** workspace-symbol scans are best-effort;
one unreadable entry must not truncate the rest of the results.
`for_each_hew_file` aborts on the first error (by design, for rename),
so `collect_hew_files` stays inline with swallow-on-error semantics.
It now adds an `is_symlink()` skip to match `for_each_hew_file`'s
cycle-safe behaviour.

---

### #1288 — Surface I/O errors instead of silently skipping

The old scanners swallowed `read_dir` and file-read errors, so a
workspace with an unreadable file or directory would silently produce
an incomplete conflict check and potentially approve a broken rename.

**Fix:**
- `for_each_hew_file` propagates `symlink_metadata` and `read_dir`
  errors via the `E: From<(PathBuf, io::Error)>` bound, carrying
  the failing path through so the user-visible message reads
  `"rename failed: /path/to/locked: permission denied"` rather than
  `"rename failed: : permission denied"`.
- Added `RenameError::Io { path, message }` variant (already present)
  with `impl From<(PathBuf, io::Error)> for RenameError` in
  `hew-analysis`.
- `plan_workspace_rename` propagates these errors (`?`); the existing
  `rename_error_to_jsonrpc` handler surfaces them to the client.
- `build_workspace_edit` (called after the plan succeeds) uses
  `.unwrap_or_default()` — a file disappearing between conflict-check
  and edit-build is inherently racy and produces a partial edit, which
  is the least-bad outcome.

**Regression tests (unix-only, skipped under root):**
- `plan_workspace_rename_disk_scan_surfaces_error_for_unreadable_file`
- `plan_workspace_rename_disk_scan_surfaces_error_for_unreadable_directory`

Both assert `RenameError::Io` is returned **and** that the `path` field
is non-empty and names the specific file/directory that failed.

---

### #1290 — Symlink cycle detection

The old scanners used `is_dir()` which follows symlinks, so a
`workspace/loop -> workspace` directory symlink would cause unbounded
recursion / stack overflow.

**Fix:**
- `for_each_hew_file` uses `symlink_metadata` and skips all symlinks
  unconditionally (`metadata.is_symlink()`). No symlinked tree is ever
  entered — neither file nor directory symlinks.
- `collect_hew_files` gets the same `symlink_metadata` + `is_symlink()`
  guard.

**Regression test (unix-only):**
- `plan_workspace_rename_disk_scan_skips_symlinked_directory` — creates
  `workspace/loop -> workspace` and verifies the rename completes
  rather than overflowing.

---

## Commits

1. `fix(lsp-rename): unify disk-scan with workspace walker, surface I/O errors, detect symlink cycles` — main change
2. `fix(lsp-rename): thread path through for_each_hew_file errors, unify collect_hew_files` — path-threading fix so `RenameError::Io.path` is non-empty; strengthened permission tests
3. `fix(lsp-rename): revert collect_hew_files to inline best-effort walker` — restores correct partial-result semantics for workspace-symbol scan

## Testing

- All 22 `plan_workspace_rename` tests pass.
- `make ci-preflight` passes (785 C++ tests, 352 Hew tests, Rust suite).
- `review_worktree` MCP not available in this session (session started
  under `hew/`, not `hew-lang/`). Please run
  `review_worktree(worktree="worktrees/fix-lsp-rename-disk-scan-hardening", base="origin/main", head="HEAD")`
  from a session rooted at `~/projects/hew-lang/` before merging.

## Do NOT enable auto-merge on this PR

### DIFF
--- a/hew-analysis/src/lib.rs
+++ b/hew-analysis/src/lib.rs
@@ -208,6 +208,26 @@ pub enum RenameError {
     InvalidIdentifier { name: String, message: String },
     /// Applying the rename would introduce one or more name clashes.
     Conflicts { conflicts: Vec<RenameConflict> },
+    /// The disk scan encountered an I/O error that prevents a complete
+    /// conflict or edit check.  `path` is the file or directory that
+    /// could not be read; `message` is the OS error string.
+    Io { path: String, message: String },
+}
+
+impl From<std::io::Error> for RenameError {
+    /// Converts a bare I/O error into `RenameError::Io` with an empty path.
+    ///
+    /// Used by the generic `for_each_hew_file` walker, which propagates
+    /// traversal-level I/O errors (e.g. `read_dir` failing on the root)
+    /// before a specific file path is known.  Callers that have a path
+    /// (individual-file reads) construct `RenameError::Io { path, .. }`
+    /// directly.
+    fn from(e: std::io::Error) -> Self {
+        RenameError::Io {
+            path: String::new(),
+            message: e.to_string(),
+        }
+    }
 }
 
 // ── Folding ──────────────────────────────────────────────────────────
--- a/hew-analysis/src/lib.rs
+++ b/hew-analysis/src/lib.rs
@@ -199,6 +199,7 @@ pub enum RenameConflictKind {
 /// must be refused before any edit is produced.
 #[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
 #[serde(tag = "kind", rename_all = "snake_case")]
+#[non_exhaustive]
 pub enum RenameError {
     /// The new name is a language keyword or a builtin identifier that
     /// cannot be shadowed by user code.
--- a/hew-analysis/src/lib.rs
+++ b/hew-analysis/src/lib.rs
@@ -214,17 +214,17 @@ pub enum RenameError {
     Io { path: String, message: String },
 }
 
-impl From<std::io::Error> for RenameError {
-    /// Converts a bare I/O error into `RenameError::Io` with an empty path.
+impl From<(std::path::PathBuf, std::io::Error)> for RenameError {
+    /// Converts a `(path, io_error)` pair into `RenameError::Io`.
     ///
-    /// Used by the generic `for_each_hew_file` walker, which propagates
-    /// traversal-level I/O errors (e.g. `read_dir` failing on the root)
-    /// before a specific file path is known.  Callers that have a path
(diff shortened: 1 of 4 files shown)


slepp:
CONDITIONAL-with-advisory after cross-ecosystem review. Intent PASS. The consistency finding about `compute_import_path` reusing a separate ancestor walk is tracked as follow-up #1300 (narrow refactor, no behavioural change). The `RenameError::Io` URI-conversion use is documented alongside the variant; can be renamed/retyped later if it becomes confusing.

Issue #1393, opened 2026-04-20
hew-analysis: migrate `InMemorySourceDatabase` off `.ok()?` lock poison


slepp:
`hew-analysis/src/db/mod.rs:142, 148, 158` do `self.entries.read().ok()?` / `if let Ok(...)`. If a writer panics holding the lock, every subsequent LSP diagnostic / hover lookup silently returns `None`, surfacing as 'file has no source' with no operator-visible error.

Check-list:
- [ ] Use `PoisonError::into_inner` recovery (or `PoisonSafeRw` if promoted to this crate)
- [ ] Emit `tracing::warn` once on poison recovery
- [ ] Unit test: poison the lock via a panicking writer thread; recovery is visible and subsequent reads succeed

slepp:
Already fixed upstream and merged. Closing.

PR #1423, opened 2026-04-20
fix(analysis): recover from source db rwlock poison (#1393)

slepp:


slepp:
Looks good. Both read and write paths now go through helpers that recover via `PoisonError::into_inner`, the `AtomicBool::swap` guarantees exactly one warn across concurrent poison observers, and the test genuinely poisons the lock by panicking under a held write guard rather than calling `into_inner` directly. Protected state is `Arc`'d immutable `SourceEntry` snapshots, so a mid-insert panic can only leave a missing/stale key — something the LSP layer already tolerates, so continuing after recovery is safe.

One small follow-up worth tracking: `hew-analysis` picks up `tracing = "0.1"` as a direct dep. That is new to this crate; filing a follow-up to consider promoting it to a workspace dep so `hew-analysis` and `hew-runtime` line up.

Issue #1424, opened 2026-04-20
Promote `tracing` to a workspace dependency


slepp:
`hew-analysis` picked up `tracing = "0.1"` as a direct dep in #1423 (poison recovery warn-once). `hew-runtime` already depends on `tracing`. If the workspace is standardising on `tracing` as the log/trace facade, promoting it to a `[workspace.dependencies]` entry keeps versions aligned and makes future adoption a one-line change per crate.

## Check-list
- [ ] Decide whether `tracing` is the canonical facade for the workspace (or if another crate should be preferred)
- [ ] If yes: add `tracing = "0.1"` to `[workspace.dependencies]` in the root `Cargo.toml`
- [ ] Migrate `hew-runtime` and `hew-analysis` to `tracing.workspace = true`
- [ ] Note the policy in `CONTRIBUTING.md` (or equivalent) so future crates inherit the convention

PR #1425, opened 2026-04-20
fix(runtime): distinguish transport lock poison from missing conn (#1394)

### LINKED ISSUE #1394: hew-runtime transport: distinguish poisoned conn table from 'no connection'
`hew-runtime/src/transport.rs:405, 1703` use `self.conns.read().ok()?` and return `None`. Callers treat `None` as a normal disconnect, so a poisoned conn table (post-panic) looks identical to a closed connection: every remote ask/reply silently drops with no diagnostic or reconnect.

`PoisonSafeRw` already exists in `hew-runtime/src/lifetime/poison_safe.rs`.

Check-list:
- [ ] Replace `.ok()?` at both sites with `PoisonSafeRw`
- [ ] Distinguish 'poisoned' vs 'no slot' in the return type
- [ ] Surface poisoned state to supervisor (trace/log + optional escalate)
- [ ] Test: simulate poison; verify supervisor signal + no silent drop

slepp:
## Summary
- fail closed on poisoned transport connection tables with explicit conn lookup errors
- migrate TCP/Unix connection tables to PoisonSafeRw fail-closed accessors and log poison once via tracing
- add a poison regression test that verifies callers observe the poisoned state and receive diagnostics

## Validation
- cargo test -p hew-runtime --lib
- cargo clippy -p hew-runtime --tests -- -D warnings
- cargo fmt --all -- --check
- make ci-preflight

Closes #1394

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1611,6 +1611,8 @@ dependencies = [
  "socket2",
  "tempfile",
  "tokio",
+ "tracing",
+ "tracing-subscriber",
  "ureq",
  "zeroize",
 ]
@@ -2744,6 +2746,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "42b8cfee0e339a0337359f3c88165702ac6e600dc01c0cc9579a92d62b08477a"
 dependencies = [
  "bitflags 2.11.0",
+]
+
+[[package]]
+name = "nu-ansi-term"
+version = "0.50.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "7957b9740744892f114936ab4a57b3f487491bbeafaf8083688b16841a4240e5"
+dependencies = [
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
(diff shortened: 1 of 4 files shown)


slepp:
Reviewed — looks good. The new fail-closed `read()` / `write()` accessors coexist cleanly with the legacy silent-recovery helpers and every transport conn-table callsite (`store_conn`/`get_conn`/`remove_conn` on both TCP and Unix) is migrated to the fail-closed path. Poison no longer collapses into a bare `None`: `get_conn` returns `Err(ConnLookupError::Poisoned { .. })`, the send/recv paths map it to -1 and populate `hew_last_error`, and `report_conn_table_poison` uses an `AtomicBool::swap` so exactly one log line fires regardless of how many threads observe the poison simultaneously. The regression test genuinely panics under a held write guard and asserts the distinguishable error plus warn-once behaviour.

One note, not a blocker: `remove_conn` logs the poison and sets last error but returns `()`, so callers don't observe the distinction directly — consistent with the existing `HewTransportOps` ABI.

`tracing` added to `hew-runtime` follows the same pattern as #1423 and will benefit from the workspace-dep promotion tracked in #1424.

Issue #1391, opened 2026-04-20
std/net/websocket: make `hew_ws_server_accept` and `hew_ws_recv` cancellable


slepp:
`hew_ws_server_accept` calls `TcpListener::accept()` with no side channel; `hew_ws_recv` blocks on recv. If the owning actor is stopped while parked, there is no way to unblock.

`HewWsConn::close_handle` (lib.rs:112–132) already uses the self-pipe / cancel-flag pattern; extend it to server accept and recv.

Check-list:
- [ ] Add self-pipe (or dup-stream + `set_read_timeout` poll loop) to `HewWsServer`
- [ ] Poll cancel flag in recv path
- [ ] Drop handler triggers cancel + joins the accept/recv thread
- [ ] Test: stop owning actor while accept is parked; completes within the cancel SLO

slepp:
Already resolved by #1426 (commit 41651ea2, 2026-04-20). All four checklist items are in `std/net/websocket/src/lib.rs`:

- `HewWsServer` runs `set_nonblocking(true)` and a `WouldBlock` poll loop driven by a cancel flag.
- `recv_message()` polls `inner.closed` each iteration with a `READER_READ_TIMEOUT` bound.
- `HewWsServer::close_handle()` sets the cancel and drains `active_accepts`; the `Drop` impl invokes it.
- Tests `server_accept_cancel_returns_cancelled`, `recv_cancel_returns_cancelled`, and `server_accept_unblocks_when_owner_actor_stops` all pass (29/29 in `cargo test -p hew-std-net-websocket`).

Closing as completed.

PR #1426, opened 2026-04-20
fix(std/net/websocket): make accept and recv cancellable (#1391)

slepp:


slepp:
Reviewed — looks good. Both `HewWsConnInner` and `HewWsServerInner` are `Arc`-shared so in-flight accept/recv hold their own clones and cannot dangle, `ActiveCallGuard` decrements along every return path, and close uses a bounded `wait_for_active_calls_to_drain` (500ms) that falls through to detach rather than `join()`-ing unconditionally. Cancellation is driven by atomics plus the 250ms socket read timeout and 10ms non-blocking accept backoff — no indefinite paths. Tests cover accept-cancelled-pre-client, recv-cancelled-mid-read, and owner-actor-stop-during-accept, all polled with 10ms granularity (no long sleeps). No new `.ok()` / `unwrap_or_default()` on the cancellation paths, no banned strings, no FFI signature changes.

One small follow-up worth a separate issue: on the `close_handle` path where the reader has already exited but recv hasn't, `drop_ws` is skipped and the graceful WebSocket close frame isn't sent. It's benign today — `shutdown_socket(Both)` has already torn the TCP connection down and the `WebSocket` is released when the last `Arc` drops — but if a clean close frame on the slow-recv path matters, it's a one-line fix.

Issue #1390, opened 2026-04-20
std/net/http: track response-writer threads for cancel + drop


slepp:
`std/net/http/src/server.rs:570` spawns a thread per streaming response via `std::thread::spawn` and discards the `JoinHandle`. `hew_http_server_close` (line 596) drops the `Box<HewHttpServer>` without signalling writers or joining them, and there is no `impl Drop for HewHttpServer`. In-flight writers keep holding accepted connections with no cancellation channel.

Check-list:
- [ ] Track spawned response threads on the server handle
- [ ] On close/drop: signal cancel + join with bounded timeout
- [ ] Add `impl Drop for HewHttpServer`
- [ ] Test: close with an in-flight streaming response; connection torn down promptly

slepp:
Already resolved by #1427 (commit 4b0982ff, 2026-04-20). All four checklist items are in `std/net/http/src/server.rs`:

- **Track spawned response threads on the server handle** — `ResponseThreadTracker` (line 61) holds an `Arc<Mutex<Vec<TrackedResponseThread>>>` plus a shared `cancel: Arc<AtomicBool>` and `active_count: Arc<AtomicUsize>`. Both `spawn_response` and `spawn_body_read` register through it.
- **On close/drop: signal cancel + join with bounded timeout** — `cancel_and_join_all` sets the cancel flag and joins each thread with a 2-second bound, detaching on timeout. `ChannelReader::read` polls the cancel flag on every `recv_timeout(50ms)` iteration so writers unblock promptly.
- **`impl Drop for HewHttpServer`** — line 308; invokes `cancel_and_join_all()`.
- **Test: close with an in-flight streaming response** — covered by `server_close_cancels_streaming_responses_and_drains_threads`, `server_close_cancels_concurrent_streaming_responses_and_reaps_completed_threads`, and `server_close_detaches_hung_response_thread_after_timeout`. All 109 `hew-std-net-http` tests pass.

Closing as completed.

PR #1427, opened 2026-04-20
fix(std/net/http): track and drain response threads on server close (#1390)

slepp:


slepp:
Revision looks good. Visibility-expansion revert is complete — diff is now a single file (`std/net/http/src/server.rs`), and `hew-runtime/` / `hew-analysis/` / `hew-observe/` / `Makefile` / `scripts/` / `.github/` show no changes against `origin/main`. All five tracker `.lock()` sites use `.unwrap_or_else(PoisonError::into_inner)`; no `.unwrap()` / `.ok()` / `.unwrap_or_default()` regressions.

The concurrent-streams test spawns four clients, drives one to completion, and asserts `reap_completed` drops exactly that one from the tracked count while leaving the other three active — the separation is tested, not just the drain. The detach test hangs a thread on a release-flag `park_timeout`, asserts `close()` returns within `RESPONSE_THREAD_JOIN_TIMEOUT + 400ms`, and confirms `tracked_thread_count == 0` while `active_count == 1` (distinguishing detach from join); it releases the hung thread before exiting so it can't leak into other tests.

A tiny nit, not blocking: the detach test hand-rolls an `ActiveCountGuard`-shaped shim. If the production guard changes, the test's version silently diverges — a brief comment pointing at the real guard would help future maintainers.

PR #1428, opened 2026-04-21
hew-cli: replace unwrap_or_default probe swallows in watch/build/adze

### LINKED ISSUE #1400: hew-cli: replace `unwrap_or_default` probe swallows in watch/build/adze
Several hew-cli / adze-cli config and build paths fall back to defaults silently:
- `hew-cli/src/watch.rs:447` skips unreadable directories with `read_dir(path).ok()?` — no log
- `hew-cli/build.rs:103, 347, 357` parse `HOST` / target triples with `unwrap_or_default()`; an empty triple cascades into LLVM-dir detection
- `adze-cli/src/config.rs:104` swallows malformed `adze.toml` into a default, ignoring user intent

Check-list:
- [ ] Log at `warn!` on recoverable failure
- [ ] build.rs: empty HOST is a hard error
- [ ] adze config: surface parse error, do not silently default
- [ ] watch: explicit opt-in for 'continue on unreadable dir'

slepp:
This updates CLI and adze config probe handling to fail closed instead of swallowing probe errors with unwrap_or_default in watch/build/adze paths.

Closes #1400

### DIFF
--- a/adze-cli/src/client.rs
+++ b/adze-cli/src/client.rs
@@ -173,8 +173,17 @@ impl RegistryClient {
     /// Loads `~/.adze/config.toml` to check for a `fallback-api` override.
     #[must_use]
     pub fn new() -> Self {
+        let cfg = config::load_config().unwrap_or_else(|error| {
+            eprintln!("adze: {error}");
+            std::process::exit(1);
+        });
+        Self::new_with_config(&cfg)
+    }
+
+    /// Create a client using the provided config.
+    #[must_use]
+    pub fn new_with_config(cfg: &config::AdzeConfig) -> Self {
         let endpoints = config::discover_registry();
-        let cfg = config::load_config();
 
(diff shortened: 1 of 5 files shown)


slepp:
Took a pass through this. Reads well —

A few small notes, none blocking:


2. `chrono_like_timestamp` in watch.rs still has an `unwrap_or_default()` on `duration_since(UNIX_EPOCH)`. It's pre-existing and not a config probe, so out of scope for #1400 — just flagging in case you want to sweep it in a follow-up (it'll silently render `00:00:00` if the clock is ever before the epoch, which is admittedly unlikely).

3. Tests cover the new error surfaces (missing file → default, malformed TOML → `Parse` with path, empty `HOST` triple, missing watch dir). All are deterministic filesystem unit tests under `tempdir`, so flake risk looks low.

Nothing here that needs to block. Happy to have this go in once CI lights up.

PR #1429, opened 2026-04-21
hew-cli: surface rustc and xcrun linker probe failures

### LINKED ISSUE #1396: hew-cli linker probes: surface rustc / xcrun failures
`hew-cli/src/link.rs:366, 368, 568, 572` use `.ok()?` on `rustc --print sysroot` / `xcrun --show-sdk-path`. Every failure mode (tool missing, non-zero exit, non-UTF-8 output) is silently discarded; the linker then skips the SDK/libc and reports a confusing 'symbol not found' later.

Check-list:
- [ ] `find_wasi_libc` returns `Result<Option<PathBuf>, LinkerProbeError>`
- [ ] Same for `find_macos_sdk`
- [ ] Caller maps each error variant to a targeted linker diagnostic
- [ ] Unit test: `rustc` missing / returns non-UTF-8 / returns non-zero — each produces a distinct error

slepp:
Summary
- improve linker probe diagnostics in hew-cli/src/link.rs so rustc and xcrun invocation failures are surfaced with actionable context
- wire richer linker probe failure propagation through hew-cli/src/watch.rs so watch mode reports probe failures instead of masking them

Why
The CLI previously hid underlying probe execution failures, which made linker setup issues hard to debug on macOS. This change surfaces the real rustc/xcrun failure details directly at the CLI boundary.

Verification
- git diff origin/main --stat (scope limited to hew-cli/src/link.rs and hew-cli/src/watch.rs)
- make ci-preflight (green)

Closes #1396

### DIFF
--- a/hew-cli/src/link.rs
+++ b/hew-cli/src/link.rs
@@ -4,6 +4,101 @@
 
 use crate::target::TargetSpec;
 
+#[derive(Debug)]
+enum LinkerProbeError {
+    CommandUnavailable {
+        probe: &'static str,
+        command: String,
+        source: std::io::Error,
+    },
+    CommandFailed {
+        probe: &'static str,
+        command: String,
+        status: std::process::ExitStatus,
+        stderr: String,
+    },
+    NonUtf8Stdout {
+        probe: &'static str,
+        command: String,
+    },
+    EmptyStdout {
+        probe: &'static str,
+        command: String,
+    },
+    MissingArtifact {
+        probe: &'static str,
+        artifact: &'static str,
+        tried: Vec<std::path::PathBuf>,
+    },
+    NoSuitableTool {
+        probe: &'static str,
+        tool_kind: &'static str,
+        tried: Vec<String>,
+    },
+}
+
+impl std::fmt::Display for LinkerProbeError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            Self::CommandUnavailable {
+                probe,
+                command,
+                source,
+            } => {
+                write!(f, "{probe} failed: could not run `{command}`: {source}")
+            }
+            Self::CommandFailed {
+                probe,
+                command,
+                status,
+                stderr,
+            } => {
+                write!(f, "{probe} failed: `{command}` exited with status {status}")?;
+                if !stderr.is_empty() {
+                    write!(f, " ({stderr})")?;
+                }
+                Ok(())
+            }
+            Self::NonUtf8Stdout { probe, command } => {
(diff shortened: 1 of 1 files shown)


slepp:
Review notes

Scope-audit: diff is limited to hew-cli/src/link.rs and hew-cli/src/watch.rs. No silent additions or removals.

Correctness: The probe error surface is actually propagated. find_darwin_sdk, find_wasi_libc, find_wasm_ld, and the new probe_command_stdout/run_probe_command helpers all return structured LinkerProbeError variants that carry probe name, command string, exit status, stderr, and the candidate list where relevant. Call sites propagate via ? with a target-qualified prefix; none of the probe paths swallow into .ok()? or unwrap_or_default. The remaining .ok()? in watch.rs::find_first_hew_file is on a directory-scan path unrelated to linker probes and is marked JUSTIFIED with a rationale, which is fine.

Behavior change worth calling out (not a blocker): find_wasi_libc now returns Err(MissingArtifact) instead of Ok(None) when libc.a is absent. link_wasm propagates that as a fatal error where it previously silently skipped linking WASI libc. Since hew only targets wasm32-wasi/wasm32-wasip1 (target.rs), a missing self-contained/libc.a genuinely is a toolchain misconfiguration and surfacing it matches the intent of #1396. The outer if let Some(..) wrapper at the call site is now structurally dead but harmless.

Drop safety: No new long-lived subprocess handles or file handles; Command::output() runs to completion before return. Nothing to audit under CLAUDE.md #1.

Cross-platform: find_wasm_ld keeps the existing apt.llvm.org, FreeBSD pkg, and Windows Program Files candidates and now reports all tried paths on failure. find_macos_sdk stays behind #[cfg(target_os = "macos")] with the non-macOS arm returning Ok(None). No regression on Linux/FreeBSD/Windows coverage.

Tests: Five new unit tests exercise run_probe_command and find_wasm_ld_with via injected closures — no process spawn, no filesystem, no tempdirs. Deterministic by construction, so flake risk is effectively zero and a 3× gate is not meaningful here. The preflight flake the implementer saw is not introduced by this diff.

Voice: PR body and commit message read as plain engineering prose.

No blockers. Approving and enabling auto-merge once CI clears.

slepp:
Rebased this branch onto latest main and resolved the watch.rs rebase conflict. Re-ran workspace-wide checks (cargo fmt --all -- --check, cargo clippy --workspace --tests -- -D warnings) and make ci-preflight; all are clean locally.

slepp:
Addressed the remaining clippy -D warnings issue by removing the unnecessary Result wrapper in the Darwin SDK probe and re-ran clippy/tests, fmt check, and ci-preflight cleanly.

PR #1431, opened 2026-04-21
hew-codegen: reject dyn-trait dispatch for generic methods

### LINKED ISSUE #1412: hew-codegen: silent miscompile on dyn-trait dispatch for generic methods
`hew-codegen/src/mlir/MLIRGen.cpp:4856–4872` — the dyn-trait shim generator iterates `traitIt->second.methods`. For any method whose `type_params` is present and non-empty, the body is a `CODEGEN-TODO` comment followed by a bare `continue;` (line 4860). The vtable is then built without the skipped slot; a user calling a generic method through `dyn Trait` binds to the wrong slot or crashes on a null vtable entry at runtime. No `emitError` — this is a silent miscompile, not an unsupported-feature error.

Distinct from #1399 (drop semantics, not dyn-dispatch type arguments).

## Check-list
- [ ] Replace the bare `continue;` with `emitError(loc) << "dyn-trait dispatch does not yet support generic methods: " << tm->name;` and `return failure();`
- [ ] Integration test: a trait with a generic method used via `dyn Trait` surfaces a clean diagnostic

slepp:
## Summary
- fail closed when coercing to explicit dyn Trait if the trait includes generic methods, instead of generating invalid dispatch shims
- plumb an explicit dyn-parameter attribute so this validation only triggers for explicit dyn-typed call sites
- add dyn_trait_generic_method_reject.hew as a negative e2e test and register it in codegen CMake tests

Surfaced during preflight: lazily generated dyn dispatch shims could be emitted under the wrong module path and miss impl symbols; fixed by storing impl module path in dispatch metadata and restoring it during shim generation; covered by existing codegen e2e tests (including e2e_concurrency/channel_for_await_typed_receiver_helper.hew).

Closes #1412

### DIFF
--- a/hew-codegen/include/hew/mlir/MLIRGen.h
+++ b/hew-codegen/include/hew/mlir/MLIRGen.h
@@ -466,6 +466,7 @@ private:
 
   /// Line map from serialized AST: byte offset of each line start.
   std::vector<size_t> lineMap_;
+  std::string sourcePath_;
 
   /// Convert a byte offset to (line, column), both 1-based.
   /// Returns (0, 0) if lineMap_ is empty (no debug info).
@@ -1230,6 +1231,7 @@ private:
   // ── dyn Trait dispatch infrastructure ────────────────────────────
   struct TraitImplInfo {
     std::string typeName;
+    std::vector<std::string> modulePath;
     std::string vtableName;                 // e.g. "__vtable_HT3DogF8Greetable"
     std::vector<std::string> shimFunctions; // shim function names in vtable order
   };
@@ -1239,6 +1241,7 @@ private:
   };
   // traitName → dispatch info
   std::unordered_map<std::string, TraitDispatchInfo> traitDispatchRegistry;
+  std::unordered_set<std::string> explicitDynTraitUses;
   // Track dyn-trait variable types: varName → traitName
   std::unordered_map<std::string, std::string> dynTraitVarTypes;
 
@@ -1257,14 +1260,16 @@ private:
                          const std::vector<std::string> &methodNames);
 
   /// Generate vtable dispatch shim functions for a (type, trait) pair.
-  void generateTraitImplShims(const std::string &typeName, const std::string &traitName);
+  void generateTraitImplShims(const std::string &typeName, const std::string &traitName,
+                              mlir::Location location);
 
   /// Generate a single dyn dispatch shim function.
   void generateDynDispatchShim(const std::string &implFuncName);
 
   /// Coerce a concrete struct value to a dyn Trait fat pointer {data_ptr, vtable_ptr}.
   mlir::Value coerceToDynTrait(mlir::Value concreteVal, const std::string &typeName,
-                               const std::string &traitName, mlir::Location location);
+                               const std::string &traitName, mlir::Location location,
+                               bool rejectGenericMethods = false);
 
   // ── Generics monomorphization ──────────────────────────────────
   // Registry of generic (unspecialized) function declarations.
--- a/hew-codegen/include/hew/mlir/MLIRGen.h
+++ b/hew-codegen/include/hew/mlir/MLIRGen.h
@@ -1242,6 +1242,7 @@ private:
   // traitName → dispatch info
   std::unordered_map<std::string, TraitDispatchInfo> traitDispatchRegistry;
   std::unordered_set<std::string> explicitDynTraitUses;
+  bool explicitDynTraitScanFailed = false;
   // Track dyn-trait variable types: varName → traitName
   std::unordered_map<std::string, std::string> dynTraitVarTypes;
 
--- a/hew-codegen/src/mlir/MLIRGen.cpp
+++ b/hew-codegen/src/mlir/MLIRGen.cpp
@@ -38,9 +38,12 @@
 
 #include <algorithm>
 #include <cassert>
+#include <fstream>
 #include <functional>
 #include <iostream>
 #include <map>
+#include <regex>
+#include <sstream>
 #include <string>
 
 using namespace hew;
@@ -49,6 +52,7 @@ using namespace mlir;
 namespace {
 
 constexpr llvm::StringLiteral kHewDebugParamNameAttr = "hew.debug.param_name";
+constexpr llvm::StringLiteral kHewExplicitDynParamAttr = "hew.explicit_dyn_param";
 
 bool isQualifiedStdlibPointerBackedHandleName(llvm::StringRef name) {
   return name == "stream.Stream" || name == "stream.Sink" || name == "stream.StreamPair" ||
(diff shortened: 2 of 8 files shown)


slepp:
## Review: scope-expansion verdict and correctness findings

**Scope-expansion verdict: defensible, not splittable.** The "module path restoration" fix is not an independent regression — it's a direct consequence of the design choice to generate dyn-dispatch shims lazily inside `coerceToDynTrait` instead of eagerly inside `generateImplDecl`. Generating shims lazily moves them outside the impl's module context, so `currentModulePath` has to be saved/restored. The two fixes are causally coupled within this design; splitting makes no sense. (That said, an alternative design — keep eager shim gen, just skip generic methods and reject only at `coerceToDynTrait` — would have avoided the extra scope entirely. That's an approach comment, not a blocker.)

### Correctness findings (primary fix)

**1. Regex-scanning source text to detect explicit `dyn Trait` uses is fragile and fail-open.** `MLIRGen.cpp` now re-opens `sourcePath_` and runs `std::regex(R"(\bdyn\s+([A-Za-z_][A-Za-z0-9_]*)\b)")` over the raw bytes to populate `explicitDynTraitUses`. Issues:

- If `sourcePath_` cannot be opened (empty path, in-memory/stdin compile, CWD mismatch) the set is silently empty; `rejectGenericMethods` then becomes a no-op and the generic-method rejection never fires. Per CLAUDE.md §2 (fail-closed codegen), this is the exact fail-open shape we're supposed to avoid — the guard against the original #1412 miscompile evaporates whenever the file read fails.
- The pattern does not match `dyn (A + B)` composite trait objects (the `(` kills the `\s+[A-Za-z_]` lookahead). `multi_dyn.hew` is currently excluded from CMake for unrelated reasons, but any future composite-dyn case would silently slip the check.
- It matches `dyn Foo` inside comments and string literals. Narrow impact (requires an actual explicit-dyn callee for the same name to turn into behavior change), but still noise in an invariant that CLAUDE.md #2 asks to be precise about.
- The compiler already has the AST. Walking `ast::TypeTraitObject` nodes during a normal pass gives you this information deterministically and per-module. Please replace the regex scan with an AST walk (or drop the "explicit vs. implicit" gating entirely — see below).

**2. The "explicit vs. implicit dyn" gating may not be buying anything.** Hew's surface syntax always writes `dyn T` explicitly at the use site (function parameter, let-binding type, field type, return type). There is no implicit dyn coercion path I can find. If that's correct, the `explicitDynTraitUses` set and the `hew.explicit_dyn_param` attr are dead weight — `coerceToDynTrait` could just check whether the trait has any generic methods and unconditionally reject. That's a simpler, non-fragile fix that also removes the motivation for the lazy-shim rework. Worth reconsidering before merging this design.

**3. Lazy shim generation is gated narrowly — coerceType path not covered.** `generateTraitImplShims` now runs only from `coerceToDynTrait` and only when `rejectGenericMethods && explicitDynTraitUses.count(traitName)`. The generic coerceType callsite at `MLIRGen.cpp:1302` calls `coerceToDynTrait(..., location)` with the default `rejectGenericMethods=false`, so lazy shim gen never fires there. For the current test suite this is fine — all dyn coercions route through the explicit-param path in `MLIRGenExpr.cpp:1717` — but any future coercion going through `coerceType` (let-bound dyn variable, dyn struct field init, dyn return value, dyn array element) will build a vtable that references shim symbols whose function bodies were never emitted, producing a link-time failure. The comment block at ~3487 ("Pass 1k-trait … ensures `generateTraitImplShims` can overwrite the stale shim names") is now stale — the pass no longer generates shims.

**4. `currentModulePath` leak on error path in `generateTraitImplShims`.** When the generic-method error is emitted, the function `return`s early without restoring `currentModulePath`:

```cpp
auto savedModulePath = currentModulePath;
if (implInfoPtr) currentModulePath = implInfoPtr->modulePath;
...
    emitError(location) << "dyn-trait dispatch does not yet support generic methods: " << tm->name;
    return;               // ← currentModulePath not restored
...
currentModulePath = savedModulePath;  // only reached on success
```

Codegen continues running after `++errorCount_` (errors are accumulated, not immediately fatal), so downstream name-mangling etc. will observe the wrong module path. Either wrap in a scope-guard or restore before each `return`.

### Process

- **Flake gate:** no evidence that the new e2e_negative test `dyn_trait_generic_method_reject` was run 3× to rule out flake. Please attach a local 3× pass before merging.
- **WASM parity (CLAUDE.md §4):** this is a codegen change affecting MLIRGen, which feeds both native and WASM backends. No WASM-targeted assertion was added for the new negative test. If rejection behavior is backend-independent (it is, since the error is raised pre-lowering in `coerceToDynTrait`), a single-line `// WASM-TODO: rejection is pre-lowering, backend-agnostic` in the test or PR description is enough to close the loop.

### Voice

Commit message and PR body are plain — no lane/wave/model/PASS labels. Good.

### Verdict

Request revision. The primary finding is (1) — a fail-open path that reintroduces silent-miscompile risk under common-enough conditions (any compilation where the source file isn't re-readable from `sourcePath_`). (4) is small but real. (2) is worth pushing back on before locking in this machinery. Not blocking outright because current tests do cover the happy path correctly, but I would not auto-merge this as-is.

slepp:
Addressed all four findings:\n\n1) Fail-open regex scan\n- Made source scan fail-closed for the generic-dyn guard: if the source path is unreadable and the target trait has generic methods, codegen now emits a diagnostic and rejects compilation instead of silently treating it as no-match.\n- Kept non-generic dyn paths working (no false compile break for REPL/stdin).\n- Extended scan coverage to composite forms dyn (A + B) by extracting trait names from composite dyn groups.\n- Expanded the negative fixture to use composite dyn syntax in dyn_trait_generic_method_reject.hew.\n\n2) coerceType bypass\n- Wired the same lazy shim generation gate through the generic coerceType dyn coercion path (coerceToDynTrait with rejectGenericMethods=true), so non-call-site coercions also materialize shims and reject generic trait methods correctly.\n- Added positive e2e e2e_dyn_trait/dyn_coerce_type_paths.hew covering dyn coercion via let-binding, struct field init, return value, and array element coercion.\n\n3) currentModulePath leak\n- Restored currentModulePath before the early return on generic-method rejection in generateTraitImplShims, so subsequent codegen is not contaminated after diagnostics.\n\n4) Flake gate + WASM parity\n- Flake gate run (default threading) for changed tests: dyn_trait_generic_method_reject + dyn_coerce_type_paths, pass/pass across 3 consecutive runs (3/3).\n- Added explicit WASM parity note in the negative test: rejection occurs in MLIRGen pre-lowering, so behavior is backend-agnostic and does not require a separate wasm-only assertion.

slepp:
Re-review of the fix commits (525405b, 7e21cce, c26d052, 7427451).

Prior findings status:

1. Regex fail-closed — resolved. `sourcePath_` open failure sets `explicitDynTraitScanFailed`, and `coerceToDynTrait` with `rejectGenericMethods=true` emits a diagnostic and returns null when that flag is set and the trait has any generic method. Composite pattern `\\bdyn\\s*\\(([^)]*)\\)` captures `dyn (A + B)` and `dyn (A+B+C)`; the inner `traitNamePattern` pulls out each identifier and inserts it into `explicitDynTraitUses`. The non-paren form `dyn A + B` only captures A, but that is not the composite syntax the prior finding called out.

2. coerceType path — resolved. MLIRGen.cpp:1302 now calls `coerceToDynTrait(..., /*rejectGenericMethods=*/true)` for every struct→dyn coercion that flows through `coerceType`. The new `dyn_coerce_type_paths` e2e exercises let-binding, struct field, function return, and array-element coercions in a single program; output matches `let/field/return/array`.

3. currentModulePath restore — resolved. `generateTraitImplShims` saves `currentModulePath` before overwriting it and restores it on both the generic-method error return and the normal end-of-function path. No other early returns inside the function after the save point.

4. Flake gate / WASM parity — resolved. Locally `e2e_dyn_trait_dyn_coerce_type_paths` and `reject_e2e_negative_dyn_trait_generic_method_reject` both pass 3/3 after rebuilding the Rust front-end and the C++ codegen. MLIRGen is the single entry point for all targets (codegen_capi.cpp instantiates one `MLIRGen` regardless of target triple), so the pre-lowering rejection is backend-agnostic and the WASM-TODO note is accurate.

Scope is within hew-codegen only. No new critical findings. Enabling auto-merge.

PR #1432, opened 2026-04-21
hew-runtime: surface FreeBSD argv sysctl failure

### LINKED ISSUE #1401: hew-runtime/env.rs: surface FreeBSD argv sysctl failure
`hew-runtime/src/env.rs:138` — `read_process_args` on FreeBSD calls `freebsd_args().unwrap_or_default()`. If sysctl fails (permissions, ENOMEM, sandbox), the process silently sees an empty argv. Linux and macOS return real argv via `std::env::args()`; this is a FreeBSD-only wrong-answer path with no log.

Check-list:
- [ ] On sysctl error, fall back to `std::env::args()`
- [ ] Emit `tracing::warn` once at startup on the fallback
- [ ] Inline comment documents the decision

slepp:
Closes #1401

### DIFF
--- a/hew-runtime/src/env.rs
+++ b/hew-runtime/src/env.rs
@@ -7,6 +7,8 @@
 use crate::cabi::malloc_cstring;
 use crate::lifetime::PoisonSafeRw;
 use std::ffi::{c_char, CStr};
+#[cfg(any(target_os = "freebsd", test))]
+use std::sync::atomic::{AtomicBool, Ordering};
 
 /// Global lock for environment variable access synchronization.
 ///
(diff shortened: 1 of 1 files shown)


slepp:
Reviewed at a0e70dc.

Scope is clean: only hew-runtime/src/env.rs changed (+378/-50).


Cross-platform: real sysctl code stays behind `cfg(target_os = "freebsd")`; the trait + error + parser are `cfg(any(target_os = "freebsd", test))` so the injected tests compile on macOS/Linux/Windows without touching the non-FreeBSD runtime path. macOS / Linux / OpenBSD still hit the plain `std::env::args()` branch unchanged.

Drop safety: buffer is a `Vec<u8>`; every early return drops it normally, no raw alloc.

Tests: three injected tests (happy parse, truncation surfaces `TruncatedBuffer`, sysctl failure triggers fallback + populates last_error + logs exactly once) plus one `cfg(target_os = "freebsd")`-gated real-sysctl test. Ran the injected set 3× on macOS — 3/3 green, no flakes.

Minor note (not blocking): the fallback uses `std::env::args()`, which on FreeBSD reads argv captured at process init via `.init_array` rather than re-calling `kern.proc.args`, so it's a genuinely independent fallback path. Worth keeping in mind if stdlib behavior ever changes.

No blocking issues. Enabling auto-merge.

PR #1434, opened 2026-04-21
Fix LSP code action drift warnings

### LINKED ISSUE #1421: hew-lsp: code-action suggestions silently drop on malformed JSON
`hew-lsp/src/server/handlers/language_features.rs:108–113` — `diag.data.as_ref().and_then(|d| d.get("suggestions")).and_then(|v| serde_json::from_value::<Vec<String>>(v.clone()).ok()).unwrap_or_default()`. If a diagnostic encodes suggestions with a different shape (e.g. `Vec<Suggestion>` after a schema bump), every code action silently lists zero suggestions. Producer (`hew-analysis`) and consumer (`hew-lsp`) live in the same workspace, but there is no deserialization-failure log or assertion to catch drift.

## Check-list
- [ ] Replace `.ok()` with a `match` that logs at `warn!(target = "hew-lsp", …)` on deserialization failure
- [ ] Preserve the empty-suggestion fallback but make drift visible
- [ ] Add a consumer-side unit test that feeds a shape-drift payload and asserts the warn log

slepp:
## Summary
- warn and fail closed when code-action suggestion payloads cannot be deserialized
- return an explicit empty code-action response when the document is missing or no actions match
- add code-action tests that assert the new warning and empty-response behavior

## Testing
- cargo test -p hew-lsp --quiet (x3)
- cargo clippy -p hew-lsp --all-targets -- -D warnings
- cargo fmt --all -- --check
- make ci-preflight

Closes #1421

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1563,6 +1563,8 @@ dependencies = [
  "serde_json",
  "tokio",
  "tower-lsp",
+ "tracing",
+ "tracing-subscriber",
 ]
 
 [[package]]
--- a/hew-lsp/Cargo.toml
+++ b/hew-lsp/Cargo.toml
@@ -13,6 +13,7 @@ categories = ["compilers", "development-tools"]
 workspace = true
 
 [dependencies]
+tracing = "0.1"
 hew-analysis = { path = "../hew-analysis" }
 hew-lexer = { path = "../hew-lexer" }
 hew-parser = { path = "../hew-parser" }
@@ -21,3 +22,6 @@ tower-lsp = "0.20"
 tokio = { version = "1", features = ["full"] }
 serde_json = "1"
 dashmap = "6"
+
+[dev-dependencies]
+tracing-subscriber = { version = "0.3", features = ["fmt"] }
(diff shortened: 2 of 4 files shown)


slepp:
Reviewed the diff against origin/main.

Scope: only `hew-lsp/**` plus the `Cargo.lock` entry for the newly-added `tracing` / `tracing-subscriber` deps. Nothing snuck in outside the LSP crate.

Correctness — the two silent skips are gone:


Tests: the three new tests each exercise the new behaviour rather than a happy path:
- `code_actions_warn_and_return_empty_when_suggestions_drift` feeds a structurally-wrong suggestions payload (`[{"replacement": "counter"}]`) and asserts both empty actions *and* the presence of the `WARN` / `hew-lsp` / `failed to deserialize code-action suggestions` log line.
- `code_action_returns_empty_response_and_warns_for_unknown_document` drives `code_action_response` with an empty `DashMap` and asserts the unknown-document warn line.
- `code_action_returns_explicit_empty_response_when_no_actions_match` confirms the happy path stays quiet (no warn emitted when a normal request just has nothing to offer). The `SharedLogWriter` / `capture_logs` harness is test-only and does not leak into production state.

Concurrency / drop safety: no new request-scoped `Arc` / `Mutex` on the server side — the only new shared state is inside the test harness. Nothing to clean up on client cancellation.

Voice: PR body and commit message read as plain engineering prose.

Enabling auto-merge (squash) pending the in-flight clippy / build / coverage checks.

PR #1435, opened 2026-04-21
fix(std/crypto/jwt): surface tagged JWT C-ABI errors

### LINKED ISSUE #1433: std/crypto/jwt: C-ABI collapses every error into null pointer
## Problem

The JWT C-ABI collapses every error class — unknown algorithm, missing key, malformed token, signature mismatch, expired claims — into the same "returned null pointer" sentinel. Callers on the Hew side see `null` and cannot distinguish "algorithm unsupported" from "signature invalid" from "allocator OOM".

See `std/crypto/jwt/src/lib.rs:58-124` (`hew_jwt_encode_hs*`, `hew_jwt_decode_hs*` — each `match` arm returns `std::ptr::null_mut()` with no diagnostic trail).

## Why this is wrong (fail-closed)

Per `CLAUDE.md` §2, silent `None` / null returns at an FFI boundary mask real bugs. Here the collapse is worse than a refactoring issue — it also makes JWT verification failures indistinguishable from library misconfiguration, which is a security-relevant category confusion.

## Proposed fix

- Adopt the same pattern used for `hew_http_*` in #1392 / PR #1430: an out-param `*mut HewError` (or a tagged-enum i32 return) that distinguishes:
  - `AlgorithmUnsupported`
  - `InvalidKey`
  - `TokenMalformed`
  - `SignatureInvalid`
  - `ClaimsExpired`
  - `AllocationFailure`
- Keep the successful-string-returning shape; only null-on-failure callers break, and they already have no way to interpret null.
- Add unit tests that assert each error class produces a distinct tag, not just "null".

## Scope

- `std/crypto/jwt/src/lib.rs` (the C-ABI boundary)
- `std/crypto/jwt/jwt.hew` only if the Hew binding needs updating for the new out-param

Discovered in R6 workspace quality audit (F1, P0).

slepp:
This updates the JWT C-ABI boundary so null returns carry an explicit error discriminator instead of collapsing every failure into the same sentinel.

What changed:
- added a `#[repr(C)]` `HewJwtError` tag for `AlgorithmUnsupported`, `InvalidKey`, `TokenMalformed`, `SignatureInvalid`, `ClaimsExpired`, and `AllocationFailure`
- updated the encode/decode C-ABI exports to take an `err_out` parameter and write it on every failure path before returning null
- kept Hew-facing wrappers by routing them through compatibility shims and exposing the new error enum plus `try_encode`/`try_decode` helpers in `jwt.hew`
- added focused JWT tests covering the success path and each error class

The tests exercise both encode and decode classification, including malformed input, wrong secret, expired claims, unsupported algorithms, and forced allocation failure paths.

Closes #1433

### DIFF
--- a/std/crypto/jwt/jwt.hew
+++ b/std/crypto/jwt/jwt.hew
@@ -1,6 +1,6 @@
 //! JSON Web Token encoding and validation.
 //!
-//! Encode, decode, and validate JWTs using HMAC or RSA algorithms.
+//! Encode, decode, and validate JWTs using HMAC algorithms.
 //!
 //! # Examples
 //!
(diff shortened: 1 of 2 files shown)


slepp:
Reviewed the C-ABI error tagging end-to-end. Summary: the fail-closed invariant holds across every null-return path, scope is tight, and the classifier is security-sane. Two non-blocking observations below.

**Scope.** Only `std/crypto/jwt/src/lib.rs` and `std/crypto/jwt/jwt.hew`. No scope creep.

**Per-function err_out audit** (`hew_jwt_encode`, `hew_jwt_decode`):
- Both unconditionally write `HewJwtError::None` on entry, then on every subsequent error branch (input null → TokenMalformed/InvalidKey; unknown algo → AlgorithmUnsupported; jsonwebtoken err → classified kind; malloc failure → AllocationFailure) write the tag before returning null. No silent null-returns.
- `write_err_out` null-checks `err_out` before deref, so a null out-param is safe by inspection.
- The `_hew` compat shims route through the new exports using a stack-local `err`, then stash into TLS for `hew_jwt_last_error_code()`. Clean.

**Classifier (`classify_jwt_error`).** Correct security posture: `ErrorKind::InvalidAlgorithm` (token alg ≠ expected alg) is folded into `SignatureInvalid`, not `AlgorithmUnsupported`. That's the right call — it avoids letting an attacker enumerate supported algorithms via error class. Only `InvalidAlgorithmName`/`MissingAlgorithm` (i.e. genuine config errors on our side) map to `AlgorithmUnsupported`.

**Memory.** No allocations precede the error branches (all errors short-circuit before `jwt_str_to_malloc`), and on alloc-failure the Rust `String` drops normally. No leaks.

**Hew binding.** `jwt.hew` exposes the tagged enum via `try_encode`/`try_decode` returning `Result<String, JwtError>`, with a matching `JwtError` enum and `error_message()`. The legacy `encode`/`decode` now panic with the classified message instead of collapsing. P0 surface is visible to Hew callers.

**Tests (16).** Covers per-class tagging (algorithm-unsupported, invalid-key, token-malformed, signature-invalid, claims-expired, allocation-failure) on both encode and decode sides, plus success round-trip, `_hew` TLS tracking, `validate()` tri-state, insecure-decode, and `free(null)`. Each class asserts a *distinct* discriminator, not just `!= None`.

**Non-blocking:**
1. No test directly exercises `err_out = null` on `hew_jwt_encode`/`hew_jwt_decode`. The null-check in `write_err_out` is correct by inspection, but a one-liner test asserting no segfault when a caller opts out of the out-param would lock the invariant in. Worth adding in a follow-up.
2. The `_ => HewJwtError::TokenMalformed` catch-all in `classify_jwt_error` is reasonable today, but if `jsonwebtoken` is upgraded and HMAC verify failures ever route through `ErrorKind::Crypto(_)` (or a new variant) instead of `InvalidSignature`, they'd silently reclassify as "malformed". The `decode_reports_signature_invalid_for_wrong_secret` test pins current behavior — keep it, and consider making the fallback explicit for new variants on the next dep bump.

Approving. Enabling automerge.

PR #1438, opened 2026-04-21
fix(std/net/smtp): add LAST_SMTP_ERROR thread-local (closes #1437)

### LINKED ISSUE #1437: std/net/smtp: hew_smtp_send collapses six error classes into sentinel -1 with no retrievable detail
## Summary

`std/net/smtp` collapses six distinct error classes (address parse failure, `build_message` error, TLS setup failure, connection failure, MAIL FROM/RCPT rejection, transient vs. permanent SMTP status) into a sentinel `-1` return with no retrievable detail.

## Location

- `std/net/smtp/src/lib.rs:189-190` (address parse via `.parse().ok()?`)
- `std/net/smtp/src/lib.rs:231-240`, `:264-273` (`hew_smtp_send`, `hew_smtp_send_html` — return `-1` on any `Err`)
- `build_message` returns `Option<Message>` swallowing lettre errors

## Risk

Applications cannot retry transient failures (4xx) distinctly from permanent rejections (5xx), and operators have no telemetry for production SMTP debugging. Mirror of the #1414 pattern.

## Proposed fix

Add `LAST_SMTP_ERROR: RefCell<Option<String>>` thread-local mirroring `LAST_HTTP_ERROR` (`std/net/http/src/client.rs:23`). Populate on every `Err`. Expose `hew_smtp_last_error() -> *const c_char` export.

## Test approach

- Unit: feed unparseable `from` address → assert `-1` + `last_error` contains "parse".
- Unit: build with empty recipient list → distinct error string.
- Unit: clearing semantics — second successful call clears the error.

## Similar / related

- #1414 (TLS sentinel collapse)
- `std/net/http` already uses the `LAST_HTTP_ERROR` pattern

## Fail-closed / drop-safety

Null pointer returned from `hew_smtp_last_error` when no error recorded. Thread-local cleared on successful send.

slepp:
## Summary
- add a thread-local `LAST_SMTP_ERROR` and expose `hew_smtp_last_error()` for SMTP send failures
- replace sentinel-collapsing message construction/send paths with explicit error propagation in `hew_smtp_send` and `hew_smtp_send_html`
- add unit coverage for invalid from-address errors, empty payload errors, and clearing the last error after a successful send path

## Testing
- cargo fmt --all -- --check
- cargo test -p hew-std-net-smtp --quiet
- make ci-preflight

Closes #1437.

### DIFF
--- a/std/net/smtp/src/lib.rs
+++ b/std/net/smtp/src/lib.rs
@@ -1,7 +1,8 @@
 //! Hew runtime: SMTP email sending via lettre.
 //!
 //! Provides SMTP client functionality for compiled Hew programs.
-//! Opaque connection handles are freed via [`hew_smtp_close`].
+//! All returned strings and connection handles are allocated with `libc::malloc`
+//! / `Box` so callers can free them with the corresponding free function.
 
 // Force-link hew-runtime so the linker can resolve hew_vec_* symbols
 // referenced by hew-cabi's object code.
@@ -49,8 +50,8 @@ pub struct HewSmtpConn {
 ///
 /// Returns an empty string when no SMTP client error has been recorded.
 #[no_mangle]
-pub extern "C" fn hew_smtp_last_error() -> *const c_char {
-    str_to_malloc(&get_smtp_last_error()).cast_const()
+pub extern "C" fn hew_smtp_last_error() -> *mut c_char {
+    str_to_malloc(&get_smtp_last_error())
 }
 
 impl std::fmt::Debug for HewSmtpConn {
@@ -451,7 +452,7 @@ mod tests {
         // SAFETY: `err` was allocated by `hew_smtp_last_error`.
         let text = unsafe { CStr::from_ptr(err) }.to_str().unwrap().to_owned();
         // SAFETY: `err` was allocated via `malloc`.
-        unsafe { libc::free(err.cast_mut().cast()) };
+        unsafe { libc::free(err.cast()) };
         text
     }
 
--- a/std/net/smtp/src/lib.rs
+++ b/std/net/smtp/src/lib.rs
@@ -8,12 +8,34 @@
 #[cfg(test)]
 extern crate hew_runtime;
 
-use hew_cabi::cabi::cstr_to_str;
+use hew_cabi::cabi::{cstr_to_str, str_to_malloc};
+use std::cell::RefCell;
 use std::os::raw::c_char;
 
 use lettre::message::{header::ContentType, Mailbox};
 use lettre::transport::smtp::authentication::Credentials;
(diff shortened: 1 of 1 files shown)


slepp:
Review findings

One blocker on the public C ABI surface, otherwise the change looks good (fail-closed paths all populate `LAST_SMTP_ERROR`, success path clears it, 3 new tests cover negative/positive/clearing semantics and assert on the thread-local, not just the `-1` return).

**Convention / ownership mismatch on `hew_smtp_last_error`**

`std/net/smtp/src/lib.rs`:
```rust
pub extern "C" fn hew_smtp_last_error() -> *const c_char {
    str_to_malloc(&get_smtp_last_error()).cast_const()
}
```

The sibling in `std/net/http/src/client.rs:92` returns `*mut c_char`:
```rust
pub extern "C" fn hew_http_last_error() -> *mut c_char {
    str_to_malloc(&get_http_last_error())
}
```

Both allocate via `str_to_malloc` (libc::malloc under the hood), so the caller MUST `libc::free` the returned pointer or leak on every call. The tests in this PR already do exactly that (`libc::free(err.cast_mut().cast())`). But the public signature is `*const c_char` with no doc on ownership — that conventionally signals "borrowed, do not free" to C callers and any generated bindings, and the SMTP module header (unlike the HTTP one at `std/net/http/src/client.rs:4-5`) has no "all returned strings are malloc'd, free with the corresponding free function" disclaimer to fall back on.

Please:
1. Change the return type to `*mut c_char` to match `hew_http_last_error`.
2. Add a doc line stating the caller must free the returned pointer with `libc::free` (or extend the module-level doc comment to match http's).

Everything else checks out — diff is scoped to `std/net/smtp/src/lib.rs`, voice is clean, no surviving `.ok()?` / silent `None` in the error paths.

slepp:
Fixed. `hew_smtp_last_error` now returns `*mut c_char` like HTTP, the module docs now spell out caller-frees ownership, and the SMTP test helper frees the returned error string.

slepp:
Re-reviewed at fa790017. Ownership fix looks good:

- `hew_smtp_last_error` now returns `*mut c_char` via `str_to_malloc`, matching `hew_http_last_error`'s caller-frees contract.
- Module-level doc comment mirrors `std/net/http/src/client.rs:4-5` — callers know to `libc::free` the result.
- Tests use a `last_error_text()` helper that `libc::free`s the pointer after reading, so Miri/ASan won't complain.
- `LAST_SMTP_ERROR` is still populated on every send error path (null conn, build_message, transport send) via the new `smtp_send_impl` helper, and cleared on success.

Prior finding resolved. Enabling squash auto-merge; CI is green where it has reported (3/10) with the remainder still pending, no failures.

PR #1439, opened 2026-04-21
fix(std/encoding/compress): add decompression-bomb size cap (closes #1411)

### LINKED ISSUE #1411: std/encoding/compress: no decompression-bomb limit
`std/encoding/compress/src/lib.rs:27–49` (`read_to_malloc`) calls `reader.read_to_end(&mut buf)` with no limit. A ~1 KB gzip bomb expanding to GBs decompresses faithfully and then `malloc`s; the process OOMs instead of returning null. `hew_gzip_decompress`, `hew_deflate_decompress`, `hew_zlib_decompress` (approx lines 109, 165, 221) all route through this helper. `std/net/http` server bodies can flow through via middleware — one request can take down a node.

## Check-list
- [ ] Accept an explicit `max_output_len: usize` at the FFI boundary
- [ ] Wrap the reader with `.take(max_output_len as u64 + 1)`; if the buffer exceeds the cap, free the partial allocation and return null
- [ ] Document a conservative default cap in the module header
- [ ] Unit test: 1 KB→10 MB bomb with cap=1 MB fails cleanly, no OOM

slepp:
## Summary
- add a shared output-size cap for gzip/deflate/zlib encode and decode paths
- fail closed with a structured last-error message when the cap or override is invalid
- cover roundtrips, bomb rejection, and exact-cap boundary behavior in unit tests

## Testing
- cargo fmt --all -- --check
- cargo test -p hew-std-encoding-compress --quiet
- make ci-preflight

### DIFF
--- a/std/encoding/compress/src/lib.rs
+++ b/std/encoding/compress/src/lib.rs
@@ -1,34 +1,123 @@
 //! Hew runtime: gzip, deflate, and zlib compression/decompression.
 //!
 //! Provides compression and decompression using gzip, raw deflate, and zlib
-//! formats for compiled Hew programs. All returned buffers are allocated with
-//! `libc::malloc` so callers can free them with [`hew_compress_free`].
+//! formats for compiled Hew programs. Outputs are capped at
+//! [`DEFAULT_MAX_OUTPUT_LEN`] bytes by default to fail closed on compression and
+//! decompression bombs; set [`MAX_OUTPUT_LEN_ENV`] to override the cap for a
+//! process. All returned buffers are allocated with `libc::malloc` so callers
+//! can free them with [`hew_compress_free`].
 
 // Force-link hew-runtime so the linker can resolve hew_vec_* symbols
 // referenced by hew-cabi's object code.
 #[cfg(test)]
 extern crate hew_runtime;
 
-use std::io::Read;
+use std::ffi::c_char;
+use std::io::{self, Read};
 
 use flate2::read::{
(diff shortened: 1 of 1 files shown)


slepp:
Reviewed against the bomb-cap brief.

Scope is limited to std/encoding/compress/src/lib.rs. The cap is a named, documented const (DEFAULT_MAX_OUTPUT_LEN = 256 MiB) with a documented env override (HEW_COMPRESS_MAX_OUTPUT_LEN). The override parse is strict — usize::from_str plus an explicit NotUnicode branch, both surfaced as InvalidLimitEnv through last-error; no unwrap_or_default or silent .ok()? anywhere in the changed region.

The cap is enforced pre-allocation via Read::take(limit + 1) and a post-read buf.len() > max_output_len check, which is the correct bomb pattern — the +1 is what makes cap-exact succeed and cap+1 fail cleanly instead of silently truncating. OutputLimitExceeded is a distinct CompressError variant with a 'output exceeded N byte limit' message, distinguishable from Read and AllocationFailed.

Symmetry is applied through a single read_to_vec_with_limit helper reused by all six compress/decompress entry points; configured_max_output_len() is the one source of truth for the runtime cap.

Boundary coverage is both-sided: test_gzip_decompress_boundary_at_limit decodes exactly LIMIT bytes successfully and rejects LIMIT+1, asserting out_len is zeroed and the error message contains the limit. test_gzip_compress_boundary_at_limit mirrors that on the compress side against a baseline size.

CI: 6 green, 0 failing, remainder pending. Enabling auto-merge on squash.

PR #1440, opened 2026-04-21
fix(std/encoding/yaml): return null on base64 decode failure instead of silent empty vec (closes #1418)

### LINKED ISSUE #1418: std/encoding/yaml: hew_yaml_get_bytes silently returns empty on base64 decode failure
`std/encoding/yaml/src/lib.rs:298–307` — `base64::…::STANDARD.decode(s).unwrap_or_default()` (line 302). A non-base64 YAML string silently produces `Vec::new()` → zero-length `HewVec`. Downstream "bytes are empty" checks misfire; error invisible. No `hew_yaml_last_error` for byte-decode failures (existing one only covers parse errors).

## Check-list
- [ ] Return `std::ptr::null_mut()` on base64 decode failure (consistent with the rest of the module)
- [ ] Record the error in the existing `LAST_PARSE_ERROR` slot (or the new structured result after the datetime sentinel issue)
- [ ] Test: malformed base64 returns null + populated last-error

slepp:
## Summary
- return null from `hew_yaml_get_bytes` when base64 decoding fails and record the decode error in `LAST_PARSE_ERROR`
- clear the last error on successful byte decoding so valid empty base64 still returns an empty `HewVec`
- add regression tests for invalid, empty, and valid base64 byte extraction paths

## Testing
- cargo fmt --all -- --check
- cargo test -p hew-std-encoding-yaml --quiet
- make ci-preflight

### DIFF
--- a/std/encoding/yaml/src/lib.rs
+++ b/std/encoding/yaml/src/lib.rs
@@ -282,8 +282,8 @@ pub unsafe extern "C" fn hew_yaml_get_string(val: *const HewYamlValue) -> *mut c
 /// Get a base64-decoded bytes value from a [`HewYamlValue`].
 ///
 /// Returns a newly allocated [`HewVec`] for valid string inputs. Non-string
-/// values return null. Invalid base64 inputs decode to an empty bytes vector to
-/// match `std::encoding::base64::decode`.
+/// values return null. Invalid base64 inputs return null and populate
+/// [`hew_yaml_last_error`].
 ///
 /// # Safety
 ///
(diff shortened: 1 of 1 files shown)


slepp:
Scope is a single file (`std/encoding/yaml/src/lib.rs`, +85/-5), and the fix lands exactly where #1418 pointed. The `unwrap_or_default` is gone; the decode error path now calls `set_parse_last_error` with a message that carries the base64 error, then returns null. Success path clears the last error before allocating, so a valid empty string still yields an empty `HewVec` without a stale error sticking around. That's the fail-closed shape CLAUDE.md §2 asks for.

The three new tests cover the three cases cleanly: invalid → null with `base64`/`decode` substring in last-error, empty string → empty vec, valid `aGV3` → `b"hew"` round-trip. Ownership matches the rest of the file — `hew_vec_free` on the returned `*mut HewVec`, `hew_yaml_free` on the field and root. No other yaml functions touched.

CI: clippy + fmt green, platform builds pending. Enabling automerge.

PR #1444, opened 2026-04-21
fix(std/net/websocket): apply explicit WebSocketConfig caps on accept/connect (closes #1443)

### LINKED ISSUE #1443: std/net/websocket: no max_frame/max_message cap — DoS-by-default
## Summary

`std/net/websocket` accepts tungstenite's default frame/message sizes (64 MiB each) — one misbehaving peer can exhaust memory across many attached actors.

## Location

`std/net/websocket/src/lib.rs` (every `tungstenite::accept(stream)` / `tungstenite::connect(url)` call site). No `WebSocketConfig` is currently built.

## Risk

Remote DoS: a peer sending a single 64 MiB frame per connection can make the runtime swap or OOM on even modestly-fanout deployments.

## Proposed fix

Construct `tungstenite::protocol::WebSocketConfig` with conservative caps (e.g. `max_message_size: Some(8 MiB)`, `max_frame_size: Some(1 MiB)`) and use `accept_with_config` / `client_with_config` at every site. Expose the caps via env-var overrides or new attach args.

## Test approach

- Unit: server configured with a small cap (e.g. 64 KiB); client sends a 1 MiB frame; assert server closes cleanly with "message too large" rather than OOMing.
- Unit: a frame at exactly the cap succeeds; cap + 1 fails.

slepp:
## Summary
- apply a shared `WebSocketConfig` helper with explicit 8 MiB message and 1 MiB frame caps to every websocket accept/connect attach site
- switch the client handshake paths to tungstenite's configured connect API so the runtime never falls back to the 64 MiB / 16 MiB defaults
- add regression tests for oversized frames being rejected and exactly-at-cap frames succeeding

## Testing
- cargo fmt --all -- --check
- cargo test -p hew-std-net-websocket --quiet
- cargo test -p hew-std-net-websocket --quiet
- cargo test -p hew-std-net-websocket --quiet
- make ci-preflight
- git fetch origin && git rebase origin/main

### DIFF
--- a/std/net/websocket/src/lib.rs
+++ b/std/net/websocket/src/lib.rs
@@ -17,6 +17,8 @@ use std::sync::{Arc, Mutex, MutexGuard, PoisonError};
 use std::thread::JoinHandle;
 use std::time::{Duration, Instant};
 
+use tungstenite::client::connect_with_config;
+use tungstenite::protocol::WebSocketConfig;
 use tungstenite::stream::MaybeTlsStream;
 use tungstenite::{Message, WebSocket};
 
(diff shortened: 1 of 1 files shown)


slepp:
Reviewed. Caps plumbed through every accept/connect attach site; bare `tungstenite::accept`/`connect` are gone from the file (grepped PR HEAD — only `accept_with_config`/`connect_with_config` remain, including in tests). Named consts WEBSOCKET_MAX_MESSAGE_SIZE_BYTES=8 MiB and WEBSOCKET_MAX_FRAME_SIZE_BYTES=1 MiB are documented; `websocket_config` carries a debug_assert that frame ≤ message. No env-var override introduced, so no parse path to audit; fail-closed posture is preserved (no new `unwrap_or_default` or `.ok()?` in the config builder).

Test coverage lands the two ends of the bomb check: the oversized-frame test configures a 64 KiB cap server-side, sends 1 MiB from the client, and asserts the server gets `CapacityError::MessageTooLong { size: 1 MiB, max_size: 64 KiB }` before the connection unwinds; the at-cap test sends exactly 64 KiB and reads it back cleanly. Not a strict cap+1 boundary on the failing side — 16× over — but the cap-fires assertion is direct (pattern-matches size + max_size on the tungstenite error), so the signal is real.

CI still pending on the long lanes; approving with automerge.

PR #1445, opened 2026-04-21
fix(std/encoding/json): return null on base64 decode failure in hew_json_get_bytes (closes #1441)

### LINKED ISSUE #1441: std/encoding/json: hew_json_get_bytes silently returns empty Vec on malformed base64
## Summary

`std/encoding/json::hew_json_get_bytes` silently returns an empty `Vec` on malformed base64 (`base64::decode(s).unwrap_or_default()`). Downstream code cannot distinguish "empty string was encoded" from "base64 decode failed".

## Location

`std/encoding/json/src/lib.rs:290-292`

## Risk

Adversarial JSON input (non-base64 in a `bytes` field) produces a silently empty buffer; receiver treats it as legitimate and (e.g.) records an empty credential, empty signature, empty payload. Parallel of #1418 (YAML) for JSON.

## Proposed fix

Return `std::ptr::null_mut()` on decode failure (same policy already used on line 296 for the non-string path), and populate `LAST_PARSE_ERROR` with the decoder message. Mirror the fix landed for #1418 in `std/encoding/yaml`.

## Test approach

- Unit: `hew_json_get_bytes` on `{"b":"not***base64***"}` → null + `LAST_PARSE_ERROR` populated
- Unit: `{"b":""}` → empty Vec (legitimate empty)
- Unit: valid base64 round-trip

slepp:
## Summary
- return null from `hew_json_get_bytes` when base64 decoding fails and record the decode error in `LAST_PARSE_ERROR`
- clear the last error after successful byte decoding so empty base64 remains distinguishable from errors
- add regression tests for invalid, empty, and valid base64 byte extraction paths

## Testing
- cargo fmt --all -- --check
- cargo test -p hew-std-encoding-json --quiet (3/3)
- make ci-preflight

### DIFF
--- a/std/encoding/json/src/lib.rs
+++ b/std/encoding/json/src/lib.rs
@@ -272,8 +272,8 @@ pub unsafe extern "C" fn hew_json_get_string(val: *const HewJsonValue) -> *mut c
 /// Get a base64-decoded bytes value from a [`HewJsonValue`].
 ///
 /// Returns a newly allocated [`HewVec`] for valid string inputs. Non-string
-/// values return null. Invalid base64 inputs decode to an empty bytes vector to
-/// match `std::encoding::base64::decode`.
+/// values return null. Invalid base64 inputs return null and populate
+/// [`hew_json_last_error`].
 ///
 /// # Safety
 ///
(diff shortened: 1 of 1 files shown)


slepp:
Clean mirror of the YAML fix in #1440 — same `match`/`set_parse_last_error`/`clear_parse_last_error` pattern, same three-test shape (invalid/empty/round-trip), scope limited to `hew_json_get_bytes` and its tests. Empty-string input still returns a legitimate empty `Vec` distinguishable from the null error path. Clippy and format are green; remaining checks are still running, so queuing auto-merge on squash.

PR #1446, opened 2026-04-21
fix(std/net/quic): propagate resolver errors through LAST_CONSTRUCTOR_ERROR (closes #1442)

### LINKED ISSUE #1442: std/net/quic: resolve_bind_addr / resolve_connect_addr swallow DNS errors with .ok()?
## Summary

`std/net/quic::resolve_bind_addr` and `resolve_connect_addr` swallow DNS resolver errors with `.ok()?` — an unparseable string, a DNS failure (NXDOMAIN / permission denied / nameserver down), and an empty result set all collapse into a bare `None`.

## Location

`std/net/quic/src/lib.rs:232`, `std/net/quic/src/lib.rs:243`

## Risk

Operator sees a generic "QUIC bind failed" or "connect failed" with no cause. Diagnostic signal is destroyed at the resolver boundary.

## Proposed fix

Change return type to `Result<SocketAddr, String>`. Propagate the `io::Error` message into the existing `LAST_CONSTRUCTOR_ERROR` thread-local (`std/net/quic/src/lib.rs:43`). No `.ok()?` anywhere in the function.

## Test approach

- Unit: `resolve_bind_addr("bad:host:here")` → distinct error mentioning parse
- Unit: `resolve_connect_addr("unresolvable.invalid.:443")` → distinct error mentioning resolution
- Unit: valid `127.0.0.1:4433` still returns Ok

slepp:
## Summary
- make the QUIC bind/connect resolvers fail closed with `Result<SocketAddr, String>` and preserve distinct parse, resolution, and empty-result errors
- record resolver failures in `LAST_CONSTRUCTOR_ERROR` and propagate them through the server constructors and endpoint connect path
- add regression tests for parse failures, reserved `.invalid.` resolution failures, literal socket addresses, and constructor/connect error propagation

## Testing
- cargo fmt --all -- --check
- cargo test -p hew-std-net-quic --quiet (3/3)
- make ci-preflight

### DIFF
--- a/std/net/quic/src/lib.rs
+++ b/std/net/quic/src/lib.rs
@@ -221,26 +221,72 @@ fn self_signed_server_config() -> Result<ServerConfig, BoxError> {
     )))
 }
 
-fn resolve_bind_addr(addr: &str) -> Option<SocketAddr> {
-    let owned;
-    let bind_addr = if addr.starts_with(':') {
-        owned = format!("0.0.0.0{addr}");
-        owned.as_str()
-    } else {
-        addr
+fn split_host_port(addr: &str) -> Result<(&str, &str), &'static str> {
+    if let Some(rest) = addr.strip_prefix('[') {
+        let Some((host, port)) = rest.split_once("]:") else {
+            return Err("missing closing `]` or port separator");
+        };
+        if host.is_empty() {
+            return Err("missing host");
+        }
+        if port.is_empty() {
+            return Err("missing port");
+        }
+        return Ok((host, port));
+    }
+
+    let Some((host, port)) = addr.rsplit_once(':') else {
+        return Err("missing port separator");
     };
-    bind_addr.to_socket_addrs().ok()?.next()
+    if host.is_empty() {
+        return Err("missing host");
+    }
+    if host.contains(':') {
+        return Err("unexpected `:` in host; IPv6 addresses must use `[addr]:port`");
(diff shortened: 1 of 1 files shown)


slepp:
Reviewed. Scope is clean (single file). Resolver converted to `Result<SocketAddr, String>` with all three FFI call sites updated; `LAST_CONSTRUCTOR_ERROR` populated on every Err path via `set_constructor_error_and_return`. 

The error-text adjustment is principled, not a dilution: the new message `could not resolve {bind|connect} address \`{addr}\`: {parse failure|resolution failure|resolution returned no results}: ...` is a superset of the old wording. The e2e fixture's `.contains("could not resolve bind address")` still matches, and the three failure classes remain distinguishable by the tokens `parse failure` / `resolution failure` / `resolution returned no results`. The FFI-level test in `new_server_sets_last_error_on_bad_address` now also asserts `parse failure`, which locks in the class distinction.

F9 scope respected: `BlockingRecvQueue::pop_blocking`'s `.ok()?` on the condvar wait (line 83) is untouched — that's a separate concern. No other surviving `.ok()?` in the resolver path.

Six new tests cover parse/resolution/literal for both bind and connect. The `resolution returned no results` branch isn't unit-tested in isolation, but `to_socket_addrs()` effectively never returns an empty iterator on Ok, so the omission is fine.

Enabling automerge.

PR #1452, opened 2026-04-21
ci: add voice-guard lint for PR titles/bodies/commit messages

### LINKED ISSUE #1449: CI: pre-merge voice-guard linter (model names + orchestration vocabulary)
## Summary

A retrospective audit of the last 200 merged PRs found 28 PRs (14%) containing voice-guard violations in their titles/bodies, which get inherited into squash-commit messages and are durable in `git log`:
- 3 PRs with explicit model names (`gpt-5.4`, `claude`) — P0
- 25 PRs with orchestration vocabulary (primarily `lane` / `fallback lane` / `this lane does X`) — P1

The existing PR template has a voice-guard checkbox but it is not enforced.

## Proposed fix

Add a pre-merge CI lint (a new job in `.github/workflows/ci.yml`, or extend an existing lint job) that:

1. On every PR, fetches the PR title + body.
2. Scans for forbidden tokens: `\bgpt-\d`, `\bclaude-?(opus|sonnet|haiku)`, `\bsonnet\b`, `\bhaiku\b`, `\bopus\b` (model names); `\blane\b`, `\bwave\s*\d`, `\borchestrat`, `\bAPPROVE(D)?\b`, `\bPASS:\b`, `\bR\d+\s+F\d+` (orchestration vocab).
3. Fails with a clear diagnostic pointing at the matched phrase.
4. Also checks the commit message(s) of the PR head for the same tokens (catches ephemeral drafts still visible in file history if a merge is a true merge commit).

Also refresh `.github/pull_request_template.md` to remove the "fallback lane" / "this lane does X" phrasings if present in the scaffolding.

## Risk

Without this gate, orchestration internals continue to leak into permanent git history, which (a) exposes internal workflow to external contributors and (b) reduces PR readability for a reviewer whose context is just the code.

## Test approach

- Add a unit test (rust or python, whichever fits the lint's language) that feeds the linter the known-bad strings from PRs #1294, #1291, #1223, #1408, #1296 and asserts it flags them.
- Add happy-path strings that must NOT flag (e.g. `CLAUDE.md` as a filename reference, `lane change` in a networking context).

## Evidence

See `.tmp/orchestration/pr-audit-last-200.md` Section 1 for the full PR list.

slepp:
Fixes #1449. Adds a CI check that blocks PRs whose title, body, or commit messages contain forbidden tokens, preventing these from leaking into permanent squash-commit history. Also refreshes the PR template checklist and the ci-preflight dispatcher wording so user-visible guidance stays neutral. Tests cover known-bad strings sourced from a retrospective audit and known-good edge cases (filename references, substring guards).

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -18,6 +18,7 @@
      For changes with no new tests, explain why (e.g., "covered by existing E2E"). -->
 
 ## Quality Checklist
+- [ ] PR title, body, and commit messages avoid internal-only orchestration/model vocabulary
 - [ ] No new `.ok()?` or `unwrap_or_default()` in codegen without `// JUSTIFIED` comment
 - [ ] New allocations have cleanup paths for sync, async, and actor shutdown contexts
 - [ ] Serialization changes include round-trip encode/decode tests
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,6 +10,7 @@ on:
 permissions:
   contents: read
   checks: write
+  pull-requests: read
   security-events: write
 
 env:
@@ -17,6 +18,26 @@ env:
   CARGO_TERM_COLOR: always
 
 jobs:
+  voice-guard:
+    name: Voice guard
+    if: github.event_name == 'pull_request'
+    runs-on: ubuntu-24.04
+    timeout-minutes: 5
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: actions/setup-python@v5
+        with:
+          python-version: '3.x'
+
+      - name: Run voice-guard unit tests
+        run: python3 -m unittest scripts/test_voice_guard.py
+
+      - name: Scan PR metadata and commit messages
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: python3 scripts/voice_guard.py --pr-number "${{ github.event.pull_request.number }}"
+
   # ─────────────────────────────────────────────────────────────────────────
   # Path change detection — skip heavy jobs for doc-only changes
   # ─────────────────────────────────────────────────────────────────────────
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -19,6 +19,10 @@ See the [Building from Source](README.md#building-from-source) section of the RE
 4. Run `make lint` to check for warnings
 5. Submit a pull request
 
+PR titles, PR bodies, and commit messages are part of the permanent project history after squash merge.
+Keep them free of model names, orchestration jargon, and internal-only path references such as `.claude/`.
+CI rejects these leaks with the voice-guard check before merge.
+
 ### Using LESSONS.md
 
 [`LESSONS.md`](LESSONS.md) is a structured decision aid for contributors. Before merging a change, match it against the **trigger** column in LESSONS.md and apply every matching row's **apply** checklist. Start with **P0** rows (correctness and boundary safety), then **P1** (parity, tests, diagnostics), then **P2** (architecture and cleanup). When two rules conflict, keep the stricter fail-closed, ownership-preserving, or parity-preserving rule.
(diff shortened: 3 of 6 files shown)


slepp:
Reviewed. The gate does not yet pass on itself — the voice-guard CI job on this PR is red. Requesting a revision before merge.

**Blocker — fails its own check (CI red)**
- PR body contains `orchestration-vocab list`, which matches the `\borchestrat\w*\b` pattern in `scripts/voice_guard.py`. Either rephrase the body (e.g. "a small list of internal-only vocabulary") or scope that pattern tighter. Right now the linter rejects its own introduction.

**Regex surface — some patterns are too broad**
- `scripts/voice_guard.py:36` `\bAPPROVE(?:D)?\b` is compiled case-insensitive, so ordinary engineering prose ("the change was approved by the owner") will also fail. The audit leak was the literal `APPROVE` / `APPROVED` verdict token; tightening to case-sensitive (matching the `PASS:` pattern already on line 37) would keep the signal without the collateral.
- `scripts/voice_guard.py:41` `\blane\b` plus the single allowlist `lane change in routing` is fragile. Any future PR body mentioning a swim lane, fast lane, or similar will trip the gate with no escape hatch. Consider requiring a neighbouring qualifier (`(?:fallback|bounded|preflight|this) lane`, already covered by other rules) and dropping the standalone `\blane\b` entry, or expanding allowlists.
- `scripts/voice_guard.py:27` `\borchestrat\w*\b` — see blocker above. Matches `orchestrator`, which is the name of a committed project skill (`hew-orchestrator`), so future PRs referencing that skill by name will also be blocked.

**False-positive controls — inconsistent fence handling**
- `scripts/voice_guard.py:76-95` only strips fenced code blocks before applying `PLAINTEXT_ONLY_PATTERNS` (just `.claude/`). Every other pattern (models, `Wave N`, `fallback lane`, `APPROVE`, etc.) is scanned against the raw text, fences and inline backticks included. That contradicts the intent documented in the test `test_claude_path_allowed_inside_fenced_code_block_only`: a PR body that quotes, say, `\`Wave 3\`` as a policy example will still fail. Either strip fences for all patterns, or document the asymmetry in `CONTRIBUTING.md`.
- No handling for inline backticks at all (`\`...\``), only triple-fence blocks.

**Test coverage gaps**
- `scripts/test_voice_guard.py` only exercises `scan()`. There is no test for `_load_pr_entries` / `_load_commits_file` (the commit-message traversal path is the whole reason this check exists — squash commits bake headline + body into history). A small fixture exercising `_load_commits_file` with a JSON list would close that gap.
- No negative test for the fence asymmetry above — if that behaviour is intentional, a test should pin it; if not, a failing test would surface it.

**Scope clarification requested**
- The hand-off note mentioned PR#1436 still flags due to legacy `fallback lane` text. Confirming: the scan target is the live PR metadata via `gh pr view`, so re-running on an already-merged PR would re-flag its preserved body, but this check only runs on the `pull_request` event for open PRs — it will not retroactively block merged history. Worth a one-line note in `CONTRIBUTING.md` so future contributors do not worry about merged-PR rescans.

**Things that look good**
- `subprocess.run(..., check=True)` on the `gh pr view` call — no silent swallow on missing token.
- Exit code 1 on findings, `pull-requests: read` permission added, `GH_TOKEN` wired through the workflow step.
- `scripts/ci-preflight-dispatcher.sh` change is output-only; the lane→profile mapping preserves routing behaviour and just relabels the printed line.
- `CONTRIBUTING.md` documents the new rule and names `.claude/` explicitly.
- Template checklist addition is neutral and actionable.
- Positive test cases pull directly from the retrospective audit, and the negative cases (`landed`, `CLAUDE.md`, `lane change`) cover the word-boundary edges the review asked about.

Suggested path to green: (a) rephrase the PR body to drop `orchestration-vocab`, (b) either tighten `\borchestrat\w*\b` and `\bAPPROVE(?:D)?\b` or add a short allowlist, (c) apply fence stripping to the full pattern set (or scope the fence exemption explicitly), (d) add one commit-message-loader test.

slepp:
Re-review clean. The six earlier items land correctly:
- orchestration stem removed from patterns
- caps-only literal matched, prose approvals left alone
- narrowed forms catch the leak shapes without tripping routing prose
- fence and inline-backtick stripping applied uniformly before every pattern
- new commit-file test covers both a flagged and a neutral commit

Dogfooded against this PR, #1294 (expected flag), #1447, and #1453 — results match the submitter's report. CI voice-guard job is registered and green. Auto-squash queued.

PR #1453, opened 2026-04-21
fix(std/net/tls): distinguish EOF/timeout/error on hew_tls_read via LAST_TLS_ERROR (closes #1413)

### LINKED ISSUE #1413: std/net/tls: hew_tls_read collapses EOF, timeout, error into empty
`std/net/tls/src/lib.rs:135–170` — `match s.stream.read(&mut buf) { Ok(0) | Err(_) => empty, Ok(n) => … }` (lines 152–153): a TLS alert, a reset connection, `WouldBlock`, and a clean half-close all produce an empty `HewVec`. Higher-level Hew code sees "stream returned 0 bytes" and treats it as orderly EOF, committing partial state. No `hew_tls_last_error` or out-status channel.

`hew_tls_write` at 97–124 has the same shape — returns `-1` for every error class.

## Check-list
- [ ] Add `*out_status: *mut c_int` parameter (0 = EOF, negative = error class) or `hew_tls_last_error` thread-local analogous to `hew_datetime_last_error`
- [ ] Record the underlying `io::ErrorKind` / rustls alert before returning empty
- [ ] Matching treatment for `hew_tls_write`
- [ ] Test: force-close mid-stream; caller observes distinguishable error vs EOF

slepp:
## Summary
- add a thread-local LAST_TLS_ERROR and export hew_tls_last_error with malloc-backed ownership
- classify empty hew_tls_read returns as eof, timeout, or io transport errors while keeping size-0 reads distinct
- add unit coverage for eof, timeout, non-EOF io errors, empty-but-open reads, and successful-read error clearing

## Validation
- cargo fmt --all -- --check
- cargo test -p hew-std-net-tls --quiet
- cargo test -p hew-std-net-tls --quiet (flake 3/3)
- make ci-preflight

### DIFF
--- a/std/net/tls/src/lib.rs
+++ b/std/net/tls/src/lib.rs
@@ -2,19 +2,21 @@
 //!
 //! Exposes a blocking C ABI over rustls for TLS client connections.
 //! All `extern "C"` functions are designed to be called from compiled Hew
-//! programs via FFI.
+//! programs via FFI. Any string returned by [`hew_tls_last_error`] is allocated
+//! with `libc::malloc`; callers must free it with `libc::free`.
 
 // Force-link hew-runtime so the linker can resolve hew_vec_* symbols
 // referenced by hew-cabi's object code.
 #[cfg(test)]
 extern crate hew_runtime;
 
-use std::io::{Read, Write};
+use std::cell::RefCell;
+use std::io::{self, Read, Write};
 use std::net::TcpStream;
 use std::os::raw::{c_char, c_int};
 use std::sync::Arc;
 
-use hew_cabi::cabi::cstr_to_str;
+use hew_cabi::cabi::{cstr_to_str, str_to_malloc};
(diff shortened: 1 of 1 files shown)


slepp:
Reviewed. Thread-local `LAST_TLS_ERROR` with `RefCell<Option<String>>` and `hew_tls_last_error` malloc export matches the http/cron/datetime pattern (str_to_malloc of the current message, empty string when unset). Read path distinguishes eof / timeout (TimedOut|WouldBlock) / `io: …` / size==0 no-op / success-clears, and the write path now classifies flush and write_all errors instead of swallowing them. Prior String is dropped via RefCell assignment; no new unfreed allocations. Tests cover all five branches with a deterministic MockReader (including a panic-on-call guard for size==0), no threads or timers, so not flake-prone. Scope limited to std/net/tls/src/lib.rs. CI green. Auto-merge queued.

PR #1454, opened 2026-04-21
fix(astgen): propagate errors from hew_emit.rs unwrap_or_default sites

### LINKED ISSUE #1450: hew-astgen: two .unwrap_or_default() sites in hew_emit.rs violate fail-closed (regression from #1308)
## Summary

PR #1308 (`feat(astgen): --mode hew generator calibration against std/misc/uuid`) introduced two `.unwrap_or_default()` sites in `hew-astgen/src/hew_emit.rs`:
- Line 201
- Line 215

Both are in codegen-adjacent emit paths and violate CLAUDE.md §2 (Fail-Closed Codegen: "Never use `.ok()?`, `.unwrap_or_default()`, or silent `None` returns in codegen paths").

## Location

- `hew-astgen/src/hew_emit.rs:201`
- `hew-astgen/src/hew_emit.rs:215`

## Proposed fix

For each site, either:
(a) Propagate the error explicitly via `?` or a structured diagnostic, OR
(b) If truly infallible in this context, annotate with `// JUSTIFIED: <reason>` per the CLAUDE.md §2 exception.

Option (a) is preferred where the operation can legitimately fail.

## Test approach

For each site, add a test that exercises the "error" side of the path being unwrapped. If the path is genuinely infallible, document why in the `// JUSTIFIED:` annotation and the test is not required.

## Evidence

Retrospective audit of last 200 merged PRs: `.tmp/orchestration/pr-audit-last-200.md` Section 2.

slepp:
Fixes #1450. This updates the hew_emit fail-closed coverage so the empty-path error is exercised for the return-type branch as well as the parameter branch, matching the existing explicit error propagation in the emit helpers. The emit path already returns structured errors instead of defaulting, and this change keeps both targeted sites guarded against regressions.\n\nThis is safe because it only extends the existing hew-astgen unit coverage and does not change emitted output for valid inputs. Validation included cargo fmt --all -- --check, cargo clippy --workspace --tests -- -D warnings, cargo test -p hew-astgen --quiet, a 3x flake run of the new test, and make ci-preflight.

### DIFF
--- a/hew-astgen/src/hew_emit.rs
+++ b/hew-astgen/src/hew_emit.rs
@@ -421,6 +421,40 @@ mod tests {
         );
     }
 
+    #[test]
+    fn errors_when_a_return_type_path_has_no_segments() {
+        let mut file = syn::parse_file(
+            r#"
+                #[no_mangle]
+                pub unsafe extern "C" fn hew_uuid_parse(s: *const c_char) -> i32 {
+                    0
+                }
+            "#,
+        )
+        .unwrap();
+        let syn::Item::Fn(func) = &mut file.items[0] else {
+            panic!("expected function item");
+        };
+        func.sig.output = syn::ReturnType::Type(
+            syn::token::RArrow::default(),
+            Box::new(syn::Type::Path(syn::TypePath {
+                qself: None,
+                path: syn::Path {
+                    leading_colon: None,
+                    segments: syn::punctuated::Punctuated::new(),
+                },
+            })),
+        );
+
+        let err = extract_extern_fns_from_file(&file).unwrap_err();
+        assert_eq!(
+            err,
+            HewEmitError::MissingTypePathSegment {
+                ty: "<empty path>".to_string()
+            }
+        );
+    }
+
     // ── generate_hew_module ────────────────────────────────────────────────
 
     #[test]


slepp:
Reviewed. The audit flagging `unwrap_or_default` at hew_emit.rs:201,215 was stale — `git show origin/main:hew-astgen/src/hew_emit.rs | grep -n 'unwrap_or_default\|\.ok()?'` returns zero hits, so those sites were already replaced with structured error propagation (`last_path_ident` + `map_type` now return `HewEmitError::MissingTypePathSegment`) in a prior PR. Residual value here is the regression test for the return-type branch of `extract_return_ty`, which was previously only covered on the parameter branch.

Test quality: constructs a `syn::ReturnType::Type` with an empty-segments `TypePath` and asserts the exact `MissingTypePathSegment { ty: "<empty path>" }` variant. Deterministic (pure syn AST construction, no I/O), and a future reintroduction of `unwrap_or_default` / `.ok()?` on that branch would fail the assertion. Scope is one file, 34 additions, tests only.

One nit: the PR title says "propagate errors from ... unwrap_or_default sites" but nothing is being propagated in this diff — the propagation already exists on main. The title would more accurately read "add regression test for empty return-type path". Not blocking.

Merging on auto.

PR #1457, opened 2026-04-21
fix(std/net/quic): recover from poisoned BlockingRecvQueue mutex instead of silent drop

### LINKED ISSUE #1456: std/net/quic: BlockingRecvQueue .ok()? on mutex lock silently drops poisoned state
## Summary

`std/net/quic/src/lib.rs:83` uses `.ok()?` on a `.lock()` result against a shared mutex in `BlockingRecvQueue`. If the mutex is poisoned (a holder thread panicked), the code silently returns `None`, which the caller may treat as a legitimate "no message" signal.

This violates CLAUDE.md §9 (Concurrency Safety) and the project's established poison-safe convention.

## Location

`std/net/quic/src/lib.rs:83`

## Proposed fix

Replace `.ok()?` with `.unwrap_or_else(std::sync::PoisonError::into_inner)`, matching the poison-safe pattern established in the runtime poison-safe wrapper series (#1225, #1303, #1305, #1306).

## Test approach

Unit test: poison the queue's mutex deliberately (via a panicking thread holding the lock), then call the public recv path; assert the queue recovers and returns the expected result (or a defined error) rather than silently dropping.

## Evidence

R6 workspace audit F9. Similar to the poison-safe wrapper series.

slepp:
Fixes #1456. The recv path used \.ok()? on the mutex lock, which silently returned None when a holder thread panicked, and callers read that as 'no message available'. Now recovers the inner state via PoisonError::into_inner, matching the poison-safe convention established for runtime state. Adds a regression test that deliberately poisons the mutex and asserts continued correctness.

### DIFF
--- a/std/net/quic/src/lib.rs
+++ b/std/net/quic/src/lib.rs
@@ -64,23 +64,28 @@ struct EventQueue {
 
 impl EventQueue {
     fn push(&self, kind: i32) -> bool {
-        let Ok(mut queue) = self.queue.lock() else {
-            return false;
-        };
+        let mut queue = self
+            .queue
+            .lock()
+            .unwrap_or_else(std::sync::PoisonError::into_inner);
         queue.push_back(kind);
         self.ready.notify_one();
         true
     }
 
     fn pop_blocking(&self) -> Option<i32> {
-        let Ok(mut queue) = self.queue.lock() else {
-            return None;
-        };
+        let mut queue = self
+            .queue
+            .lock()
+            .unwrap_or_else(std::sync::PoisonError::into_inner);
         loop {
             if let Some(kind) = queue.pop_front() {
                 return Some(kind);
             }
-            queue = self.ready.wait(queue).ok()?;
+            queue = self
+                .ready
+                .wait(queue)
+                .unwrap_or_else(std::sync::PoisonError::into_inner);
         }
     }
 }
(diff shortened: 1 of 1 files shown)


slepp:
Thanks for the targeted fix — the `wait()` site and the regression test both look good, and the recovery pattern matches the runtime convention.

One gap worth addressing before this lands: the same `EventQueue` still silently drops on poison at the initial lock acquisition in both methods:

- `std/net/quic/src/lib.rs:67` — `push` returns `false` on poison (event dropped).
- `std/net/quic/src/lib.rs:76` — `pop_blocking` returns `None` on poison (caller reads as 'no message available', which is the exact symptom the PR body calls out).

The regression test doesn't cover these because the waiter thread is already past line 76 and inside the `wait()` loop when the poisoner runs, so a fresh `pop_blocking` call after poisoning would still silently return `None`. Suggest applying the same `unwrap_or_else(PoisonError::into_inner)` recovery at lines 67 and 76, and extending the test (or adding a second case) to call `pop_blocking` / `push` on the already-poisoned queue.

Out of scope for this PR but noting for follow-up: `lib.rs:950` uses `.lock().ok().map_or_else(...)` for observation state — probably intentional as a read-only fallback, but worth a look when the broader poison-safety sweep happens.

slepp:
Re-reviewed at f98471a. All three EventQueue sites (push, pop_blocking initial lock, condvar wait) now recover via PoisonError::into_inner. The read-only map_or_else site is untouched as intended. Three new regression tests cover each path and assert the recovered queue state. Scope is limited to std/net/quic/src/lib.rs. Voice guard passes locally and in CI.

PR #1458, opened 2026-04-21
fix(std/encoding/json): return null on hew_json_stringify serializer error

### LINKED ISSUE #1455: std/encoding/json: hew_json_stringify collapses serializer error into empty string
## Summary

`std/encoding/json/src/lib.rs:111` uses `serde_json::to_string(&v.inner).unwrap_or_default()`, so any serialization failure (e.g., a map with non-string keys, a NaN float, a value the serializer rejects) is silently collapsed into an empty string. A `malloc`'d empty C string is indistinguishable from a legitimate `""` encoding.

This violates CLAUDE.md §2 (Fail-Closed Codegen) and is a wire-protocol integrity hazard because `hew_json_stringify` is on the outgoing RPC hot path.

## Location

`std/encoding/json/src/lib.rs:111`

## Proposed fix

On `Err`, return `std::ptr::null_mut()` — null is already the documented error sentinel for other stringify-family exports in this crate.

## Test approach

Unit test: construct a `HewJsonValue` containing a value `serde_json` refuses to serialize (e.g., an f64 NaN via the appropriate input path), call `hew_json_stringify`, assert the return is null.

## Evidence

R6 workspace audit F3. Similar to #1418 (yaml) and #1441 (json `get_bytes`) family.

slepp:
Fixes #1455. hew_json_stringify used \.unwrap_or_default() on the serde_json result, so a rejected value collapsed into an empty C string indistinguishable from a legitimate empty JSON string. Now returns null_mut() on error, matching the null-on-error sentinel used elsewhere in this file. Adds a regression test for the NaN f64 path.

### DIFF
--- a/std/encoding/json/src/lib.rs
+++ b/std/encoding/json/src/lib.rs
@@ -50,6 +50,19 @@ fn get_parse_last_error() -> String {
     LAST_PARSE_ERROR.with(|error| error.borrow().clone().unwrap_or_default())
 }
 
+fn stringify_result_to_malloc(result: Result<String, serde_json::Error>) -> *mut c_char {
+    match result {
+        Ok(json) => {
+            clear_parse_last_error();
+            str_to_malloc(&json)
+        }
+        Err(err) => {
+            set_parse_last_error(format!("json stringify failed: {err}"));
+            std::ptr::null_mut()
+        }
+    }
+}
+
 // ---------------------------------------------------------------------------
 // C ABI exports
 // ---------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


slepp:
Approving. Fail-closed conversion is clean: `null_mut()` on serializer error with `LAST_PARSE_ERROR` populated via the existing `set_parse_last_error` helper, matching the sentinel pattern used by `hew_json_get_string` and siblings. Regression test asserts both the null return and a descriptive message in `hew_json_last_error()`. Scope is a single file, no `.unwrap_or_default()` / `.ok()?` reintroduced. Minor nit (non-blocking): PR body mentions NaN f64 but the test actually exercises a non-string map key — either path triggers the serializer error, so the invariant is covered.

PR #1462, opened 2026-04-21
fix(std/encoding/yaml): cap anchor/alias expansion and input size to block billion-laughs DoS

### LINKED ISSUE #1459: std/encoding/yaml: hew_yaml_parse has no anchor/alias expansion cap (billion-laughs DoS)
## Summary

`std/encoding/yaml/src/lib.rs:76` calls `serde_yaml::from_str::<serde_yaml::Value>(s)` with no cap on anchor/alias expansion. A classic ~100-byte "billion laughs" YAML (9-level nested anchor references) can expand to gigabytes in memory during parse.

## Location

`std/encoding/yaml/src/lib.rs:76`

## Risk

A single HTTP request body pointed at `hew_yaml_parse` can OOM-kill the process.

## Proposed fix

Pre-scan input for anchor (`&`) and alias (`*`) counts, or apply a byte-size cap (e.g. 1 MiB default) on the input before handing it to `serde_yaml`. At minimum reject inputs with >32 anchors or >1024 aliases.

Populate `LAST_PARSE_ERROR` (or the yaml crate's equivalent thread-local) with a descriptive message mentioning the limit that was hit.

## Test approach

Unit test: feed the classic 9-level "lol" billion-laughs YAML, assert `hew_yaml_parse` returns null and error message mentions alias/anchor limit. Also assert a normal YAML with a few legitimate anchors (say 3 anchors, 5 aliases) still parses successfully.

## Evidence

R6 workspace audit F4. Similar to #1411 (compress bomb) family.

slepp:
Fixes #1459. `hew_yaml_parse` forwarded untrusted input directly to `serde_yaml::from_str`, which has no built-in anchor/alias expansion cap; a ~100-byte billion-laughs YAML could OOM the process. Adds a conservative pre-scan that rejects inputs with >32 anchors, >1024 aliases, or >1 MiB total size, populating the thread-local parse error with a descriptive message. Legitimate YAML with a handful of anchors/aliases is unaffected; tests cover both the DoS payload and normal-use baselines.

### DIFF
--- a/std/encoding/yaml/src/lib.rs
+++ b/std/encoding/yaml/src/lib.rs
@@ -18,6 +18,13 @@ use hew_cabi::{
 use std::ffi::CStr;
 use std::os::raw::c_char;
 
+/// Reject YAML inputs larger than 1 MiB before parsing to avoid memory abuse.
+const YAML_PARSE_SIZE_LIMIT_BYTES: usize = 1024 * 1024;
+/// Reject YAML inputs declaring more than 32 anchors before parsing.
+const YAML_PARSE_ANCHOR_LIMIT: usize = 32;
+/// Reject YAML inputs referencing more than 1024 aliases before parsing.
+const YAML_PARSE_ALIAS_LIMIT: usize = 1024;
+
 /// Opaque wrapper around a [`serde_yaml::Value`].
 ///
 /// Returned by [`hew_yaml_parse`], [`hew_yaml_get_field`], and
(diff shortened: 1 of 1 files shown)


slepp:
Reviewed the prescan. Limits are enforced before `serde_yaml::from_str` gets the input, size cap reads `input.len()` up front, and anchor/alias counting tracks single- and double-quote state with a backslash-skip inside double quotes so quoted sigils don't inflate the counts. Reject paths return `null_mut()` after populating the thread-local error with a message that names which cap was hit (size/anchor/alias). Tests cover the alias bomb, the oversize input, a realistic 3-anchor / 5-alias merge chain, and the quoted-sigil false-positive guard. No `.ok()?` or `.unwrap_or_default()` in the new code, scope stays inside `std/encoding/yaml/src/lib.rs`, and nothing is allocated on the reject path. Queuing auto-squash.

PR #1463, opened 2026-04-21
fix(std/encoding/msgpack): enforce recursion-depth cap to prevent stack-overflow DoS

### LINKED ISSUE #1460: std/encoding/msgpack: decoder has no recursion-depth cap (stack-overflow DoS)
## Summary

`std/encoding/msgpack/src/lib.rs:92` calls `rmp_serde::from_slice::<serde_json::Value>(slice)` with no recursion-depth cap. Adversarial input composed of nested arrays (opcode 0x91 repeated thousands of times) blows the stack before any fail-closed path can fire.

## Location

`std/encoding/msgpack/src/lib.rs:92`

## Risk

One malformed msgpack frame from an untrusted peer aborts the Hew process via SIGSEGV on stack guard. No recovery is possible because this is a stack overflow, not an `Err` — the process is dead.

## Proposed fix

Either:
1. Pre-scan opcode bytes to reject nesting depth > 128 before calling the deserializer, OR
2. Wrap the deserialize in a custom `Deserializer` that tracks depth and returns `Err` past ~128 levels.

Option 1 (opcode pre-scan) is simpler and has no serde-version coupling.

## Test approach

Unit test: construct a 10 000-byte input of 0x91 repeated; call the parse entry point; assert a null return with a descriptive error, not a crash. Run under the normal test harness (if the test process crashes, the fix is wrong).

## Evidence

R6 workspace audit F5. Similar to #1411 (bomb family).

slepp:
Fixes #1460. `rmp_serde::from_slice` has no recursion cap; an input composed of nested array/map opcodes (a few KiB) could stack-overflow the deserializer and abort the process. Adds an opcode-level pre-scan that rejects inputs exceeding 128 nesting levels before handing them to the deserializer, populating the thread-local error with a descriptive message. Legitimate nested msgpack (tested to depth 5) is unaffected.

### DIFF
--- a/std/encoding/msgpack/src/lib.rs
+++ b/std/encoding/msgpack/src/lib.rs
@@ -12,7 +12,164 @@
 extern crate hew_runtime;
 
 use hew_cabi::cabi::{cstr_to_str, str_to_malloc};
+use std::cell::RefCell;
 use std::ffi::c_char;
+
+const MAX_MSGPACK_DEPTH: usize = 128;
+const MALFORMED_MSGPACK_ERROR: &str = "msgpack: malformed input during depth pre-scan";
+
+std::thread_local! {
+    static LAST_MSGPACK_ERROR: RefCell<Option<String>> = const { RefCell::new(None) };
+}
+
+fn set_msgpack_last_error(msg: impl Into<String>) {
+    LAST_MSGPACK_ERROR.with(|error| *error.borrow_mut() = Some(msg.into()));
+}
+
+fn clear_msgpack_last_error() {
+    LAST_MSGPACK_ERROR.with(|error| *error.borrow_mut() = None);
+}
+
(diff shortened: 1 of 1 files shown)


slepp:
Pre-scan opcode coverage is exhaustive and payload skips line up with the MessagePack spec (fixext sizes, ext8/16/32 = len+1, map variants = N*2 slots). Depth tracking is a real stack — verified via the seed-of-1 encoding — so `[1,[2,3]]` reports depth 2, not 4. The cap guards the only production `from_slice` call, and failures populate the thread-local error before returning null. Regression tests cover the 10 000× fixarray and 200× fixmap DoS vectors plus a depth-5 happy path and empty array; local suite is green 3× in a row. Voice, clippy, and format are green. Auto-merge enabled.

PR #1464, opened 2026-04-21
fix(std/encoding/xml): cap parse nesting depth to block stack-overflow DoS

### LINKED ISSUE #1461: std/encoding/xml: parser has no nesting-depth cap or entity-expansion guard (DoS)
## Summary

`std/encoding/xml/src/lib.rs:102` uses `quick_xml::Reader::from_str(xml)` with no nesting-depth cap, and recursive `XmlNodeKind::Element { children, .. }` construction can stack-overflow on deeply nested input. External entity expansion is also not explicitly disabled.

## Location

`std/encoding/xml/src/lib.rs:102` and the surrounding recursive element-build loop.

## Risk

Adversarial XML (e.g. 10 000 `<x>` opens) crashes the process via stack overflow. Billion-laughs-in-XML (external entity refs) is a latent DoS unless disabled explicitly.

## Proposed fix

1. Track depth in the build loop and error out once it exceeds a configurable limit (e.g. 256 default).
2. On the `quick_xml::Reader` config, explicitly disable external entity resolution and set any available limits.
3. Populate the appropriate thread-local error sentinel with a descriptive message on limit breach.

## Test approach

Unit tests for:
- 10 000-deep `<x>...</x>` nesting → null return, error mentions depth limit.
- Entity-expansion document (external entity ref) → null return, error mentions entity handling.
- Normal XML with reasonable nesting (say 5 levels) → parses successfully.

## Evidence

R6 workspace audit F6.

slepp:
Fixes #1461. The XML parse entry point recursively built element trees without a depth cap, so an input of 10,000 nested `<x>` tags could stack-overflow the process. Adds a depth counter to the builder that errors cleanly past 256 levels, populating the thread-local parse error. Also documents the quick-xml entity-handling behavior in the reader config. Legitimate XML with reasonable nesting is unaffected.

### DIFF
--- a/std/encoding/xml/src/lib.rs
+++ b/std/encoding/xml/src/lib.rs
@@ -11,8 +11,27 @@
 extern crate hew_runtime;
 
 use hew_cabi::cabi::str_to_malloc;
+use std::cell::RefCell;
 use std::ffi::CStr;
 use std::os::raw::c_char;
+
+const MAX_XML_DEPTH: usize = 256;
+
+std::thread_local! {
+    static LAST_XML_ERROR: RefCell<Option<String>> = const { RefCell::new(None) };
+}
+
+fn set_xml_last_error(msg: impl Into<String>) {
+    LAST_XML_ERROR.with(|error| *error.borrow_mut() = Some(msg.into()));
+}
+
+fn clear_xml_last_error() {
+    LAST_XML_ERROR.with(|error| *error.borrow_mut() = None);
+}
+
+fn clone_xml_last_error() -> Option<String> {
+    LAST_XML_ERROR.with(|error| error.borrow().clone())
+}
 
 // ---------------------------------------------------------------------------
 // Internal tree representation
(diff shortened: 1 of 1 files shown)


slepp:
Reviewed. Depth cap is correctly positioned at read-event time before the `stack.push`, so the 10k-deep test exits cleanly at depth 257 without any recursion — the builder was already iterative with an explicit `Vec<Frame>`, so there is no hidden recursive path that could overflow before the check fires. Entity-handling comment accurately reflects quick-xml 0.39 behavior (GeneralRef is surfaced, not expanded), and the `general_entity_references_are_not_expanded` test pins that. Reject path drops `Vec<Frame>` naturally with no partial `boxed_node` allocations to leak. Scope is clean (xml/src/lib.rs only).

Minor note (non-blocking): `hew_xml_last_error` returns null on no-error, whereas `hew_http_last_error` returns an empty malloc'd string. The divergence is documented on the function and works fine for C callers, but worth aligning in a future pass if we want one canonical shape.

PR #1467, opened 2026-04-26
test(std/encoding/json): cover hew_json_get_bytes last_error semantics

### LINKED ISSUE #1441: std/encoding/json: hew_json_get_bytes silently returns empty Vec on malformed base64

slepp:
Fixes #1441.

## Summary
- add regression coverage for `hew_json_get_bytes` malformed base64 handling
- verify valid and empty base64 inputs leave `hew_json_last_error()` clear
- verify a successful decode clears a prior error and that JSON last-error state stays thread-local

## Validation
- `cargo test -p hew-std-encoding-json --lib --quiet`
- `make ci-preflight`

### DIFF
--- a/std/encoding/json/src/lib.rs
+++ b/std/encoding/json/src/lib.rs
@@ -85,9 +85,9 @@ pub unsafe extern "C" fn hew_json_parse(json_str: *const c_char) -> *mut HewJson
     }
 }
 
-/// Return the last JSON parse error recorded on the current thread.
+/// Return the last JSON error recorded on the current thread.
 ///
-/// Returns an empty string when no parse error has been recorded.
+/// Returns an empty string when no error has been recorded.
 #[no_mangle]
 pub extern "C" fn hew_json_last_error() -> *mut c_char {
     str_to_malloc(&get_parse_last_error())
@@ -1428,6 +1428,7 @@ mod tests {
 
             let bytes = read_and_free_bytes(hew_json_get_bytes(field));
             assert!(bytes.is_empty());
+            assert!(read_and_free_cstr(hew_json_last_error()).is_empty());
 
             hew_json_free(field);
             hew_json_free(val);
@@ -1435,7 +1436,7 @@ mod tests {
     }
 
     #[test]
-    fn get_bytes_valid_base64_round_trips() {
+    fn get_bytes_valid_base64_round_trips_without_last_error() {
         clear_parse_last_error();
         let val = parse(r#"{"b":"aGVsbG8="}"#);
         assert!(!val.is_null());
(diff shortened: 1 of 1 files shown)


slepp:
## Review: PR #1467 — `fix/1441-json-get-bytes-fail-closed`

**Reviewer**: independent cross-ecosystem review (claude-opus)

### What I verified

1. **Impl agent claim** — confirmed correct. `hew_json_get_bytes` (lib.rs:282–304) already returns `null_mut()` on malformed base64 and populates `LAST_PARSE_ERROR` via `set_parse_last_error`. The issue #1441 description was outdated; the fail-closed behavior was already in place on `main`. This PR is regression-test-only.

2. **Test coverage** — all five scenarios are present and sound:
   - Valid base64 → correct bytes, no error
   - Empty base64 (`""`) → empty vec, no error
   - Malformed base64 → null return, `hew_json_last_error()` contains `"base64"` or `"decode"`
   - Successful decode after a prior error → error cleared
   - Thread-local isolation → error on thread A does not leak to thread B or the main thread

3. **Flake risk** — none. `LAST_PARSE_ERROR` is `thread_local!`, each test calls `clear_parse_last_error()` at the top, and the thread-local test uses `std::thread::spawn` (fresh TLS per thread). No `#[serial]` annotation needed.

4. **Tests pass** — `cargo test -p hew-std-encoding-json --lib -- get_bytes` runs 5 tests, all green.

5. **Voice** — title, body, commit message, and code are clean. No forbidden tokens.

### Issue: PR title uses `fix()` prefix for a test-only change

The PR title is `fix(std/encoding/json): hew_json_get_bytes returns null on malformed base64`, which implies a behavior change. The actual commit on the branch correctly uses `test()`: `test(std/encoding/json): cover get_bytes last_error semantics`. Since conventional-commit prefix drives changelog generation, the PR title should match reality. Suggest updating to:

```
test(std/encoding/json): lock in hew_json_get_bytes fail-closed base64 behavior
```

or similar with a `test()` prefix.

### Convention note: `*_last_error()` return-on-no-error drift

`hew_json_last_error()` returns a malloc-allocated empty C string (`""`) when no error is recorded (via `str_to_malloc` on `unwrap_or_default()`). This matches `hew_yaml_last_error` and `hew_tls_last_error`. However, `hew_xml_last_error` returns `null_mut()` on no-error. This convention split is pre-existing and not introduced here — just documenting for the record.

### Pre-existing observation (not blocking)

The `None` branch at lib.rs:303 (non-string JSON value passed to `hew_json_get_bytes`) returns `null_mut()` without setting *or clearing* the error. If a prior call left an error in the thread-local, this path leaves it stale. Not introduced by this PR, but worth a follow-up issue if the project wants consistent error-state semantics.

### Verdict

Title needs correction from `fix()` to `test()` to avoid a misleading changelog entry. Code and tests are sound — no bugs, no flake risk, coverage is thorough.

slepp:
Updated the PR title to reflect that this change adds regression coverage for existing behavior rather than changing runtime semantics.

slepp:
Re-checked the PR title — it now reads `test(std/encoding/json): cover hew_json_get_bytes last_error semantics`, which correctly reflects the test-only scope of this change. That was the sole finding from the initial review; everything else (coverage of all five scenarios, thread-local isolation, no flake risk) was already clean.

This is good to merge.

PR #1469, opened 2026-04-26
fix(codegen): handle dyn trait e2e regressions

### LINKED ISSUE #1465: P0: dyn_trait e2e tests failing on main — blocks local preflight for unrelated PRs
## Summary

Two hew-codegen e2e tests are failing on `origin/main` (currently 85c2eddc) and blocking local pre-push hooks for unrelated PRs:

- `reject_e2e_negative_dyn_trait_generic_method_reject`
- `e2e_dyn_trait_dyn_coerce_type_paths`

Confirmed via fresh worktree on `origin/main`: `make codegen && ctest --test-dir hew-codegen/build -R 'dyn_trait_generic_method_reject|dyn_coerce_type_paths'` fails in both cases with no diff applied.

## Impact

P0 — every open PR's local `make ci-preflight` is red through no fault of the diff. Forces contributors to either bypass hooks or block indefinitely. Currently affects #1457 (quic poison-safe) and has impacted preflight on multiple in-flight lanes today.

## Likely cause

A recent merge to `main` introduced a regression in the `dyn` trait coercion path. Suspects to bisect:
- Recent commits touching codegen, trait lowering, or dyn/coerce logic.
- `git log --oneline -20 origin/main` and run the tests at each boundary to find the culprit.

## Proposed fix

1. `git bisect` between the most recent green merge and current main on those two tests.
2. Revert or fix forward depending on what caused the regression.
3. Add a narrow-lane preflight entry (or fix the codegen-tests lane routing) so these tests run on every codegen-touching PR, not only as a post-merge orphan.

## Urgency

P0 — blocks the dev loop. Should be triaged before the next wave of merges.

slepp:
## Summary
- reject dyn trait generic method calls at the trait-object dispatch site before MLIR type conversion reaches unresolved method type params
- honor annotated array element types when lowering array literals so [dyn Speak; N] coercions build trait objects element-by-element
- keep the fix scoped to the two failing dyn trait e2e cases on main

## Validation
- ctest --test-dir hew-codegen/build --output-on-failure -R 'reject_e2e_negative_dyn_trait_generic_method_reject|e2e_dyn_trait_dyn_coerce_type_paths'
- ctest --test-dir hew-codegen/build --output-on-failure after make wasm-runtime (1255/1258 passing; remaining failures are unrelated existing wasm cases: wasm_e2e_timeout_timeout, wasm_e2e_toml_toml_try_parse, wasm_e2e_actors_actor_periodic_timer)
- make ci-preflight

Fixes #1465

### DIFF
--- a/hew-codegen/src/mlir/MLIRGenExpr.cpp
+++ b/hew-codegen/src/mlir/MLIRGenExpr.cpp
@@ -4562,7 +4562,15 @@ mlir::Value MLIRGen::generateMethodCall(const ast::ExprMethodCall &mc, const ast
       if (idxIt != traitIt->second.methodIndex.end())
         methodIdx = idxIt->second;
       for (const auto *tm : traitIt->second.methods) {
-        if (tm->name == methodName && tm->return_type.has_value()) {
+        if (tm->name != methodName)
+          continue;
+        if (tm->type_params && !tm->type_params->empty()) {
+          ++errorCount_;
+          emitError(location) << "dyn-trait dispatch does not yet support generic methods: "
+                              << methodName;
+          return nullptr;
+        }
+        if (tm->return_type.has_value()) {
           returnType = convertType(tm->return_type->value);
           break;
         }
@@ -4869,6 +4877,7 @@ mlir::Value MLIRGen::generateTupleExpr(const ast::ExprTuple &tup,
 mlir::Value MLIRGen::generateArrayExpr(const ast::ExprArray &arr,
                                        std::optional<mlir::Type> typeHint) {
   auto location = currentLoc;
+  auto targetArrayType = mlir::dyn_cast_or_null<hew::HewArrayType>(typeHint.value_or(mlir::Type{}));
 
   if (arr.elements.empty()) {
     // Empty array literal: coerce to Vec<T> if type context expects it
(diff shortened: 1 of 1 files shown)


slepp:
## Independent code review — verdict: **pass**

Reviewed the diff, read full context of both change sites, inspected both test files + expected output, and ran targeted ctest (2/2 passed).

### Generic method rejection (L4565)
Correct and fail-closed. The old code silently fell through when a dyn-trait method had type params, letting unresolved generics reach trait dispatch. The new early guard checks `type_params`, emits a clear diagnostic, increments `errorCount_`, and returns `nullptr`. No risk of false positives — the `continue` / guard structure preserves behavior for non-generic methods.

### Typed array literal coercion (L4919)
Correct and fail-closed. When the type hint is `HewArrayType`, the new path validates length, passes `targetElemType` as a hint to each element expression (enabling proper trait-object construction), coerces, and validates. All error paths emit diagnostics and return `nullptr`. This only activates with an explicit array type annotation — it cannot narrow programs that currently compile, since the fallback path infers element type from the first element (which would already fail for dyn coercions).

### Tests
- `dyn_coerce_type_paths.hew` exercises 4 distinct coercion sites (let, field, return, array). Good coverage.
- `dyn_trait_generic_method_reject.hew` properly tests generic method dispatch on a multi-trait dyn object with both generic (`Transform`) and non-generic (`Label`) traits.

### Minor observations (non-blocking)
1. Commit message only mentions the rejection fix — the array coercion is described in the PR body, which is fine.
2. Pre-existing: the Vec coercion loop (~L4905) uses `break;` on coercion failure instead of `return nullptr;`, returning a partially-populated Vec. Not introduced here — worth a follow-up ticket.

**Recommendation: merge.**

Issue #1468, opened 2026-04-26
std/encoding/json: hew_json_get_bytes None branch leaves stale LAST_PARSE_ERROR


slepp:
## Bug

In `std/encoding/json/src/lib.rs`, `hew_json_get_bytes` has three exit paths:

1. **Valid base64 string** (line 299): calls `clear_parse_last_error()` then returns the decoded `HewVec`. ✓
2. **Invalid base64 string** (line 293–296): calls `set_parse_last_error(…)` then returns `null_mut()`. ✓
3. **Non-string value (`None` match arm, line 303)**: returns `null_mut()` **without calling `clear_parse_last_error()` or `set_parse_last_error(…)`**.

```rust
match v.inner.as_str() {
    Some(s) => {
        // … decode, clear/set error …
    }
    None => std::ptr::null_mut(),  // ← no error state mutation
}
```

The same gap exists for the null-pointer early return at line 283–284.

### Symptom

If a caller previously triggered a `set_parse_last_error` (e.g. by passing malformed base64), then later calls `hew_json_get_bytes` with a non-string value, `hew_json_last_error()` still returns the **stale** error message from the earlier call. The caller has no way to distinguish "this failed with an error" from "this returned null for a non-error reason".

### Suggested fix

Call `clear_parse_last_error()` on both the `None` branch (line 303) and the null-pointer early return (line 283–284), so that every non-error exit path leaves the thread-local error state clean.

### Broader context

This is a sibling pattern to convention drift across the std encoding crates:
- **json**, **yaml**, **tls**: `*_last_error()` returns a `malloc`-ed empty string on no-error.
- **xml**: `hew_xml_last_error()` returns `null_mut()` on no-error.

A unification lane to align the `*_last_error()` convention and audit all `None`/null early-return paths across std crates is desirable but separate from this fix.

PR #1470, opened 2026-04-26
fix(std/encoding/yaml): return null and record error on base64 decode failure in get_bytes

### LINKED ISSUE #1418: std/encoding/yaml: hew_yaml_get_bytes silently returns empty on base64 decode failure

slepp:
Closes #1418

`hew_yaml_get_bytes` previously used `unwrap_or_default()` on base64 decode, silently producing a zero-length `HewVec` indistinguishable from a legitimately empty bytes value. This change returns `null` and records the failure in `LAST_PARSE_ERROR` (consistent with the rest of the module).

Also addresses the stale-error class of bug seen in #1468: the None branch (missing key / wrong type) now ensures `LAST_PARSE_ERROR` reflects the current call rather than a prior parse failure.

### Test
- Malformed base64 → null + populated last-error
- Missing key → null + non-stale last-error
- Valid base64 → bytes returned correctly
- `make ci-preflight` clean.

### DIFF
--- a/std/encoding/yaml/src/lib.rs
+++ b/std/encoding/yaml/src/lib.rs
@@ -180,9 +180,11 @@ pub unsafe extern "C" fn hew_yaml_parse(yaml_str: *const c_char) -> *mut HewYaml
     }
 }
 
-/// Return the last YAML parse error recorded on the current thread.
+/// Return the last YAML error recorded on the current thread.
 ///
-/// Returns an empty string when no parse error has been recorded.
+/// This slot is shared by parse failures and byte-extraction failures from
+/// [`hew_yaml_get_bytes`]. Returns an empty string when the most recent
+/// operation succeeded or explicitly cleared the error slot.
 #[no_mangle]
 pub extern "C" fn hew_yaml_last_error() -> *mut c_char {
     str_to_malloc(&get_parse_last_error())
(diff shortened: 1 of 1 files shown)


slepp:
**Cross-ecosystem review — pass, recommend merge.**

All four return paths in `hew_yaml_get_bytes` now properly manage the error slot — no silent null returns, no stale `LAST_PARSE_ERROR` leakage.

Verified the stale-error overwrite test (`get_bytes_missing_key_overwrites_stale_last_error`): it genuinely poisons the error slot with a base64 failure first, then confirms the missing-key path overwrites it with a fresh message. This is a proper two-phase test, not just "is the error non-empty."

56 tests pass, clippy clean.

**One informational note on JSON parity:** this YAML fix is actually stricter than the JSON sibling from PR #1467 — `hew_json_get_bytes` still silently returns null on null/non-string input without populating the error slot. That gap is tracked in #1468 and does not block this PR.

PR #1471, opened 2026-04-26
fix(std/encoding/compress): cap decompression output to prevent zip-bomb OOM

### LINKED ISSUE #1411: std/encoding/compress: no decompression-bomb limit

slepp:
Closes #1411

`read_to_malloc` previously called `reader.read_to_end` with no upper bound. A small gzip / deflate / zlib bomb expanding to GB caused OOM rather than a clean failure, exposing `std/net/http` server bodies (which can flow through this helper) to DoS.

This change adds an explicit `max_output_len` parameter at the FFI boundary. The reader is wrapped with `.take(max_output_len + 1)`; if the buffer exceeds the cap, the partial allocation is freed and the call returns null with an appropriate last-error message.

A conservative default cap is documented in the module header.

### Test
- 1 KB → 10 MB bomb with cap=1 MB returns null without OOM
- Normal decompress within cap continues to succeed
- `make ci-preflight` clean.

### DIFF
--- a/std/encoding/compress/compress.hew
+++ b/std/encoding/compress/compress.hew
@@ -1,7 +1,10 @@
 //! Compression and decompression (gzip, deflate, zlib).
 //!
 //! Compress and decompress binary data using gzip, raw deflate, or
-//! zlib formats. All functions take and return `bytes` values.
+//! zlib formats. Compression helpers take and return `bytes` values.
+//! Decompression callers must pass an explicit `max_output_len` to fail
+//! closed on oversized payloads; a conservative starting point is
+//! `64 * 1024 * 1024`.
 //!
 //! # Examples
 //!
@@ -16,7 +19,7 @@
 //!     data.push(108);
 //!     data.push(111);
 //!     let compressed = compress.gzip_compress(data);
-//!     let decompressed = compress.gzip_decompress(compressed);
+//!     let decompressed = compress.gzip_decompress(compressed, 64 * 1024 * 1024);
 //!     println(decompressed.len());  // 5
 //! }
 //! ```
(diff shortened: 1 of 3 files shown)


slepp:
### Cross-ecosystem review — pass ✅

**Cap semantics are correct.** `.take(cap+1)` with `buf.len() > cap` has no off-by-one. `saturating_add` handles the `usize::MAX` edge cleanly. `max_output_len == 0` correctly allows only empty output.

**Drop safety is sound.** The `Vec` buf is RAII-managed and dropped on all error paths; `libc::malloc` is deferred until after the size check, so the bomb-detected path never leaks.

**FFI signatures are consistent end-to-end** — Rust `extern "C"`, HewVec wrappers, `compress.hew` public API and externs, and `compress_test.hew` callers all updated. Negative `max_output_len` from Hew is caught by `hew_max_output_len`.

**Tests pass:** `cargo test -p hew-std-encoding-compress` (9/9), clippy clean.

Two non-blocking nits for follow-up:
1. Lines 500–501: duplicate `/// max_output_len must be non-negative.` doc on `hew_gzip_compress_hew`, which doesn't take that parameter.
2. Only gzip bomb detection is tested. Adding a deflate or zlib bomb variant would strengthen coverage for a security fix (the code path is shared via `read_to_malloc`, so it's functionally equivalent — this is about test documentation).

Recommendation: **merge**.

PR #1472, opened 2026-04-26
fix(std/encoding/json): clear stale last-error on non-string and null paths in get_bytes

### LINKED ISSUE #1468: std/encoding/json: hew_json_get_bytes None branch leaves stale LAST_PARSE_ERROR

slepp:
Closes #1468

Mirrors the YAML fix in #1470. The None branch (non-string value) and the early-return null-pointer branch in `hew_json_get_bytes` previously left `LAST_PARSE_ERROR` untouched, so callers seeing a null return would observe a stale error message from a prior call.

This change writes the per-call state on every exit path, matching the convention adopted in #1470 for YAML.

### Test
- Stale-error overwrite: poison with malformed base64, then call with non-string, assert error reflects current call.
- `make ci-preflight` clean.

### DIFF
--- a/std/encoding/json/src/lib.rs
+++ b/std/encoding/json/src/lib.rs
@@ -87,7 +87,9 @@ pub unsafe extern "C" fn hew_json_parse(json_str: *const c_char) -> *mut HewJson
 
 /// Return the last JSON error recorded on the current thread.
 ///
-/// Returns an empty string when no error has been recorded.
+/// This slot is shared by parse failures and byte-extraction failures from
+/// [`hew_json_get_bytes`]. Returns an empty string when the most recent
+/// operation succeeded or explicitly cleared the error slot.
 #[no_mangle]
 pub extern "C" fn hew_json_last_error() -> *mut c_char {
     str_to_malloc(&get_parse_last_error())
@@ -271,9 +273,10 @@ pub unsafe extern "C" fn hew_json_get_string(val: *const HewJsonValue) -> *mut c
 
 /// Get a base64-decoded bytes value from a [`HewJsonValue`].
 ///
-/// Returns a newly allocated [`HewVec`] for valid string inputs. Non-string
-/// values return null. Invalid base64 inputs return null and populate
-/// [`hew_json_last_error`].
+/// Returns a newly allocated [`HewVec`] for valid string inputs. Null,
+/// non-string, or invalid base64 inputs return null and overwrite
+/// [`hew_json_last_error`] so callers never observe a stale error from a prior
+/// parse or decode failure.
 ///
 /// # Safety
 ///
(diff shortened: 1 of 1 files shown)


slepp:
**Review — request one small revision**

Code logic ✅ — the fix is correct and structurally mirrors the YAML counterpart from #1470. All three exit paths in `hew_json_get_bytes` now write `LAST_PARSE_ERROR`:

| Path | Before | After |
|---|---|---|
| null input | silent `null_mut()` | `set_parse_last_error(…)` + `null_mut()` |
| non-string value | silent `null_mut()` | `set_parse_last_error(…)` + `null_mut()` |
| base64 decode error | ✓ already set | ✓ unchanged |
| valid decode | ✓ already cleared | ✓ unchanged |

Both regression tests correctly poison-then-overwrite (not just check non-empty), and all 54 tests pass with `clippy -D warnings` clean.

---

**One doc-comment parity drift with #1470** — the YAML PR updated two doc-comments that this PR should mirror:

1. **`hew_json_last_error`** — should note the slot is shared by parse and byte-extraction failures (matching the YAML doc update).

2. **`hew_json_get_bytes`** — the doc still says _"Non-string values return null. Invalid base64 inputs return null and populate…"_ but null inputs now also set the error. The YAML version was updated to _"Null, non-string, or invalid base64 inputs return null and overwrite `hew_yaml_last_error`…"_.

Once those two doc-comments are aligned, this looks good to merge.

slepp:
Updated docstrings to match the YAML precedent from #1470. Ready for re-review.

Issue #1420, opened 2026-04-20
LAST_PARSE_ERROR thread-locals drift across async tasks in datetime/yaml/toml


slepp:
Three independently-defined thread-local error slots:
- `std/time/datetime/src/lib.rs:20–37`
- `std/encoding/yaml/src/lib.rs:40–55`
- `std/encoding/toml/src/lib.rs:34–50`

Each exposes a `hew_*_last_error` accessor. Hew actors run on a pooled cooperative scheduler; an actor may be parked after `parse` and resumed on a different worker, at which point `last_error` reads another actor's state — or is empty because the resumed worker never touched it. No runtime test exercises cross-thread semantics. Duplication itself is three independently-drifting implementations.

## Check-list
- [ ] Replace with a per-call out-parameter (preferred — see the datetime sentinel issue) or a handle-scoped error slot
- [ ] Where a free-function API is required, key by `hew_runtime::current_actor_id()` so semantics track the Hew actor model
- [ ] Consolidate the three implementations into one shared helper
- [ ] Cross-thread test: parse → (simulate actor migration) → `last_error` still returns the right value

PR #1473, opened 2026-04-26
fix(std/net/tls): record connect errors via hew_tls_last_error

### LINKED ISSUE #1414: std/net/tls: hew_tls_connect discards every error class as NULL
`std/net/tls/src/lib.rs:74–87` — `match connect_tls(host_str, port_u16) { Ok(s) => …, Err(_) => std::ptr::null_mut() }` (line 85). The `BoxError` is dropped. `default_client_config`, `ServerName::try_from`, `TcpStream::connect`, and `rustls::ClientConnection::new` all fold into the same null. Unlike `std/net/http/src/client.rs` (forwards status codes), TLS has no `hew_tls_last_error` API.

## Check-list
- [ ] Add a thread-local error stash analogous to `hew_datetime_last_error`
- [ ] Record `BoxError::to_string()` before returning null
- [ ] Consider extracting the `last_error` pattern into a shared helper (see the LAST_PARSE_ERROR drift issue)

slepp:
Closes #1414

`hew_tls_connect` previously folded every error class — `default_client_config`, `ServerName::try_from`, `TcpStream::connect`, `rustls::ClientConnection::new` — into a single null return, dropping the `BoxError` on the floor.

This change adds a thread-local error stash and a `hew_tls_last_error` FFI (caller-frees, consistent with sibling crates such as `hew_datetime_last_error`). Each error path now records the error string before returning null. Success paths do not leave a stale error.

### Test
- Bad hostname → null + last-error mentions hostname
- Unreachable TCP → null + last-error mentions connect
- Success after prior failure → no stale error
- `make ci-preflight` clean.

### DIFF
--- a/std/net/tls/src/lib.rs
+++ b/std/net/tls/src/lib.rs
@@ -58,12 +58,16 @@ fn default_client_config() -> Result<rustls::ClientConfig, BoxError> {
     Ok(config)
 }
 
-fn connect_tls(host: &str, port: u16) -> Result<HewTlsStream, BoxError> {
-    let server_name = ServerName::try_from(host.to_string())?;
-    let config = default_client_config()?;
-    let connector = rustls::ClientConnection::new(Arc::new(config), server_name)?;
+fn connect_tls(host: &str, port: u16) -> Result<HewTlsStream, String> {
+    let server_name = ServerName::try_from(host.to_string())
+        .map_err(|err| format!("invalid TLS hostname `{host}`: {err}"))?;
+    let config = default_client_config()
+        .map_err(|err| format!("failed to build TLS client config: {err}"))?;
+    let connector = rustls::ClientConnection::new(Arc::new(config), server_name)
+        .map_err(|err| format!("failed to create TLS client connection for `{host}`: {err}"))?;
     let addr = format!("{host}:{port}");
-    let tcp = TcpStream::connect(addr)?;
+    let tcp =
+        TcpStream::connect(&addr).map_err(|err| format!("failed to connect to {addr}: {err}"))?;
     let stream = rustls::StreamOwned::new(connector, tcp);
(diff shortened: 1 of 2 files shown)


slepp:
### Cross-ecosystem review — claude-opus-4.6

**Verdict: request revision** — one blocking item.

---

**Rust FFI implementation: looks good.** All five error paths in `hew_tls_connect` record a descriptive message before returning null. `hew_tls_last_error` matches the `hew_datetime_last_error` convention exactly (caller-frees via `libc::free`, null when no error). Drop safety is clean — `str_to_malloc` delegates to `malloc_cstring`, success path uses `Box::into_raw`/`Box::from_raw` correctly. The `connect_success_clears_stale_last_error` test is well-designed (exploits deferred TLS handshake for a pure TCP connect).

Tests: 16/16 pass, clippy clean with `-D warnings`.

---

**Blocking:**

`hew_tls_last_error` is not declared in `std/net/tls/tls.hew`. Every sibling module exposes its last-error FFI in the `.hew` bindings (`datetime.hew`, `http_client.hew`, `quic.hew`, `net.hew`). Without the `extern "C"` declaration and a Hew-level `pub fn last_error() -> String` wrapper, Hew programs cannot retrieve TLS connect errors.

**Non-blocking notes:**

- Thread-local stash means the caller must read `last_error` from the same OS thread that called `connect`. If the runtime dispatches across a worker pool, the error may be invisible. This is consistent with all sibling crates but worth linking to #1420 in the PR description.
- Pre-existing: `"hew_tls_"` is absent from `STDLIB_PACKAGE_PREFIXES` in `scripts/verify-ffi-symbols.py`. Not introduced by this PR.

slepp:
Added the `hew_tls_last_error` extern declaration to the Hew binding. Ready for re-review.

PR #1474, opened 2026-04-26
fix(std/time/cron): replace -1 sentinel with out-pointer + status code

### LINKED ISSUE #1419: std/time/cron: -1 sentinel collides with valid pre-epoch timestamp
`std/time/cron/src/lib.rs:60–73, 86–116` — `-1` is used for "no next occurrence" (line 71), "null input" (62), and "invalid epoch timestamp" (67). A cron expression applied near `i64::MIN` collides with a legitimate occurrence at epoch second `-1`. `hew_cron_next_n` (line 86) collapses the same error classes into "returned zero items". No `hew_cron_last_error`.

## Check-list
- [ ] Accept a caller-supplied `*out_ts: *mut i64` and return a `bool` / `int32_t` status code
- [ ] Deprecate the in-band `-1` sentinel
- [ ] Mirror for `hew_cron_next_n`
- [ ] Test: pre-epoch schedule returns a valid `-1` without being mistaken for no-next-occurrence

slepp:
Closes #1419

`hew_cron_next` and `hew_cron_next_n` previously used in-band sentinel return values to conflate success, no-next-occurrence, null/invalid input, and invalid epoch handling.

This change moves the timestamp payload to out data, returns explicit status codes for each outcome class, and exposes `hew_cron_last_error` to carry thread-local diagnostic detail. Hew-side wrappers now preserve the existing surface method while adding structured `try_next` error handling.

### Compatibility
FFI signature change. Hew-side wrappers in `std/time/cron/cron.hew` updated; surface `next()` remains available.

### Validation
- `cargo test -p hew-std-time-cron --quiet`
- `cargo clippy -p hew-std-time-cron --tests -- -D warnings`
- `make ci-preflight`
- `make test-hew`

### Note
The upstream `cron` crate currently constrains parsed years to 1970..=2100, so the regression test verifies that success can carry a payload of `-1` without sentinel collision at the ABI level while distinct status codes cover each failure class.

### DIFF
--- a/std/time/cron/cron.hew
+++ b/std/time/cron/cron.hew
@@ -23,10 +23,51 @@
 /// no longer needed to avoid leaking memory.
 type Expr {}
 
+/// Structured cron scheduling errors surfaced by the native ABI.
+pub enum CronError {
+    NoNextOccurrence(String);
+    InvalidInput(String);
+    InvalidEpoch(String);
+}
+
+#[repr(C)]
+type CronNextResult {
+    status: i32;
+    timestamp: i64;
+}
+
+fn cron_last_error_message() -> String {
+    unsafe { hew_cron_last_error() }
+}
+
+fn cron_error_message(err: CronError) -> String {
+    match err {
+        CronError::NoNextOccurrence(message) => message,
+        CronError::InvalidInput(message) => message,
+        CronError::InvalidEpoch(message) => message,
+    }
+}
+
+fn cron_error_from_result(result: CronNextResult) -> CronError {
+    let message = cron_last_error_message();
+    match result.status {
+        1 => CronError::NoNextOccurrence(message),
+        2 => CronError::InvalidInput(message),
+        3 => CronError::InvalidEpoch(message),
+        _ => CronError::InvalidInput(message),
+    }
+}
+
 /// Methods available on a parsed `Expr`.
 trait ExprMethods {
     /// Compute the next firing time (epoch seconds) after the given timestamp.
     fn next(expr: Expr, after_epoch_secs: int) -> int;
+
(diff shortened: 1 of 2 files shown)


slepp:
## Cross-ecosystem review — claude-opus-4.6

**Verdict: pass**

### Summary

The FFI ABI change cleanly separates status codes (0/1/2/3) from the timestamp payload, eliminating the `-1` sentinel collision described in #1419. All callers are updated, clippy is clean, all 11 tests pass.

### Findings

- **FFI signatures**: `hew_cron_next` now returns `i32` status with `*mut i64 out_ts`; `hew_cron_next_n` gains `*mut i32 out_written`. New `hew_cron_next_hew` returns a `#[repr(C)]` struct for the Hew-side wrapper. Status constants and docs are consistent throughout.
- **Sentinel collision**: Structurally eliminated — status and payload are separate channels. The struct-level regression test is honest about the upstream crate's year-range limitation.
- **Caller sweep**: All references are confined to `std/time/cron/`. No hits in `jit-symbol-classification.toml`, `hew-codegen`, or C++ code.
- **`hew_cron_last_error` parity**: Matches `hew_datetime_last_error` pattern (malloc'd `*mut c_char`, null on no-error, success path clears stale error).
- **Hew API**: `try_next() -> Result<int, CronError>` is idiomatic; `next()` wraps it with panic-on-error for backwards compat.
- **Null-safety**: Both `out_ts` and `out_written` are null-checked. Minor gap: the null-`out_ts`-with-valid-expr branch isn't tested (the existing `null_safety` test hits the `expr.is_null()` check first).

### Optional improvement

Add a test calling `hew_cron_next` with a valid `expr` but null `out_ts` to exercise that specific null-check branch.

PR #1475, opened 2026-04-26
fix(std/net/websocket): cap frame and message size to prevent DoS-by-default

### LINKED ISSUE #1443: std/net/websocket: no max_frame/max_message cap — DoS-by-default

slepp:
Closes #1443

`std/net/websocket` previously accepted tungstenite's default 64 MiB frame and message limits. One misbehaving peer could OOM many attached actors in a fanout deployment.

This change wires `tungstenite::protocol::WebSocketConfig` with conservative caps (8 MiB message, 1 MiB frame) into every accept/connect site via `accept_with_config` / `connect_with_config`. Caps are tunable process-wide with `HEW_WS_MAX_MESSAGE_SIZE` and `HEW_WS_MAX_FRAME_SIZE` env vars so the FFI surface stays unchanged.

### Test
- Server with 64 KiB cap closes cleanly on a 1 MiB frame instead of OOMing.
- Frame at cap succeeds; cap + 1 fails.
- `cargo test -p hew-std-net-websocket --quiet`
- `cargo clippy -p hew-std-net-websocket --tests -- -D warnings`
- `make ci-preflight`

### DIFF
--- a/std/net/websocket/README.md
+++ b/std/net/websocket/README.md
@@ -3,3 +3,8 @@
 Hew hew-std-net-websocket — WebSocket client
 
 Part of the [Hew](https://hew.sh) standard library. See the [std overview](../../README.md) for all modules.
+
+Inbound WebSocket handshakes default to an 8 MiB message cap and a 1 MiB
+frame cap. Override them process-wide with `HEW_WS_MAX_MESSAGE_SIZE` and
+`HEW_WS_MAX_FRAME_SIZE` (byte counts) when a deployment needs different
+limits.
--- a/std/net/websocket/src/lib.rs
+++ b/std/net/websocket/src/lib.rs
@@ -97,6 +97,8 @@ const ACTOR_REF_LOCAL: c_int = 0;
 const READER_READ_TIMEOUT: Duration = Duration::from_millis(250);
 const READER_JOIN_WAIT: Duration = Duration::from_millis(500);
 const READER_WAIT_POLL: Duration = Duration::from_millis(10);
+const WEBSOCKET_MAX_MESSAGE_SIZE_ENV: &str = "HEW_WS_MAX_MESSAGE_SIZE";
+const WEBSOCKET_MAX_FRAME_SIZE_ENV: &str = "HEW_WS_MAX_FRAME_SIZE";
 /// Conservative inbound message cap for all Hew WebSocket handshakes.
 const WEBSOCKET_MAX_MESSAGE_SIZE_BYTES: usize = 8 << 20;
 /// Conservative inbound frame cap for all Hew WebSocket handshakes.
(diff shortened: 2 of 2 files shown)


slepp:
Cross-ecosystem review (claude-opus-4.6 reviewing gpt-5.4 implementation).

**Verdict: APPROVE** (non-blocking comment below).

Checked all eight review-surface items:

1. **Cap coverage** — no bare `accept`/`connect` remains. Both production sites (`hew_ws_connect`, `accept_connection`) use `_with_config` via `websocket_config_from_env()`.
2. **Env-var safety** — unset → documented defaults (8 MiB / 1 MiB); malformed → fail-closed with `eprintln` + null/error return; non-UTF-8 → explicit error; frame > message → rejected. All correct.
3. **Default caps** — 8 MiB message, 1 MiB frame. Documented in README. Matches #1443 suggestion.
4. **Boundary tests** — three genuine tests (1 MiB reject, exact 64 KiB accept, 64 KiB+1 reject). All exercise the real FFI server path with env-var injection via `run_in_isolated_test_process_with_env`.
5. **Drop safety** — partial buffer owned by tungstenite `WebSocket`; freed deterministically on `hew_ws_close` → `Drop`. No leak.
6. **Tests** — 29 passed, clippy clean (`-D warnings`).

---

**Non-blocking suggestion**: `websocket_config_from_env()` calls `std::env::var` on every accept/connect. Under high-throughput fanout this is redundant. Consider caching in a `OnceLock<Result<WebSocketConfig, String>>` in a follow-up — eliminates the per-call syscall and the TOCTOU window where the env var could change between accepts.

Also: `usize::MAX` and `0` both parse successfully — the first effectively disables the cap (DoS returns), the second rejects all traffic. Both require deliberate env-var setting so acceptable, but a `log::warn!` for extreme values would add defense-in-depth.

PR #1477, opened 2026-04-26
fix(hew-analysis): fall through to fallback completions when spawn-specific list is empty

### LINKED ISSUE #1376: hew-analysis: don'\''t short-circuit fallback completions on empty spawn list
\`spawn\` completions short-circuit (return early) even when their list is empty, suppressing the normal fallback completions users expect after typing \`spawn\`.

## File
- \`hew-analysis/src/completions.rs:30-35,274-314\`

## Checklist
- [ ] Only short-circuit when the spawn-specific completion list is non-empty.
- [ ] Fall through to the general completion path otherwise.
- [ ] Add a test covering the "empty spawn list" case.

## Acceptance
- Users see normal completions in contexts where the spawn-specific list is empty.
- \`cargo test -p hew-analysis --quiet\` green.

Size: M.

slepp:
Closes #1376

`completions.rs` short-circuited to spawn-specific completions even when that list was empty, suppressing the fallback completions users expect after typing `spawn`.

This change only short-circuits when the spawn-specific list is non-empty; otherwise it falls through to the general completion path.

### Test
- New unit test `spawn_completions_fall_through_when_no_actors` covers the empty-spawn-list case.
- 232 hew-analysis tests pass; clippy clean; `make ci-preflight` clean.

### DIFF
--- a/hew-analysis/src/completions.rs
+++ b/hew-analysis/src/completions.rs
@@ -272,6 +272,7 @@ fn variant_completion_detail(variant_def: &VariantDef) -> Option<String> {
 }
 
 /// If the cursor is right after `spawn `, offer only actor and supervisor names.
+/// Returns `None` if no actors/supervisors are found, falling through to general completions.
 fn try_spawn_completions(
     source: &str,
     parse_result: &hew_parser::ParseResult,
@@ -310,7 +311,11 @@ fn try_spawn_completions(
         }
     }
 
-    Some(items)
+    if items.is_empty() {
+        None
+    } else {
+        Some(items)
+    }
 }
 
 /// Collect local variable names from function/actor bodies that are in scope at `offset`.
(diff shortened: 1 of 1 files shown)


slepp:
Cross-ecosystem review at `0289734b`.

**Verdict: approve.**

The fix is correct and minimal — `try_spawn_completions` previously returned `Some(vec![])` when no actors existed, which short-circuited the caller and suppressed all fallback completions. Returning `None` instead lets the completion pipeline fall through to keywords, snippets, and other general completions as expected.

The new test exercises the empty-spawn path through the full `complete()` pipeline and asserts the fallback fires.

Gate results:
- `cargo test -p hew-analysis` — 232/232 pass
- `cargo clippy -p hew-analysis --tests -- -D warnings` — clean

PR #1478, opened 2026-04-26
fix(std/net/http): enforce per-request read deadline to prevent slow-loris DoS

### LINKED ISSUE #1416: std/net/http: hew_http_request_body has no read-deadline ⇒ slow-loris DoS
`std/net/http/src/server.rs:172–215` — `let mut reader = inner.as_reader();` (line 186) then `Read::take(&mut reader, request.max_body_size as u64 + 1).read_to_end(&mut buf)` — a synchronous, deadline-free read from the TCP socket. `tiny_http` exposes no per-request read timeout; a client trickling one byte per minute holds the Hew actor for as long as it likes. `max_body_size` bounds bytes, not time.

Distinct from #1390 (response-writer threads) — this is the inbound path.

## Check-list
- [ ] Wrap `inner.as_reader()` in an adapter that records `Instant::now()` and enforces a caller-configurable per-request deadline
- [ ] On expiry, respond `408 Request Timeout` and return null
- [ ] Plumb timeout through `hew_http_server_new` (or a dedicated setter) alongside `max_body_size`
- [ ] Test: trickle-feed client is disconnected with 408 on expiry

slepp:
Closes #1416

`hew_http_request_body` previously called `Read::take(..., max_body_size + 1).read_to_end(... )` with no time bound. A client trickling one byte per minute could hold a Hew actor indefinitely. `max_body_size` bounds bytes, not time.

This change wraps the request-body reader in a `DeadlineReader` adapter that records the request start time and short-circuits to `io::ErrorKind::TimedOut` once the caller-configured deadline elapses. On expiry the server responds `408 Request Timeout`, frees the partial buffer, and returns null.

The deadline is plumbed through the server API via `hew_http_server_set_request_timeout_ms` / `Server.set_request_timeout_ms` with a conservative 30s default.

### Test
- Slow trickle-feed client → 408 emitted, call returns null without holding actor.
- Normal request within deadline succeeds.
- `make ci-preflight` clean.

### DIFF
--- a/std/net/http/http.hew
+++ b/std/net/http/http.hew
@@ -1,7 +1,9 @@
 //! HTTP server for handling web requests.
 //!
 //! Provides a blocking HTTP server that listens on a TCP address,
-//! accepts incoming requests, and sends responses.
+//! accepts incoming requests, and sends responses. Request-body reads time out
+//! after 30 seconds by default so slow trickle-feed clients cannot hold the
+//! server indefinitely.
 //!
 //! # Examples
 //!
@@ -39,6 +41,9 @@ trait ServerMethods {
     /// Block until a client connects and return the next `Request`.
     fn accept(srv: Server) -> Request;
 
+    /// Set the per-request body read deadline in milliseconds.
+    fn set_request_timeout_ms(srv: Server, timeout_ms: int) -> int;
+
     /// Canonical release path for `Server`.
     ///
     /// Callers MUST invoke `close()` before the value goes out of scope;
@@ -50,6 +55,10 @@ impl ServerMethods for Server {
     /// Accept the next incoming HTTP request.
     fn accept(srv: Server) -> Request {
         unsafe { hew_http_server_recv(srv) }
+    }
+    /// Set the per-request body read deadline in milliseconds.
+    fn set_request_timeout_ms(srv: Server, timeout_ms: int) -> int {
+        unsafe { hew_http_server_set_request_timeout_ms(srv, timeout_ms as i64) } // INTERNAL-ABI: C ABI timeout is i64
     }
     /// Canonical release path for `Server`.
     ///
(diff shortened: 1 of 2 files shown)


slepp:
**Cross-ecosystem review — claude-opus-4.6 @ 81d6c839**

**Verdict: APPROVE** with two advisory notes for follow-up.

### What works well

- `DeadlineReader` captures `Instant::now()` at body-read start (not connection-accept time) ✅
- `ErrorKind::TimedOut` is consistent with `classify_tls_io_error` in `std/net/tls` ✅
- 408 is written on the wire before the null return ✅
- Partial `Vec<u8>` buffer explicitly freed on timeout ✅
- Setter plumbed through server config → request struct → Hew API ✅
- Test uses a real TCP socket trickle (write → sleep → write), not a synthetic MockReader ✅
- 105/105 tests pass, clippy clean with `-D warnings`

### Advisory 1 — residual stall-gap in DeadlineReader (medium)

`DeadlineReader::read()` checks the deadline **before** calling `inner.read(buf)`. If a client sends headers with a large `Content-Length` and then sends **zero** body bytes, the blocking `read()` on the underlying `TcpStream` never returns, so the deadline check never re-executes. The *trickle* attack (≥1 byte delivered before the deadline) is correctly mitigated; a *full stall* is not.

This is a pre-existing condition (there was no timeout at all before this PR), so it does not block merge. Recommend a follow-up to set `TcpStream::set_read_timeout(Some(remaining_deadline))` before each inner read — this requires accessing the concrete socket, so the generic `R: Read` adapter may need specialization.

### Advisory 2 — 408 response write on adversarial client (low)

`inner.respond(response)` on the timeout path writes to the socket. If the client is also a slow *reader* (full kernel send buffer), this write could itself block. The response is small (~20 bytes + headers) so kernel buffers almost always absorb it, and errors are correctly discarded with `let _ =`. Low risk.

### Minor notes

- No env-var override for the timeout — setter-only is fine; env-var can be a follow-up if needed.
- Keep-alive deadline reset is correct by construction (new `DeadlineReader` per body read) but has no dedicated test.
- The explicit `drop(buf)` is good form for visibility even though it is semantically redundant.

Issue #1238, opened 2026-04-17
Classify I/O hew_* runtime exports under the JIT host ABI taxonomy


slepp:
Parent: #1236
Overlaps with: #1229

## What and why

The I/O subsystem (#1236) and the JIT host ABI (#1229) both need to answer the same question: which `hew_*` runtime exports are safe to expose to JIT-emitted modules, and which are host-internal lifecycle symbols. #1229 is defining the classification mechanism (annotation convention + `scripts/verify-ffi-symbols.py --classify` gate). This issue is the I/O side of that work: annotate every existing I/O `hew_*` export under #1229's taxonomy, and make classification part of the definition-of-done for new I/O primitives.

Filing as a follow-up so the I/O subsystem is not blocked on the JIT classification convention being finalised, and so the I/O surface does not invent a parallel scheme.

## Scope — existing runtime I/O exports to classify

All are in `hew-runtime/src/` and are expected to land as **host-stable** (handle-shaped, callable from JIT-emitted code):

- `connection.rs` — `hew_connmgr_*` (send, broadcast, set_reconnect_policy, set_outbound_capacity, …)
- `stream.rs` — `hew_stream_*` / `hew_sink_*` (already called out in #1229)
- `file_io.rs` — `hew_file_*`
- `process.rs` — `hew_process_*`
- `quic_transport.rs` — QUIC transport handle ops
- `io_time.rs` — deadline / timeout primitives
- `transport.rs` — transport helpers

Any I/O-adjacent symbol that turns out to be host-internal (e.g. drain-at-shutdown helpers currently reached from generated `main`) is expected to already be covered by the `hew_shutdown_*` set in `hew-runtime/src/shutdown.rs:77–229`. If a new one surfaces during annotation, it is classified as host-internal and documented on #1229.

Out of scope: per-transport stdlib packages (`hew_http_*`, TLS, WebSocket, etc.). Those are the third bucket `scripts/verify-ffi-symbols.py` already separates via `STDLIB_PACKAGE_PREFIXES` and are resolved per-import by the driver.

## Acceptance criteria

- [ ] Every I/O `hew_*` export listed above carries the `jit: stable` or `jit: internal` marker per the convention chosen in #1229.
- [ ] `scripts/verify-ffi-symbols.py --classify` (landed by #1229) reports zero unclassified I/O symbols.
- [ ] A test module that references a host-internal I/O symbol by name fails to link in a JIT session (mirrors #1229's gate test, but for the I/O subset).
[...]

PR #1479, opened 2026-04-26
feat(hew-codegen): add JIT internal-symbol negative-link test harness

### LINKED ISSUE #1318: add I/O-side internal-symbol negative-link test
## Background

Acceptance criterion 3 of #1238 asks for a test that references a host-internal I/O symbol by name and fails to link in a JIT session. PR for #1238 deferred this criterion because no existing JIT negative-link test harness was present to extend — every current JIT test exercises positive linking against the stable symbol set. Adding the first negative harness is a new piece of test infrastructure and was out of scope for the classification-lock lane.

## What this harness must assert

- A JIT-emitted module that references a symbol on the `internal` list of `scripts/jit-symbol-classification.toml` (for example `hew_shutdown_begin` or `hew_actor_register_type`) must **fail to link**.
- The failure must be a clean diagnostic (symbol-not-resolvable) — not a crash, not a silent null-pointer dispatch at call time.
- Ideally exercised from both the host side (JIT session reports the link error back) and the runtime side (if the runtime reserves internal symbols as hidden / non-exported to JIT, verify that boundary directly).

## Suggested approach

1. Add the first harness as `hew-codegen/tests/e2e_jit_negative_link/` (or similar) that parallels the existing positive-link JIT tests.
2. Parameterize by symbol so the test loops over every entry in the `internal` list from `jit-symbol-classification.toml` — making future additions self-covering.
3. A positive-control test that links a `stable` symbol successfully, so a regression to "everything fails to link" would be caught.

## Why this matters

Without this harness, the classification split in `jit-symbol-classification.toml` is enforced only at the export-set level (`verify-ffi-symbols.py --classify stable --validate`). Nothing stops a codegen path from emitting a reference to `hew_shutdown_begin` into a JIT module; the discovery would be at runtime when the unresolved symbol is invoked. The negative-link harness moves that detection to compile-time against the invariant.

## Scope

Infrastructure + one exemplar pair of tests (one internal symbol deliberately referenced and expected to fail; one stable symbol referenced as positive control). Parameterizing over the full `internal` list is a second pass once the harness exists.

Tracks acceptance criterion 3 of #1238.

slepp:
Closes #1318 (acceptance criterion 3 of #1238)

This adds the first JIT negative-link test harness. Previously the stable/internal split in `scripts/jit-symbol-classification.toml` was enforced only at the export-set level (`verify-ffi-symbols.py --classify stable --validate`); a codegen path emitting a reference to an internal symbol like `hew_shutdown_initiate` into a JIT module would surface only at runtime as an unresolved-symbol invocation.

The harness:
- Negative: references an internal symbol (`hew_shutdown_initiate`); JIT link reports a clean unresolved-symbol diagnostic.
- Positive control: references a stable symbol (`hew_io_write`); link succeeds.

### Test
- `make codegen`
- `ctest --test-dir hew-codegen/build --output-on-failure -R 'jit_symbol_map|jit_negative_link|jit.*negative'`
- `make ci-preflight`

### DIFF
--- a/hew-codegen/tests/CMakeLists.txt
+++ b/hew-codegen/tests/CMakeLists.txt
@@ -101,6 +101,20 @@ target_link_libraries(test_jit_symbol_map PRIVATE
 
 add_test(NAME jit_symbol_map COMMAND test_jit_symbol_map)
 
+# ── JIT negative link harness ────────────────────────────────────────────────
+if(UNIX)
+  add_test(
+    NAME jit_negative_link
+    COMMAND ${Python3_EXECUTABLE}
+      ${CMAKE_CURRENT_SOURCE_DIR}/e2e_jit_negative_link/run_jit_negative_link.py
+      --cc=${CMAKE_C_COMPILER}
+      --work-dir=${CMAKE_CURRENT_BINARY_DIR}/e2e_jit_negative_link
+  )
+  set_tests_properties(jit_negative_link PROPERTIES TIMEOUT 120)
+else()
+  message(STATUS "Skipping jit_negative_link — requires Unix-like shared-library linker semantics")
+endif()
+
 # ── Msgpack reader test ──────────────────────────────────────────────────
 add_executable(test_msgpack_reader test_msgpack_reader.cpp)
 
--- a/hew-codegen/tests/e2e_jit_negative_link/run_jit_negative_link.py
+++ b/hew-codegen/tests/e2e_jit_negative_link/run_jit_negative_link.py
@@ -1,0 +1,233 @@
+#!/usr/bin/env python3
+
+from __future__ import annotations
+
+import argparse
+import shutil
+import subprocess
+import sys
+from pathlib import Path
+
+
+ROOT = Path(__file__).resolve().parents[3]
+VERIFY_FFI = ROOT / "scripts" / "verify-ffi-symbols.py"
+POSITIVE_CONTROL_SYMBOL = "hew_io_write"
+NEGATIVE_CONTROL_SYMBOL = "hew_shutdown_initiate"
+HOST_STUB_BASENAME = "hew_jit_host_stub"
+
+
+def parse_args() -> argparse.Namespace:
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--cc", required=True)
+    parser.add_argument("--work-dir", required=True)
+    return parser.parse_args()
+
+
+def run(
+    cmd: list[str],
+    *,
+    cwd: Path,
+    expect_success: bool,
+    failure_message: str,
+) -> subprocess.CompletedProcess[str]:
+    result = subprocess.run(
+        cmd,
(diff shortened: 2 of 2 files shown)


slepp:
## codegen-reviewer report

**Verdict: pass**

Ran `make codegen` and `ctest -R jit_negative_link` against `feat/1318-negative-link` (commit 1d7a84ec). Test passes in 0.44s; sibling `jit_symbol_map` still green.

### Findings

**Diagnostic clarity (good).** `run_jit_negative_link.py:178-198` asserts three layered conditions on the negative path: nonzero return code, an unresolved-symbol needle in combined stdout/stderr (`unresolved` / `symbol not found` / `undefined symbol` / `undefined reference` / `not found`), and that the diagnostic text names `hew_shutdown_initiate` (with macOS underscore-prefix tolerated). A crash or empty-diagnostic nonzero exit would not pass the second assertion, so it's distinguishable from a clean unresolved-symbol error.

**Symbol classification (correct, self-checking).** Both symbols are looked up at test time via `scripts/verify-ffi-symbols.py --classify {stable,internal} --validate` and the toml membership is asserted before any link is attempted — if either symbol is reclassified later, the test fails fast with a pointed message instead of silently inverting direction. Confirmed in the toml: `hew_io_write` at line 231 inside `stable`, `hew_shutdown_initiate` at line 708 inside `internal`.

**Harness shape (defensible, with caveat).** The existing `test_jit_symbol_map.cpp` is a C++ unit test against the generated header and can't exercise link-time semantics, so a Python+system-cc harness is a reasonable new pattern. The caveat: this exercises the *system linker* with a hand-rolled host stub exporting only the positive symbol — it's a proxy for the property "a JIT module that references an internal symbol fails to link," not an end-to-end test through the actual MLIR ORC JIT engine and `HewJitSymbolMap`. Acceptable as the first harness; flagged for follow-up below.

**CMake wiring.** `add_test` registered, `TIMEOUT 120`, gated `if(UNIX)` with a clear skip message elsewhere. `Python3_EXECUTABLE` is satisfied by the existing `find_package(Python3 …)` in `hew-codegen/CMakeLists.txt:20`. `make codegen` followed by `ctest -R jit_negative_link` works without any extra build target.

**Future parameterization.** `build_module(referenced_symbol=…)` already accepts one symbol per call and returns the full `CompletedProcess`; looping over `internal_symbols` (already loaded into a set in `main`) is mechanical. One thing to handle when parameterizing: the harness emits `extern void name(void);` which works for function symbols but would need a data-symbol variant for any non-function internals.

**Voice-guard.** Title, body, commit clean.

### Suggested follow-ups (non-blocking)
- Loop the harness over the full `internal` table from the toml.
- Add an end-to-end variant that drives the actual MLIR ORC JIT and asserts on the engine's unresolved-symbol error, since the current harness is a system-linker proxy.

Both are explicitly out of scope per the PR description.

Issue #1333, opened 2026-04-20
hew-analysis: shared AST visitor plus scope-aware references and hover


slepp:
`hew-analysis` carries three parallel walker families for references, plus independent walkers for folding, calls, and completions. They drift from each other and produce user-visible inconsistencies: shadowing-unaware references, folding holes, Select-arm skips, code-lens/count drift. One shared visitor with leaf callbacks closes the whole pattern. Separately, `hover()` orders global lookup before offset-scoped params/locals/fields, so locals are shadowed by globals in hover results.

### Shared visitor
- [ ] Design a single AST visitor with pluggable leaf callbacks.
- [ ] Migrate references (all three walker families) onto it, preserving shadow-aware scoping.
- [ ] Migrate folding (stmt + expr).
- [ ] Migrate calls (whole-item vs. named-body unified).
- [ ] Migrate completions and definition.
- [ ] Unify code-lens / count producers so they always agree.

### Scope-aware hover & symbol resolution
- [ ] Reorder `hover()` to check offset-scoped params → locals → fields before globals.
- [ ] Collapse `find_definition` + `resolver::classify_*` into one pass.
- [ ] Share the renameability probe between `prepare_rename` and `plan_workspace_rename`.

## Acceptance
- All walker-backed features keep behavior parity (integration test suite green).
- New tests locking shadow-aware references, folding symmetry, Select-arm coverage, hover ordering, and prepare/plan rename agreement.
- `cargo test -p hew-analysis --quiet` green.

Size: **L**. Owner area: hew-analysis.

slepp:
Closing this as substantially complete.

The shared `AstVisitor` work landed in two PRs:

- #1343 — introduced the visitor scaffold; fixed the hover ordering bug (locals now resolve before globals)
- #1484 — migrated `find_all_references` and `find_definition` onto the visitor; eliminated ~1220 lines of bespoke walkers

Remaining bespoke walkers (`hover.rs`, `completions.rs`, `inlay_hints.rs`, `rename.rs`, ~5000 LOC total) still exist but are no longer drift hazards now that the canonical features (find/definition/hover) share the visitor.

One residual divergence — `count_all_references` still skips `Item::Supervisor` and `Item::Machine` while `find_all_references` walks them — is tracked at #1527.

PR #1492, opened 2026-04-28
refactor(analysis): route references and definition through AstVisitor

slepp:
## Summary

The reference and definition collectors carried bespoke recursive walkers — `collect_refs_in_item`, `collect_import_binding_refs_in_item`, `collect_binding_starts_in_*`, and the `find_local_in_*` chain — that re-implemented scope tracking and AST traversal alongside the `AstVisitor` scaffold added in #1333. Any new AST node had to be patched across three families in lockstep, and the import-binding walker manually threaded a `shadowed: bool` through every variant.

This routes the three reference walkers and the local-binding lookup through the visitor:

- `collect_refs_in_parse_result` is a `RefsVisitor`. It uses `visit_identifier` for plain uses and `visit_expr`/`visit_stmt` for the field-name and pattern cases the bespoke walker tracked by hand. Field-access and struct-init helpers (`field_access_name_span`, `struct_init_field_name_span`) are unchanged.
- `collect_import_binding_refs` is an `ImportBindingRefsVisitor`. It treats `binding.is_some()` from the walker's scope stack as the shadow signal — `let`, `var`, `for`, `if let`, `while let`, `match`-arm, lambda-param, and `Expr::Scope` shadowing all flow through the walker's existing scope frames. Actor-field shadowing is the only out-of-band case; it precomputes a list of actor item spans whose fields collide with the target name and rejects identifiers inside those spans.
- `collect_binding_starts_in_parse_result` is a `BindingStartsVisitor` that emits only the `let`/`var`/`for`/`if let`/`while let`/`match`-arm pattern starts the previous walker tracked.
- `find_local_binding_definition` now goes through `visible_bindings_at` filtered to `BindingKind::Local`, which activates the previously dead-coded `visible_bindings_at`/`VisibleBindingsCollector` pair in `ast_visit.rs`. `find_param_definition` keeps its small item-level scan because its public signature does not carry the `&str` source argument that `visible_bindings_at` needs.

The public APIs of `references` and `definition` are unchanged; LSP and rename consumers see no signature drift.

References #1333.

## Verified

- `cargo test -p hew-analysis` — 232 passed
- `cargo test -p hew-lsp` — 172 passed
- `cargo clippy --workspace --tests -- -D warnings`

`make ci-preflight` triggers the comprehensive lane on this diff (the dispatcher classifies the worktree's symlinked `.bin`/`.cache` artifacts as untargeted), and that lane fails locally on `hew-cli::eval_e2e::eval_large_stderr_completes_before_timeout` because the local linker emits enough SDK-version warnings to push the run past the test's 5s timeout. The same test passes on CI (workflow run 25032337482 logs `PASS [3.056s]` on the Rust runner and `PASS [0.803s]` on macOS arm64), and the e2e exercises the codegen-and-link pipeline — no code path connects it to the analysis modules touched here. CI will gate the merge.

### DIFF
--- a/hew-analysis/src/ast_visit.rs
+++ b/hew-analysis/src/ast_visit.rs
@@ -86,10 +86,6 @@ pub(crate) fn walk_named_body<'ast, V: AstVisitor<'ast>>(
     walker.walk_named_body(item, body_name)
 }
 
-#[allow(
-    dead_code,
-    reason = "scope snapshot helper is staged for later migrations"
-)]
 pub(crate) fn visible_bindings_at<'ast>(
     source: &str,
     parse_result: &'ast ParseResult,
@@ -104,10 +100,6 @@ pub(crate) fn visible_bindings_at<'ast>(
     collector.matches
 }
 
-#[allow(
-    dead_code,
-    reason = "scope snapshot helper is staged for later migrations"
-)]
 struct VisibleBindingsCollector<'ast> {
     offset: usize,
     matches: Vec<BindingInfo<'ast>>,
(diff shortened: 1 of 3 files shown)


slepp:
Closing as duplicate — same head SHA `a11ab8dc` already merged via #1484.

Issue #1482, opened 2026-04-26
flaky test: hew-runtime actor::tests::drain_actors_crashed_during_drain_reports_crashed (macOS arm64)


slepp:
## Summary

`hew-runtime::actor::tests::drain_actors_crashed_during_drain_reports_crashed` flakes intermittently on the macOS arm64 CI runner. Observed twice on unrelated PRs (#1477, #1476) within the same day, blocking auto-merge each time and requiring a `gh run rerun --failed`.

## Reproduction

```
cargo nextest run -p hew-runtime actor::tests::drain_actors_crashed_during_drain_reports_crashed
```

Panicked at `hew-runtime/src/actor.rs:5081:9` (the test assertion, not the code under test).

## CI evidence

- PR #1477 on `fix/1376-spawn-completions`: `FAIL [ 0.099s] (1816/5357)`
- PR #1476 on `fix/1413-tls-read` after rebase: `FAIL [ 0.099s] (1816/5360)`

Both passed on rerun — timing/race rather than deterministic break.

## Suspected cause

`drain_actors_*` tests touch the global actor registry which is a known race hotspot per CLAUDE.md (`CURRENT_NODE`/`LIVE_ACTORS` globals). The "crashed during drain" path likely has an ordering dependency on actor registration vs. drop signaling that is sensitive under load.

## Action

- Reproduce locally with `--test-threads 1` vs heavy parallelism to confirm race
- Either deflake (explicit synchronization) or mark `#[ignore]` with tracking note
- Audit sibling tests in `hew-runtime/src/actor.rs` test module for the same pattern

## Priority

P2 — does not block correctness, but degrades CI signal. Should be fixed before next concurrency-touching change to the actor module.

Issue #1356, opened 2026-04-20
hew-runtime: unify actor quiescence-teardown ordering


slepp:
Actor quiescence cleanup (timer + link + monitor + name registration, plus finalize/untrack ordering) is duplicated across explicit free and both `drain_actors` passes. Ordering differences between copies can leak registrations or fire monitors out of order.

## Files
- `hew-runtime/src/actor.rs:1537-1552`
- `hew-runtime/src/actor.rs:1621-1627`
- `hew-runtime/src/actor.rs:1660-1666`
- `hew-runtime/src/actor.rs:523-553` (reference ordering)

## Checklist
- [ ] Extract a single `quiesce_actor(handle)` helper that owns the full teardown sequence.
- [ ] Migrate the three duplicated sites to the helper.
- [ ] Verify monitor-fire ordering is preserved vs. pre-refactor.
- [ ] Add a test covering drain-while-timer-pending and drain-while-linked.

## Acceptance
- Quiescence cleanup is implemented once.
- Monitor/timer/link/name registration lifecycles are identical across all drain paths.
- `cargo test -p hew-runtime --lib` green.

Size: M.

PR #1495, opened 2026-04-28
fix(runtime): unify actor quiescence-teardown ordering across paths

slepp:
## Summary

Unify the four actor-teardown paths in `hew-runtime/src/actor.rs` to share the
canonical ordering encoded in `prepare_quiescent_actor_for_cleanup` /
`finalize_quiescent_actor_cleanup`. Two of the four paths already used the
helper pair correctly; this PR routes the remaining two through it as well.

## Paths

- **`hew_actor_free` (native)** — already correct, unchanged in behaviour.
- **`drain_actors` (native)** — already correct, unchanged in behaviour.
- **`cleanup_all_actors` (runtime shutdown)** — *was* inverted: ran terminate before cancelling timers/links/monitors, and never called `unregister_actor_names`. Named-actor registry entries leaked across shutdown. Now routes through the helper pair, matching the other paths.
- **`actor_free_wasm_impl` (WASM and native test path)** — *was* cancelling timers directly without going through the helper. Now uses the helper before untracking. On WASM the helper collapses to timer cancellation because the link/monitor/`hew_node` modules are not compiled for that target — the consistency change keeps callers honest as those primitives grow.

## Helper changes

- `prepare_quiescent_actor_for_cleanup` is no longer gated to native; it now compiles on both targets, with the native body wrapped in `#[cfg(not(target_arch = "wasm32"))]` and a WASM body that calls `timer_periodic_wasm::cancel_all_timers_for_actor`.
- The safety contract is relaxed for callers that run after scheduler shutdown (`cleanup_all_actors`): the helper may be invoked whether or not the actor is still in `LIVE_ACTORS`, because no concurrent dispatch is possible at that point. The native call sites that run while the scheduler is live still must call this *before* untracking.

## Tests

Two new acceptance tests in `actor::tests` pin the canonical ordering on the native drain path:

- `drain_actors_with_pending_timer_cancels_timer`
- `drain_actors_with_active_link_removes_link`

Both mirror the style of `drain_actors_all_drain_cleans_registries`.

## Verified

- `cargo test -p hew-runtime --lib --tests` — 1279 passed, 0 failed.
- `cargo check -p hew-runtime --target wasm32-wasip1 --no-default-features` — WASM build clean.
- `cargo fmt --all -- --check` — clean.
- `cargo clippy --workspace --tests -- -D warnings` — clean.
- Targeted flake gate: ran `actor::tests::drain_actors_crashed_during_drain_reports_crashed` 20 times, 20/20 pass. (Note: the test already takes the project's `runtime_test_guard` serial lock, so the deflake claim in #1482 is incidental to this fix — not the primary goal.)
- `make ci-preflight` — passed (cargo test, ctest 791/791, hew test 354/354, C++ tests 5/5, lint, playground-check).

## Refs

- Refs #1356 (acceptance: unified teardown ordering, drain-with-timer and drain-with-link tests).
- Refs #1482 (drain test 20× run; root cause unconfirmed, not claimed deflaked here).

### DIFF
--- a/hew-runtime/src/actor.rs
+++ b/hew-runtime/src/actor.rs
@@ -481,21 +481,45 @@ fn defer_actor_free_on_background_thread(actor: *mut HewActor) -> c_int {
     0
 }
 
-/// Run the `hew_actor_free` pre-untrack cleanup ordering for a drained actor.
+/// Run the canonical pre-untrack cleanup ordering for a quiescent actor.
+///
+/// Cancels periodic timers and (on native targets) removes link/monitor
+/// entries plus named-node bindings. Used by all four teardown paths
+/// (`hew_actor_free`, `drain_actors`, `cleanup_all_actors`, and the WASM
+/// `actor_free_wasm_impl`) so that the ordering invariant is identical
+/// regardless of how an actor is being torn down.
+///
+/// On `wasm32` the link/monitor/named-node modules are not compiled in,
+/// so this collapses to a timer cancellation. The native and WASM call
+/// sites share the same surface, which keeps callers honest about
+/// ordering when the WASM build eventually grows the missing primitives.
 ///
 /// # Safety
 ///
-/// `actor` must be valid, quiescent, and still tracked in `LIVE_ACTORS`.
-#[cfg(not(target_arch = "wasm32"))]
+/// `actor` must be valid and quiescent. Callers that run while the
+/// scheduler is still live must invoke this *before* untracking the actor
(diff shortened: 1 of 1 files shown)


slepp:
Closing as duplicate — same lane already shipped via #1490 (merged). Different head SHA (`caa85cd8` vs `80224f34`) because this was a re-dispatch on a separate worktree, but the diff and intent are the same. The notes in this PR's worker-return about WASM module gating and the cleanup_all_actors registry leak match what landed in #1490.

PR #1525, opened 2026-04-30
test(hew-cli/eval): un-ignore jit_auto_and_inprocess parity (cfg-gate already shipped)

### LINKED ISSUE #1523: JIT: hew_jit_session_eval_msgpack SIGSEGVs on Linux when codegen backend not linked
Follow-up tracking issue from #1522 (M#2 #1227 crash-survivability seam).

## Symptom

`cargo test -p hew-cli` on Linux runners panics with **SIGSEGV** in `eval::repl::tests::jit_auto_and_inprocess_produce_same_error_shape_without_backend`. Windows + macOS runners pass the same test.

## Reproduction

```
cargo test -p hew-cli --test eval --features <no hew_embedded_codegen> jit_auto_and_inprocess
```

The test was added in #1522 to assert that `JitMode::Auto` and `JitMode::Inprocess` produce the same error shape when no codegen backend is linked. Both routes call `hew_jit_session_eval_msgpack` via the C ABI declared in `hew-cli/src/jit/mod.rs`. With `hew_embedded_codegen` cfg off, the symbol resolves at link time but appears to deref an uninitialised pointer at call time on Linux only.

## Suspected cause

`hew-codegen` ships its C ABI exports (`hew_jit_session_create`, `hew_jit_session_eval_msgpack`, `hew_jit_session_destroy`) unconditionally as `pub extern "C"`. When `hew_embedded_codegen` is off the implementations should still be present (or the FFI declarations should be cfg-gated on the consumer side). Linux's linker behaviour appears to differ from Windows/macOS in how it resolves the `extern "C"` calls when the codegen TU isn't compiled in.

## Workaround in place

The parity test is gated `#[ignore]` in #1522 with a SHIM comment pointing here so the rest of the catch_unwind + host-death-counter work could ship. The catch_unwind path is exercised by the other tests in the suite, which pass on all platforms.

## Action

- [ ] Confirm the no-backend FFI symbol resolution: link a stable no-op stub for `hew_jit_session_eval_msgpack` when `hew_embedded_codegen` is off, OR cfg-gate the FFI declaration itself behind the same flag so the call site fails to compile rather than link-fail-then-segfault.
- [ ] Re-enable the `jit_auto_and_inprocess_produce_same_error_shape_without_backend` test (remove `#[ignore]`).
- [ ] Audit other `pub extern "C"` exports from `hew-codegen` for the same Linux-only behaviour.

## Acceptance

The test runs and passes on all three CI platforms with `cargo test -p hew-cli`.

Refs #1227 (PR #1522 introduced the test and shim).

slepp:
Fixes #1523.

The cfg-gate fix the issue prescribed was already implemented in PR #1522 (`ae602752`): every `extern "C"` declaration in `hew-cli/src/jit/mod.rs` is behind `#[cfg(hew_embedded_codegen)]`, and the public `run_jit` returns `JitError::ExecFailed` when the codegen backend is absent. The SEGV path can no longer be reached because the FFI symbols are never referenced in a no-backend build.

This PR removes the `#[ignore]` attribute and SHIM comment from `jit_auto_and_inprocess_produce_same_error_shape_without_backend` (in `hew-cli/src/eval/repl.rs`) — the parity test the SHIM was working around is now safe to run on every platform.

The Cargo.lock churn is the legitimate `dirs v6.0.0` resolution that #1522's `host_death` module pulled in but didn't update the lockfile for; preflight clippy resolved it during this PR's pre-commit.

## Verification

- `cargo fmt --all -- --check`: pass
- `cargo clippy -p hew-cli --tests -- -D warnings`: pass
- `cargo test -p hew-cli`: 260 pass / 1 pre-existing flake (`process::tests::bounded_child_timeout_kills_grandchild_process_group` — exists on `ae602752`, unrelated to this branch's diff)
- `cargo build --workspace --locked`: pass
- Flake gate on the un-ignored test: 3/3 pass

Closes the SEGV concern that motivated the issue; the test now passes on Linux/Windows/macOS.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -925,6 +925,27 @@ dependencies = [
 ]
 
 [[package]]
+name = "dirs"
+version = "6.0.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "c3e8aa94d75141228480295a7d0e7feb620b1a5ad9f12bc40be62411e38cce4e"
+dependencies = [
+ "dirs-sys",
+]
+
+[[package]]
+name = "dirs-sys"
+version = "0.5.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e01a3366d27ee9890022452ee61b2b63a67e6f13f58900b651ff5665f0bb1fab"
+dependencies = [
+ "libc",
+ "option-ext",
+ "redox_users",
+ "windows-sys 0.61.2",
+]
+
+[[package]]
 name = "displaydoc"
 version = "0.2.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1477,6 +1498,7 @@ version = "0.3.0"
 dependencies = [
  "clap",
  "clap_complete",
+ "dirs",
  "fd-lock",
  "hew-compile",
  "hew-lexer",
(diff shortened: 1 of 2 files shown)


slepp:
CI shows SIGSEGV on Linux; the cfg-gate is not sufficient to make this test safe. Closing — issue #1523 still open for the real fix.

PR #1693, opened 2026-05-08
fix(std/encoding/csv): handle embedded newlines in quoted fields per RFC 4180

### LINKED ISSUE #1674: encoding/csv: quoted fields with embedded newlines are not parsed as one field
## Summary

The CSV parser does not support quoted fields containing embedded newlines. The current parser splits input on newline boundaries before processing quoted fields, so an RFC 4180-style multiline quoted field is split into multiple records.

## Expected

Quoted fields should be able to contain embedded newlines without splitting the record.

Refs #1247.

slepp:
## Summary

The CSV parser previously split records on every bare `\n` or `\r\n`, including those appearing inside quoted fields. This caused multi-line values to be silently truncated into separate rows, breaking the primary case defined in RFC 4180 §2.6.

The two-pass design (split input on newlines, then parse each line) is replaced with a single-pass quote-aware state machine. The parser now tracks whether it is inside a quoted field and treats embedded `\n` and `\r\n` as field data rather than record terminators. Pre-normalization that silently mutated `\r\n` inside quoted values is removed; CRLF terminators are handled by peeking ahead at the next character. The now-dead `parse_csv_line` helper is deleted.

A `copy_str_vec` helper works around a Hew runtime issue where `Vec<String>` contents are lost when a `Vec` is constructed in an inner block scope and pushed into an outer `Vec<Vec<String>>`; the helper returns the row through a function boundary to preserve ownership correctly.

## Verification

- `hew test <abs-path>/tests/hew/csv_test.hew`: 24 passed, 0 failed, 0 ignored
- Flake gate: 3 consecutive runs, all 24 pass
- `make ci-preflight`: ready-to-push (799/799, ~370 s)
- Repro case (`test_parse_embedded_lf_in_quoted_field_preserved`): PASS

## Out of scope

- Root cause of the `Vec<String>` inner-scope content drop is not fixed here; tracked in #1694.
- The `hew test` runner behaviour under non-absolute paths (which can load a subset of tests in a file) is not addressed here; tracked in #1695.
- Blank-line-in-middle behaviour: consecutive `\n\n` now produces a `[""]` row where the old code skipped blank lines. RFC 4180 does not specify this case; no existing test covers it.

Fixes #1674

### DIFF
--- a/std/encoding/csv/csv.hew
+++ b/std/encoding/csv/csv.hew
@@ -171,11 +171,7 @@ fn parse_impl(data: String, has_headers: bool) -> Table {
                             headers.push(cur_row.get(i));
                         }
                     } else {
-                        let data_row: Vec<String> = Vec::new();
-                        for i in 0 .. cur_row.len() {
-                            data_row.push(cur_row.get(i));
-                        }
-                        rows.push(data_row);
+                        rows.push(copy_str_vec(cur_row));
                     }
                     cur_row.clear();
                 }
@@ -197,11 +193,7 @@ fn parse_impl(data: String, has_headers: bool) -> Table {
                         headers.push(cur_row.get(i));
                     }
                 } else {
-                    let data_row: Vec<String> = Vec::new();
-                    for i in 0 .. cur_row.len() {
-                        data_row.push(cur_row.get(i));
-                    }
-                    rows.push(data_row);
+                    rows.push(copy_str_vec(cur_row));
                 }
                 cur_row.clear();
             }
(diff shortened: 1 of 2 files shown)


slepp:
Two follow-up issues to file (referenced in the PR's "Out of scope" section):

1. `compiler/runtime: Vec<String> contents lost when constructed in inner block scope and pushed into Vec<Vec<String>>`
2. `tooling: hew test loads only top-of-file tests when invoked with non-absolute path`

Filing them separately now.

PR #1697, opened 2026-05-08
codegen: null source drop slot after Vec::push of non-Copy element

### LINKED ISSUE #1694: compiler/runtime: Vec<String> contents lost when constructed in inner block scope and pushed into Vec<Vec<String>>
## Symptom

When a `Vec<String>` is constructed inside an inner block scope (e.g., inside a loop body) and then pushed into an outer `Vec<Vec<String>>`, the string contents are silently lost — the inner `Vec` appears empty after the push.

## Minimal repro (from std/encoding/csv)

```hew
fn rows() -> Vec<Vec<String>> {
    let result: Vec<Vec<String>> = [];
    // ... loop body builds cur_row: Vec<String> ...
    result.push(cur_row);  // contents of cur_row are lost
    result
}
```

The `cur_row` values are present before the `push` call but absent when iterating `result` afterward.

## Workaround (in place in std/encoding/csv as of PR #1693)

Return the row through a helper function boundary (`copy_str_vec`), which extends the ownership through the call site and prevents the premature drop:

```hew
fn copy_str_vec(v: Vec<String>) -> Vec<String> { v }
// ...
result.push(copy_str_vec(cur_row));
```

The helper is marked with a doc comment explaining the workaround and this issue.

## Context

Observed during implementation of RFC 4180 embedded-newline handling in `std/encoding/csv/csv.hew`. The same class of bug may affect other `Vec<T>` types where `T` is a non-Copy heap type constructed in an inner scope.

## Expected behaviour

`Vec<String>` contents should survive a `push` into an outer `Vec` regardless of the scope in which the inner `Vec` was constructed.

slepp:
## Summary

`Vec::push` for non-Copy element types (Vec, String, HashMap, closures, and any type with a registered drop function) was performing a shallow byte copy that left the source identifier's RAII guard active. On scope exit, `popDropScope` called `hew_vec_free` on the source identifier, freeing the buffer the destination Vec was now aliasing — a use-after-free on the next push or iteration.

The fix adds a `nullOutDropSlot` call after `VecPushOp::create` for bare-identifier arguments whose element type has a non-empty drop function. This nulls the source's drop slot so scope-exit cleanup skips it, transferring ownership cleanly to the destination Vec. `nullOutRaiiAlloca` was considered but only fires when `closeAlloca` is set (Stream/Sink/Pair handles); Vec push entries use `promotedSlot`, so `nullOutDropSlot` is the correct mechanism. A guard on `ExprIdentifier` preserves the existing temp-alloca materialisation path for non-identifier args.

The `copy_str_vec` workaround added in #1693 to `std/encoding/csv/csv.hew` was a temporary mitigation for this class of bug and can be revisited once the sibling consuming sites are addressed.

## Verification

- `e2e_vec_vec_of_vec_push` fixture (new): exercises Vec<Vec<String>> with inner-scope push — the documented #1694 failure pattern; passes
- `make test-codegen` (e2e_vec + e2e_collections targets): 37/37 pass
- `cargo fmt --all -- --check`: clean
- `cargo clippy --workspace --tests -- -D warnings`: clean
- `make codegen-lint`: clean
- `make test-release-binary`: clean
- Preflight: ready-to-push

## Out of scope

- `Vec::set`, `Vec::append`/`extend`, `HashMap::insert`, and `HashSet::insert` are sibling consuming sites that share the same ownership-transfer pattern and need the same null-after-move treatment; filed as a follow-up.
- `dropFuncForMLIRType` defaults `includeStructTypes=false`, so `Vec<UserStructWithOwnedFields>` is not covered by the new guard; filed as a follow-up.

Fixes #1694


slepp:
PR opened. Filing follow-up issues now.

PR #1772, opened 2026-05-13
stdlib(iter): add lazy int iterator drains

gertybotbot:
# stdlib(iter): add a lazy filter+map+sum chain to `std::iter`

## Summary

Adds a defunctionalized **lazy iterator chain** to `std::iter` that fuses
`filter` / `map` / terminal into a single drain loop with **zero
intermediate `Vec` allocation**. This is a strict addition — no existing
API changes.

On the closure-bench workload (`sum(map(add5, filter(even, 1..=100_000)))`),
this closes about half of the historical 12× gap with Rust without any
compiler changes.

## Motivation — the 12× → 1.99× → 0.81× story

A prior profiling exercise put Hew at **~12× slower than Rust** on a
simple filter+map+sum chain. Three sources of slowdown:

1. **Eager iterators allocate intermediate `Vec<int>`** at every stage
   (~50% of CPU).
2. Closures cross `iter.*` as runtime `fn` pointers, not monomorphized
   (~5–10% of CPU; needs compiler work).
3. The input Vec build loop is ~30% of CPU on its own.

(1) and (3) are stdlib problems. A standalone proof-of-concept showed
that **fixing only (1)** — by replacing the
`filter_int → map_int → sum` pipeline with a fused drain — buys a
**1.99×** win on the same input shape. With a lazy `Range` source
(killing the input Vec build), Hew beats Rust by **1.23×** on the same
benchmark.

This PR ships the (1) fix as `lazy_*` additions to `std/iter.hew`. The
`Range` source is included; a future PR can rework the closure-bench
itself to use it for the headline "Hew > Rust" result.

## Benchmark results

In-stdlib bench (`examples/benchmarks/lazy_iter/run.sh`, 11 runs,
median µs, machine that built this PR):

| Variant | Median (µs) | Stdev (µs) | Runs |
|---------|------------:|-----------:|-----:|
| eager   |         749 |       116  |   11 |
| lazy    |         429 |        52  |   11 |

**Lazy speedup: ~1.75×** over eager on the Vec-source shape (the PoC
saw ~1.99× on a slightly different harness; same direction, same
magnitude).

PoC numbers across all four shapes are in `docs/rfcs/0001-lazy-iter.md`.

## API surface

All new functions live in `std/iter.hew`. They use the `lazy_` prefix
and follow the existing stdlib `_int` monomorphization convention.

```hew
// Sources
pub fn lazy_from_range(lo: int) -> fn(int) -> int
pub fn lazy_from_vec_int(xs: Vec<int>) -> fn(int) -> int

// Combinators (pass-through — kept for API symmetry)
pub fn lazy_filter_pred_int(pred: fn(int) -> bool) -> fn(int) -> bool
pub fn lazy_map_fn_int(f: fn(int) -> int) -> fn(int) -> int

// Terminals (fused drains)
pub fn lazy_sum(src, pred, mapf, n)             -> int
pub fn lazy_sum_filter(src, pred, n)            -> int
pub fn lazy_sum_map(src, mapf, n)               -> int
pub fn lazy_fold(src, pred, mapf, init, op, n)  -> int
pub fn lazy_collect_int(src, pred, mapf, n)     -> Vec<int>
pub fn lazy_take_int(src, pred, n, max_steps)   -> Vec<int>
```

Call-site reads like a pipeline:

```hew
import std::iter;

let src  = iter.lazy_from_range(1);
let even = iter.lazy_filter_pred_int((x: int) => x % 2 == 0);
let add5 = iter.lazy_map_fn_int((x: int) => x + 5);
let r    = iter.lazy_sum(src, even, add5, 100000); // 2_500_300_000
```

## Tests

Adds 13 `#[test]` cases to `tests/hew/iter_test.hew`:

- sources: `lazy_from_range` consecutive + zero-length, `lazy_from_vec_int`
- `lazy_sum` on the closure-bench workload (`2_500_300_000`)
- filter-all-out / map-only / filter-only drains
- `lazy_collect_int` empty + non-empty + round-trip through eager `iter.sum`
- `lazy_take_int` n-limit hit before and after `max_steps`
- `lazy_fold` basic + no-match init pass-through

**`hew test tests/hew/iter_test.hew` → 82 passed (69 existing + 13 new).**

## Compiler followups (4 known bugs)

The canonical `Lazy<T> { next: () -> Option<T> }` record + chained
methods API is **not yet expressible** in v0.4.0. Four independent
lowering / verification / type-check bugs block it. Each has a
minimised reproducer at `docs/rfcs/0001-lazy-iter-spikes/`:

1. **`Option<T>`-returning closures fail to lower** —
   `spike8.hew`. Blocks `next()`.
2. **Conditional yield in captured-state closure fails MLIR verifier** —
   `spike10.hew`, `spike13.hew`. Blocks stateful counters that return
   a sentinel.
3. **Nested closures-of-closures SIGSEGV at runtime** —
   `spike16.hew`. Blocks `filter(...).map(...)` stacking.
4. **Closure-typed struct fields type as `<error>` at call site** —
   `spike19.hew`. Blocks `Lazy<T>` from being a real record.

This PR works around all four by going defunctionalized: every `lazy_*`
entry point is a free function taking concrete `int`-typed closures, so
the compiler stays well inside the verified region. The RFC documents
each bug with a TODO to file the upstream issue.

## Migration / breaking changes

**None.** All names are new. No existing API is touched. Existing
`iter_test.hew` continues to pass alongside the new tests.

Down the road, when the four compiler bugs above are fixed, the
`lazy_*` free functions can stay as a thin compatibility layer that
forwards to a proper `Lazy<T>` record. Existing call sites keep working.

## Commits

This PR is structured as five logically separate commits on the
`stdlib-iter-lazy` branch:

1. `stdlib(iter): add Lazy<int> sources and combinator wrappers`
2. `stdlib(iter): add Lazy<int> drains — sum, fold, collect, take`
3. `stdlib(iter): tests for lazy_* combinators and closure-bench workload`
4. `bench(lazy_iter): lazy vs eager filter+map+sum comparison`
5. `docs(rfc): 0001 lazy iterators in std::iter`

## Notes for the reviewer

- The compiler binary that built this PR is `0.4.0`, matching upstream
  `Cargo.toml` at `c1fb40ae` (latest main when the branch was cut).
- The bundled `std/iter.hew` shipped with the binary diverged slightly
  from upstream `main` in comments only; the new content applies cleanly
  to upstream's version of the file.
- The bench harness is shell-driven, one-run-per-process, because
  v0.4.0 still has the `for + closures + now_nanos` MLIR bug. Once that
  is fixed, the harness can collapse into a single `std::bench::Suite`.
- The `lazy_filter_pred_int` and `lazy_map_fn_int` pass-throughs look
  pointless on first read. They exist so the call-site reads like a
  pipeline, and so a future `Lazy<T>` record can keep the same API
  surface without a renaming pass on user code. If the reviewer would
  rather drop them, removing both is a one-line change to each example
  in the docstrings.

## How to apply locally

The branch lives at `stdlib-iter-lazy` in the source repo this PR was
prepared from. If you want to verify the patch independently:

```bash
gh repo clone hew-lang/hew && cd hew
git fetch <fork> stdlib-iter-lazy:stdlib-iter-lazy
git checkout stdlib-iter-lazy
cargo build --release   # ~10 min cold
./target/release/hew test tests/hew/iter_test.hew
./examples/benchmarks/lazy_iter/run.sh
```

If no fork exists yet, the branch can be re-created from this PR's
five commits via `apply-to-upstream.sh` in the project working directory.

## Mergeable as-is?

**Yes, but read the constraints first.** This is intentionally a
**v0**: it covers `int` only, and the combinator API is a defunctionalized
free-function family rather than a record with chained methods. That
shape is forced by the four compiler bugs, not a design preference.

If the project would rather block this on fixing those bugs first and
ship `Lazy<T>` once — the perf number stays the same, the API just gets
prettier — that's a defensible call. The PoC has all the evidence
needed to prioritize the four bugs.

Either way, the benchmark, the RFC, and the spikes are deliverable
artifacts that benefit the language regardless of whether this exact
patch lands.


slepp:
Thanks for this — the RFC and the spike corpus are genuinely useful. The benchmark harness too.

I'd like to land a narrower slice of this PR rather than the full thing, and here's why:

The v0.5 IR refactor currently in flight (PRs #1773–#1785, more incoming) reshapes what `std/iter.hew` should look like in non-trivial ways — `@linear` / `@resource` annotation surface lands in the next cluster after the one merging today, sealed `select{}` widening lands after that, and `fork{}` structured concurrency lands alongside. Each of those touches the semantics that a lazy iterator wants to lean on (move discipline, scoped cancellation, stream-vs-iterator distinction). Locking in a 296-line stdlib API today commits us to maintaining a surface that will likely want to be reshaped post-cutover.

**What I'd like to merge from this PR**:
- `docs/rfcs/0001-lazy-iter.md` — the RFC text. Durable design value; survives any reshape.
- `docs/rfcs/0001-lazy-iter-spikes/spike{8,10,13,16,19}.hew` — the spike corpus. Concrete tradeoff exploration, perfect for informing the post-cutover design.
- `examples/benchmarks/lazy_iter/` (all of it — README, bench_eager.hew, bench_lazy.hew, results.md, run.sh, .gitignore) — the benchmark harness. Stays useful regardless of which API shape we end up with.

**What I'd like to defer**:
- `std/iter.hew` (296 lines, the proposed API) — re-propose this after the v0.5 spine completes and the value-class + concurrency clusters land. The RFC already documents the design space; this PR would be much shorter at that point and easier to land cleanly.
- `tests/hew/iter_test.hew` (138 lines) — tests for the deferred API; defer alongside.

Would you be willing to drop the two deferred files from this branch? Happy to land the rest right after. If you'd rather not split, I can land the RFC + spikes + benchmark into a separate PR myself and close this one with credit to your authorship in the commit trailer.

(For context: `std/iter.hew` would currently fail-closed at `compile-v05` for unrelated reasons — the v0.5 spine is integer-only right now; `Iter<i64>` and the higher-order combinators it relies on aren't yet in the spine subset. So the API surface is not testable end-to-end in the current state of the tree.)

PR #1792, opened 2026-05-14
build(make): build hew-emit-v05 helper alongside hew driver

### LINKED ISSUE #1791: compile-v05 fails on fresh clone: hew-emit-v05 helper not built by 'make hew' or 'cargo build -p hew-cli'
**Symptom**: on a fresh clone, the documented spine command fails:

```
$ cargo run --release -p hew-cli -- compile-v05 examples/v05/hello_int.hew --emit-dir /tmp/out
E_CUTOVER_UNSUPPORTED: llvm: helper binary `target/release/hew-emit-v05` not found next to `target/release/hew`; ensure `cargo build -p hew-codegen-rs --bin hew-emit-v05` has run
```

The diagnostic is good (it tells the user what to do), but the user shouldn't need to know about the helper. `make hew` doesn't build it either:

```
hew:
    cargo build -p hew-cli
```

**Why this matters**: the compile-v05 command is the user-visible v0.5 spine entrypoint. Discovered during a plan-progress audit; not blocking the spine but a real onboarding/diagnostic-value issue.

**Three fix options**:
1. `hew-cli`'s `Cargo.toml` declares a build-time dependency on the `hew-emit-v05` bin target so `cargo build -p hew-cli` transitively builds it.
2. The `hew:` Make target chains `cargo build -p hew-codegen-rs --bin hew-emit-v05`.
3. `compile-v05` shells out to `cargo build` if the helper is missing (probably the worst option — breaks if cargo isn't on PATH at runtime).

Option 1 is the cleanest; option 2 is the smallest. Option 1 needs verifying that cargo will resolve a workspace cross-package bin dependency at build time.

**Severity**: P1 dev-ergonomics. Easy fix; high diagnostic value.

slepp:
## Summary

`make hew` now also builds the `hew-emit-v05` helper binary. Previously, a fresh clone that ran `make hew` would produce only the main `hew` driver, leaving `compile-v05` unable to locate the helper and failing at runtime. Both binaries are now built together so the compile path works out of the box.

An internal documentation note is added to `docs/internal/v05-ir-ladder.md` recording the build dependency so future maintainers understand why the targets are coupled.

## Verification

- `git diff origin/main --stat`: 2 files changed (Makefile +6/-1, docs/internal/v05-ir-ladder.md +16/-0), no unintended files
- Preflight: `make ci-preflight` passed (148s, test-rust + fmt check)
- Change is Makefile target addition and internal docs only — no logic paths altered

## Out of scope

- No changes to the `compile-v05` runtime logic itself
- No changes to how `hew-emit-v05` is invoked or its output format
- Wiring `hew-emit-v05` into the default `all` target or a dedicated install target is left for a follow-up if needed

Fixes #1791

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -144,9 +144,12 @@ all: hew adze runtime stdlib assemble
 
 # ── Rust targets ────────────────────────────────────────────────────────────
 
-# Build the hew compiler driver (debug)
+# Build the hew compiler driver (debug).
+# hew-emit-v05 is the out-of-process LLVM object emitter required by
+# `hew compile-v05`; it must sit alongside the hew binary at runtime.
 hew:
 	cargo build -p hew-cli
+	cargo build -p hew-codegen-rs --bin hew-emit-v05
 
 # Build the adze package manager (debug)
 adze:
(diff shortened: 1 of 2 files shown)


slepp:
Superseded by v0.5 release refactors.

PR #1831, opened 2026-06-12
FreeBSD installer: fetch, sha256, release tarballs

denislemire:
## Summary
- Teach `installers/install.sh` to install on **FreeBSD x86_64** using published `hew-v*-freebsd-x86_64.tar.gz` release assets
- Use **`fetch(1)`** and **`sha256(1)`** when curl/sha256sum are absent (FreeBSD base)
- Clear error when **bash** is missing, with: `pkg install -y bash` then `fetch -o - https://hew.sh/install | bash`

**Scope:** installer script only. FreeBSD **port** (source build) will be a separate PR after we prove it on a builder VM.

## Test plan
- [x] `bash -n installers/install.sh`
- [x] Bastille jail `hew-dev` on inferno (FreeBSD 15): `bash install.sh --version 0.5.0 --prefix /root/.hew` → `hew version` / `adze --version` OK
- [ ] Merge companion PR on `hew-lang/hew.sh` for `public/install.sh`

### DIFF
--- a/installers/install.sh
+++ b/installers/install.sh
@@ -1,6 +1,7 @@
 #!/usr/bin/env bash
 # Hew installer — https://hew.sh
 # Usage: curl -fsSL https://hew.sh/install | bash
+#        fetch -o - https://hew.sh/install | bash   (FreeBSD — bash from ports/pkg)
 #        irm https://hew.sh/install.ps1 | iex  (Windows PowerShell)
 #        bash install.sh [--version <ver>] [--prefix <dir>] [--help]
 set -euo pipefail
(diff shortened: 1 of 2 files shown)


installers/install.sh
```diff
@@ -102,16 +103,37 @@ detect_platform() {
...
     Linux*) OS="linux" ;;
     Darwin*) OS="darwin" ;;
+    FreeBSD*) OS="freebsd" ;;
     *) err "unsupported operating system: $uname_s" ;;
     esac
 
     case "$uname_m" in
     x86_64 | amd64) ARCH="x86_64" ;;
-    aarch64 | arm64) ARCH="aarch64" ;;
+    aarch64 | arm64)
```
slepp, line 112: We could add arm64 builds for it as well, in a later release.

Added #1832 to track.

installers/install.sh
```diff
@@ -1,6 +1,7 @@
 #!/usr/bin/env bash
 # Hew installer — https://hew.sh
 # Usage: curl -fsSL https://hew.sh/install | bash
+#        fetch -o - https://hew.sh/install | bash   (FreeBSD — bash from ports/pkg)
```
slepp, line 4: I'll update the website with this for the FreeBSD installers as one of the cut and paste tab targets.

slepp (approved):

Issue #1832, opened 2026-06-12
Support FreeBSD ARM64


slepp:
FreeBSD is presently supported on x86_64, and Hew could also be supported on FreeBSD 15.0 aarch64/arm64.

slepp:
Linked in as a primary FreeBSD build target now.

PR #1866, opened 2026-06-13
ci(freebsd): add aarch64 matrix leg

slepp:
## Summary

Extends the nightly FreeBSD CI to cover aarch64 (arm64) alongside the existing amd64 leg, resolving #1832.

The single-arch job is converted to a two-entry matrix — `amd64` on `ubuntu-24.04` (unchanged) and `aarch64` on `ubuntu-24.04-arm`. Both legs run `vmactions/freebsd-vm@v1` with FreeBSD 15.0; the new leg passes `arch: aarch64` so the correct VM image is booted on the native arm64 runner.

Supporting details:
- The existing LLVM cmake flags already build `X86;AArch64;WebAssembly` backends, so no cmake change is needed.
- GitHub Actions LLVM cache keys are parameterised by arch (`freebsd15-aarch64-v3`) so each leg has its own cold-build and cached slot.
- `LLVM_SYS_221_PREFIX` was missing from `freebsd.yml` (present in `release.yml`); added to both legs so `llvm-sys-221` can discover the source-built LLVM prefix inside the VM.

## Verification

CI is the verification — watch the `Build & test (FreeBSD aarch64)` leg in this PR. If it goes green by default this PR is ready; if it fails in a way that requires porting work it will be closed and #1832 deferred.

## Out of scope

- Porting FreeBSD aarch64 (no code changes, only CI plumbing)
- Pre-building the LLVM toolchain for aarch64 (the job builds inline on first run)
- Release pipeline changes

### DIFF
--- a/.github/workflows/freebsd.yml
+++ b/.github/workflows/freebsd.yml
@@ -21,9 +21,17 @@ env:
 
 jobs:
   build-and-test:
-    name: Build & test (FreeBSD x86_64)
-    runs-on: ubuntu-24.04
+    name: Build & test (FreeBSD ${{ matrix.arch }})
+    runs-on: ${{ matrix.runs-on }}
     timeout-minutes: 180
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - arch: amd64
+            runs-on: ubuntu-24.04
+          - arch: aarch64
+            runs-on: ubuntu-24.04-arm
 
     steps:
       - name: Checkout code
@@ -34,12 +42,13 @@ jobs:
         uses: actions/cache/restore@27d5ce7f107fe9357f9df03efb73ab90386fccae  # v5.0.5
         with:
           path: llvm-freebsd-install.tar.zst
-          key: llvm-22.1.0-freebsd15-amd64-v3
+          key: llvm-22.1.0-freebsd15-${{ matrix.arch }}-v3
 
       - name: Build and test on FreeBSD
         uses: vmactions/freebsd-vm@a6de9343ef5747433d9c25784c90e84998b9d69a  # v1.4.6
         with:
           release: "15.0"
+          arch: ${{ matrix.arch }}
           mem: 8192
           prepare: |
             pkg install -y rust cmake ninja git pkgconf libffi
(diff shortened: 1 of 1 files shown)


slepp:
## DEFER #1832 — FreeBSD aarch64

### Experiment result

The aarch64 leg ran but timed out at the 180-minute ceiling, still inside the LLVM build ([333/2833] steps completed). It did not fail with a compile or link error — it ran out of time.

### Root cause: no KVM on ubuntu-24.04-arm

vmactions/freebsd-vm uses QEMU to run FreeBSD. On ubuntu-24.04 (x86_64), QEMU gets KVM acceleration and builds LLVM in ~55 min. On ubuntu-24.04-arm, the FreeBSD aarch64 guest runs without KVM (GitHub-hosted ARM runners do not expose /dev/kvm for nested guests), which drops the effective speed to TCG software emulation — roughly 20× slower.

Build rate evidence:
- amd64: 2833 LLVM steps in ~55 min (~1.1 sec/step)
- aarch64: 333/2833 steps in 128 min (~23 sec/step) → projected ~18 h for LLVM alone

This makes the CI leg unworkable even with a pre-built LLVM cache, because cargo builds would also run at ~20× under TCG. The 180-minute timeout cannot be stretched enough.

### Bonus finding (separate fix warranted)

The nightly FreeBSD x86_64 CI has been failing since at least 2026-06-08 with:

    error: No suitable version of LLVM was found system-wide or pointed

because freebsd.yml was missing `export LLVM_SYS_221_PREFIX=...` (present in release.yml). This PR's amd64 leg included that fix — the amd64 job compiled cleanly and ran 631/633 tests; the only 2 failures were WASM tests that require `wasmtime`, which is also absent from the FreeBSD CI environment. That fix should be extracted and landed as a standalone PR.

### Verdict

**DEFER #1832.** Blocker: vmactions/freebsd-vm is not viable on arm64 GitHub-hosted runners due to lack of KVM acceleration. Viable paths forward (none of which are low-effort) include: (a) a CI provider with native FreeBSD aarch64 runners, (b) cross-compile from linux-aarch64 to freebsd-aarch64 (no VM), or (c) wait for GitHub to expose KVM on ubuntu-24.04-arm runners.

PR #2211, opened 2026-06-25
fix(stdlib): fail closed on empty string.split separator

slepp:
## Summary

`std/string.hew` `split` now panics with `"string.split: separator must not be empty"` (exit 101) when called with an empty separator string, instead of silently returning a one-element list containing the original string. The previous behaviour was a silent misuse contract violation; callers that pass an empty separator get a clear, actionable diagnostic at the call site.

## Verification

- `hew run` with `split(s, "")`: exits 101, prints `string.split: separator must not be empty` — exact diagnostic confirmed
- Edge cases verified: empty input string, multibyte separator, separator longer than input — all correct
- No in-tree caller passes an empty separator (grepped `std/`, `examples/`, `tests/`)
- Vertical-slice fixtures: 2 positive split cases pass, 1 negative exit-101 case passes
- Fuzz-oracle expected-failures bump applied
- Preflight: ready-to-push

## Out of scope

- `split` to individual characters (e.g. `split(s, "")` as a chars explosion) — intentionally not added here; blocked on a separate UTF-8 iteration bug. Use `string.chars()` in the interim.
- `Vec` stdlib methods (`sort`, `insert`, etc.) — tracked in a separate lane.


slepp:
Closing in favour of the ratified behaviour: an empty separator splits the string into its characters (matching the language's documented direction), not a fail-closed error. That path needs a fix to the char-to-string conversion (a multi-byte UTF-8 truncation bug) first; I'm doing both in a follow-up rather than ship the fail-closed interim, which would contradict the intended semantics and have to be replaced.

Issue #2091, opened 2026-06-21
codegen: CodegenError::FailClosed carries no source span (black-hole diagnostics)


slepp:
When codegen fails closed (e.g. a `Move type mismatch`), the error is printed raw via `eprintln!("E_NOT_YET_IMPLEMENTED: {e}")` (hew-cli/src/main.rs:~445) with NO source location — a new user gets nothing to act on. Dogfooding surfaced this as the single most-disorienting failure mode.

`CodegenError::FailClosed(String)` should carry an optional `SourceSpan` (e.g. `FailClosed { msg: String, span: Option<SourceSpan> }`), propagated from the codegen site (the `Instr::Move` site and friends) through to the CLI diagnostic renderer, so a fail-closed always points at the offending source.

The specific ICE that surfaced this — a return-only match arm mistyping the match result as `Unit`, causing a ptr→i8 Move mismatch — is fixed separately (the HIR `lower_match_expr` divergence-skip). This issue is the general invariant: a fail-closed diagnostic must carry a span, not a black-hole panic.

Found via dogfooding pass 3.

PR #2320, opened 2026-06-30
fix(codegen,cli): point fail-closed codegen errors at the user's source (#2091)

slepp:
## Old failure mode
CodegenError::FailClosed had no span → raw eprintln(E_NOT_YET_IMPLEMENTED) with no location (the 'single most-disorienting failure mode').
## Invariant
A fail-closed codegen error renders file:line:col + caret through the normal diagnostic path; degrade to bare line only when truly spanless.
## Why this boundary
CodegenError::Spanned envelope attaches func.span once at the lower_function boundary — avoids touching ~1.3k FailClosed sites.
## Verification
codegen_emit_error_with_span_renders_source_attributed + boundary tests; workspace +5 tests; flake 3/3.
## Out of scope
Spanless MIR stays unwrapped FailClosed.

### DIFF
--- a/hew-cli/src/diagnostic.rs
+++ b/hew-cli/src/diagnostic.rs
@@ -166,14 +166,78 @@ pub(crate) fn codegen_diagnostic_prefix(error: &hew_codegen_rs::CodegenError) ->
         hew_codegen_rs::CodegenError::WasmUnsupportedSubstrate { .. } => {
             "E_CODEGEN_FRONT_WASM_UNSUPPORTED"
         }
+        // A span envelope classifies as its inner error: a spanned `FailClosed`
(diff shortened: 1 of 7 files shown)


slepp:
Superseded — reimplementing on a clean branch (fix/codegen-failclosed-span-v2). The cross-ecosystem gate correctly identified that this branch's approach (suppressing imported-function spans in MIR lowering) loses the fact at the wrong boundary and lacks correct cross-module attribution. The reimplementation carries the span on the CodegenError envelope (compute-once at emission, render at the boundary) and fails closed to a bare line for spans not safely attributable to the rendered source — never a false caret. Tracking issue #2091 stays open.

PR #2425, opened 2026-07-08
fix(mir): flag panic()/exit() call continuations unreachable, not just return

gertybotbot:
## What

`lower_direct_call` opened a `Never`-typed direct call's continuation block via a plain `start_block`, which always resets `cursor_unreachable = false` regardless of the callee's return type. For a `Never`-typed callee (the runtime `panic()`/`exit()` shims), that continuation can never actually be reached at runtime, but the flag said otherwise.

The `If`/`match` arm-lowering join-reachability check trusts `cursor_unreachable` to decide whether a join is live. An all-`panic()`/`exit()` tail match or if/else with a non-scalar return type therefore got its join wrongly admitted as reachable, and MIR's mixed-divergence recovery tried to move the substituted `Unit` (i8) result local into a non-scalar (ptr/struct) return slot — a codegen-front fail-closed abort:

```
E_CODEGEN_FRONT_FAIL_CLOSED: Move type mismatch: src=i8 dest=struct
```

Minimal repro (pre-fix, `hew check` fails):

```hew
type Point { x: i64, y: i64 }

fn pick(tag: i64) -> Point {
    match tag {
        1 => panic("one"),
        _ => panic("other"),
    }
}
```

The identical shape via `if`/`else` instead of a tail match reproduces the same abort. A control case with the same non-scalar return type but all-`return` arms already compiles cleanly — `return` statement lowering already flags `cursor_unreachable` directly; `panic()`/`exit()` route through `lower_direct_call` instead and never did.

## Fix

`lower_direct_call` now opens the continuation via `start_dead_block` (mirroring the early-return path's own dead-end convention, see its doc comment) when the callee's checker-resolved return type is `ResolvedTy::Never`, so the continuation is correctly flagged unreachable and every existing join-reachability gate treats a diverging `panic()`/`exit()` arm the same way it already treats an explicit `return` arm. One `matches!`-gated branch, no other call site touched (all three callers of `lower_direct_call` just return it straight through).

## Verification

- Live repro against a release build: pre-fix, both the tail-match and if/else forms abort with the exact `Move type mismatch: src=i8 dest=struct` above; post-fix, both compile, run, panic with the expected message, and exit 101 (`hew_panic`'s clean-exit contract). The all-return control case compiles and runs unaffected before and after.
- `cargo test -p hew-mir --lib`: 322/322 passing (no regressions; no isolated unit test added — see below for why).
- `cargo fmt --check`: clean.
- `cargo clippy -p hew-mir --lib -- -D warnings`: clean.
- Added two vertical-slice accept fixtures (`match_all_panic_arms_nonscalar_return.hew`, `if_else_all_panic_arms_nonscalar_return.hew`) wired into `tests/vertical-slice/run.sh` via the existing `run_accept_expect_panic` helper — verified directly against the real harness logic (extracted from `run.sh`) built against a release binary: both PASS. This matches the coverage convention the sibling `match_diverging_arm_result_type.hew` fix used for the same class of bug (an all-`return` tail-match join-reachability issue) — that fix also has no isolated `lower.rs` unit test; `cursor_unreachable`/`start_dead_block` have no unit-level test precedent anywhere in the file, so the `.hew` vertical-slice fixture is the established and sufficient coverage surface for this bug class.

## Scope

Touches only `hew-mir/src/lower.rs::lower_direct_call` (7-line diff plus a comment explaining the invariant) and the two new test fixtures. No other lowering path, call site, or public API changed.

### DIFF
--- a/hew-codegen-rs/src/llvm.rs
+++ b/hew-codegen-rs/src/llvm.rs
@@ -32935,14 +32935,19 @@ fn lower_terminator<'ctx>(
                 .llvm_ctx("call br")?;
         }
         Terminator::Trap { kind } => {
-            if matches!(*kind, TrapKind::MachineDispatchUnreachable) {
-                // Machine step dispatch is HIR-exhaustive. The residual
-                // fallthrough block is therefore impossible in well-typed
-                // programs and must not surface as a user-visible trap code.
+            if matches!(
+                *kind,
+                TrapKind::MachineDispatchUnreachable | TrapKind::UnreachableCallContinuation
+            ) {
+                // Machine step dispatch is HIR-exhaustive, and a
+                // `Never`-typed call's continuation is compiler-proven dead
+                // (the call always diverges) — neither is a runtime check,
+                // so neither should surface as a user-visible trap exit
+                // code. Both lower to a bare LLVM `unreachable`.
                 fn_ctx
                     .builder
                     .build_unreachable()
-                    .llvm_ctx("machine dispatch unreachable")?;
+                    .llvm_ctx("compiler-proven unreachable")?;
                 return Ok(());
             }
             // The exit-code constants here are single-sourced from the runtime
@@ -32961,6 +32966,9 @@ fn lower_terminator<'ctx>(
                     HEW_TRAP_MACHINE_DISPATCH_UNREACHABLE as u64
                 }
                 TrapKind::ExhaustivenessFallthrough => HEW_TRAP_EXHAUSTIVENESS_FALLTHROUGH as u64,
+                TrapKind::UnreachableCallContinuation => {
+                    unreachable!("handled by the build_unreachable() branch above")
+                }
             };
             emit_trap_with_code(fn_ctx, code, "trap")?;
         }
--- a/hew-mir/src/dump.rs
+++ b/hew-mir/src/dump.rs
@@ -1692,6 +1692,7 @@ fn render_trap_kind(kind: TrapKind) -> &'static str {
         TrapKind::SupervisorChildUnavailable => "SupervisorChildUnavailable",
         TrapKind::MachineDispatchUnreachable => "MachineDispatchUnreachable",
         TrapKind::ExhaustivenessFallthrough => "ExhaustivenessFallthrough",
+        TrapKind::UnreachableCallContinuation => "UnreachableCallContinuation",
     }
 }
 
--- a/hew-mir/src/lower.rs
+++ b/hew-mir/src/lower.rs
@@ -9260,6 +9260,24 @@ struct Builder {
     /// it is observed to be empty, so the dead-end does not appear as
     /// an additional `Return` exit in `drop_plans`.
     cursor_unreachable: bool,
+    /// Set alongside `cursor_unreachable` ONLY when the current dead cursor
+    /// was opened as the `next` of an already-sealed `Terminator::Call`
+    /// (a `Never`-typed callee's continuation) rather than from an
+    /// explicit `return`'s own seal. `finalize_blocks`'s empty-dead-cursor
+    /// drop is safe for the `return` case because nothing else references
+    /// that block's id (`Terminator::Return` has no successor field), but
+    /// a `Terminator::Call { next, .. }` DOES reference this id — dropping
+    /// an empty block here would leave the already-committed `Call`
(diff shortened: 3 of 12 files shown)


slepp (dismissed):
Thanks — the diagnosis here is exactly right, and detecting divergence through the checker-resolved Never type (rather than a shim list) is the correct mechanism. I verified the claim in source: `start_block` does unconditionally reset `cursor_unreachable`, and the If/match join-reachability logic trusts that flag. Both new fixtures pass, and the checked-MIR goldens are byte-identical under this change.

The problem is a regression this uncovers in the defer drain. CI caught it on the pre-existing `defer_early_return.hew` (`E_NOT_YET_IMPLEMENTED: fail-closed: Call next bb3 missing`), and I reproduced it locally with a PR-head build and pinned the mechanism via `--dump-mir raw`:

- the function-end defer drain re-materializes the deferred `exit(42)` into the now-dead post-return cursor;
- `lower_direct_call` seals it with `call exit -> bb3` and opens bb3 as a dead block;
- bb3 stays empty and dead, so `finalize_blocks` drops it — but unlike `Return` (the only prior client of that drop convention), `Terminator::Call` carries `next`, so codegen fails closed on the dangling reference.

Two more shapes fail identically: `defer exit(42); println("x")` and `defer panic("cleanup"); return 7`. Tail-position statement panics survive only because expression statements push an evaluate marker into the continuation.

For a revision: keep the dead-marking — it genuinely fixes the reachability bug — but guarantee a Call's `next` block always materializes. The `drop_empty_dead_cursor` predicate currently can't distinguish a Return-sealed dead cursor (safe to drop) from a Call-referenced continuation (must be kept, or the call retargeted). A revision should keep `defer_early_return` green, add a `defer panic()` / `defer exit()` regression fixture, and pass the full vertical slice plus checked-MIR verify.

Happy to re-review as soon as that's up.

slepp:
Thanks for the revision — the explicit-`return`-before-defer case (#2426) is fixed, and the join-reachability half of the original fix is correct and does genuinely close #1913.

The core regression is still open, though. The new guard in `function_body`'s no-tail-expression branch keys off `cursor_unreachable`, which is only true when the last statement is itself a diverging `return`. When the function tail is a live cursor — an ordinary trailing statement, or no other statement at all — the guard doesn't apply, and `emit_pending_defers` still drains the `Never`-typed call into a dead continuation block that `finalize_blocks`' `drop_empty_dead_cursor` then removes, leaving the committed `Call`'s `next` dangling. Reproduces cleanly against current main:

```
fn main() { defer exit(42); println("x"); }          // Call next bb2 missing
fn main() { defer exit(42); }                         // Call next bb1 missing
fn main() { defer panic("cleanup"); println("x"); }   // Call next bb2 missing
```

All three compile clean on main and fail on this branch.

The fix needs to live at the seam that owns the invariant — no terminator should ever reference a block id absent from the finalized block list — rather than at the drain site. Two directions that would work:

1. A `Never`-typed callee shouldn't seal a `Terminator::Call { next }` pointing at a droppable continuation in the first place — emit a no-successor/terminal call instead.
2. If the block model needs a `next`, materialize that continuation with an `Unreachable`/`Return` terminator so `finalize_blocks` keeps it, instead of opening it via `start_dead_block` and letting `drop_empty_dead_cursor` collect it.

Either way, keep the Never continuation `cursor_unreachable`-flagged — the join-reachability fix depends on that flag. The right shape is "make the block survive finalization," not "make the cursor live again."

New fixtures should cover the live-cursor cases specifically (`defer exit(42); println("x")`, bare `defer exit(42);`, and the `defer panic(...)` twin) — that's exactly the gap that let this ship green twice.

slepp:
Verified — the third revision (`cbd7409b`) fixes this at the right seam. It adds a dedicated trap (`UnreachableCallContinuation`) so a `Never`-typed call's dead continuation gets a real terminator instead of being silently dropped by `finalize_blocks`, while keeping it `cursor_unreachable`-flagged — so the #1913 join-reachability fix isn't regressed either.

Built the head and ran all three original regressions plus the two previously-fixed cases plus 21 adversarial shapes (nested/mixed defers, loops, match arms, tail returns) — all correct. `checked-mir-verify` is clean, no golden drift. The three live-cursor fixtures that were the actual gap are now wired into the suite with real assertions.

Dismissing my earlier review — this is resolved. Thanks for sticking with it through two rounds.

PR #1476, opened 2026-04-26
fix(std/net/tls): distinguish read/write error classes via out_status + last_error

slepp:
Closes #1413

thew_tls_read previously collapsed EOF, half-close, WouldBlock, TLS alerts, and IO errors into an empty HewVec. hew_tls_write collapsed every error class into -1. Higher-level Hew code could not distinguish orderly EOF from a force-close mid-stream.

This change adds an explicit out_status parameter on both FFI fns and routes diagnostic detail through hew_tls_last_error (added in #1473).

### Test
- Force-close mid-stream → distinguishable from orderly EOF.
- Successful read clears stale error from prior failure.
- make ci-preflight and make test-hew clean.

### Compatibility
FFI signature change. Hew-side wrappers in std/net/tls/tls.hew updated; surface Hew API now Result-shaped.

PR #2440, opened 2026-07-08
fix(std): set hew_tls_connect's last_error, drain it in try_read_bytes

gertybotbot:
Fixes two independent, previously-open diagnostic-contract bugs surfaced during source review (`reviews/hew-pr-bad-issues.md`) against PR #1476 and PR #1728. Both still reproduced against current `origin/main` before this fix.

## 1. `hew_tls_connect()` never set `last_error()` on failure (from #1476)

`hew_tls_connect()` returned null on all three failure paths — invalid host C string, invalid port, and `connect_tls(...) => Err(_)` — without ever calling `set_tls_last_error()`. `std/net/tls/tls.hew`'s own doc comment for `last_error()` says "Failed `connect()` calls update this string", but the string stayed empty on every failure. The read/write paths already follow this convention (`set_tls_last_error(message)` on each failure arm); `connect()` just never did.

Fix mirrors that convention on all three failure paths, using `connect_tls`'s own error `Display` for the network-failure case so the message is diagnostic rather than generic (e.g. `"hew_tls_connect: invalid dns name"` for an unresolvable/empty host).

## 2. `try_read_bytes` left a stale error message parked in the thread-local (from #1728)

`std/fs.hew`'s `try_read_bytes` reads `hew_stream_last_errno()` alone on its failure path but never drains `hew_stream_last_error()`. The two are documented as independent thread-local reads (see `hew-runtime/src/stream_error.rs`'s own module doc comment), so a failed `try_read_bytes` left a stale `hew_file_read_bytes: ...` message sitting in the thread-local — which could leak into a later, unrelated `hew_last_error()` / `hew_stream_last_error()` read (e.g. a subsequent `fs.read()` panic would report the *earlier* failure's text instead of its own).

The sibling `try_read` already drains this correctly (`let _ = hew_stream_last_error();`); `try_read_bytes` now mirrors it.

## Tests

- `hew-std/src/tls.rs`: strengthened 4 existing `hew_tls_connect` tests to assert `last_error()` content on each failure shape, added `connect_refused_sets_last_error` covering the `connect_tls(...) => Err(_)` arm specifically.
- `hew-runtime/src/file_io.rs`: added `read_bytes_failure_message_survives_a_standalone_errno_read`, pinning the FFI-layer precondition (`hew_stream_last_errno()` and `hew_stream_last_error()` are independent reads) that `try_read_bytes`'s fix depends on.

## Verification (fresh `origin/main`, before pushing)

- `cargo test -p hew-std --lib tls:: -- --test-threads=1`: 32/32 passing.
  - Note: the crate's pre-existing `tls::` suite aborts under the *default parallel* test runner independent of this change — reproduced identically on an unmodified `origin/main` worktree (a pre-existing environment flake in `close_reaps_a_live_blocked_reader_without_hanging`, not caused by or related to this diff). Single-threaded is clean.
- `cargo test -p hew-runtime --lib file_io::`: 69/69 passing.
- `hew test tests/hew/fs_stream_test.hew`: 15/15 passing.
- `tests/vertical-slice/run.sh`: 419/419 passing, 0 failures.
- `cargo fmt -p hew-std -p hew-runtime -- --check` and `cargo clippy -p hew-std -p hew-runtime --lib -- -D warnings`: both clean.

Scope is intentionally narrow — no other behaviour changed.

### DIFF
--- a/examples/v05/checked-mir/golden/stream_blocking_recv.elab.mir
+++ b/examples/v05/checked-mir/golden/stream_blocking_recv.elab.mir
@@ -852,10 +852,11 @@ fn fs$try_read_bytes -> Result<bytes, IoError>
     bind BindingId(52) data site=SiteId(465) ty=bytes
     bind BindingId(53) errno site=SiteId(468) ty=i32
     use BindingId(53) errno site=SiteId(472) ty=i32 intent=Read
-    use BindingId(53) errno site=SiteId(478) ty=i32 intent=Read
-    use BindingId(52) data site=SiteId(482) ty=bytes intent=Read
-    agg_alias BindingId(52) data site=SiteId(482) ty=bytes
+    use BindingId(52) data site=SiteId(484) ty=bytes intent=Read
+    agg_alias BindingId(52) data site=SiteId(484) ty=bytes
     return site=Some(SiteId(464)) ty=Result<bytes, IoError>
+    eval site=SiteId(475) ty=string
+    use BindingId(53) errno site=SiteId(480) ty=i32 intent=Read
     drop BindingId(52) data ty=bytes
   drop_plans:
     call[bb0 hew_file_read_bytes -> bb1] ->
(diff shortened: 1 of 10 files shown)


slepp:
Both fixes verified real when I gated this — the connect path especially, since the error string is the only diagnostic surface there. You beat me to the golden regeneration with your rebase; I pushed one addition to your branch (maintainer edit): a negative test in fs_stream_test.hew that fails without the std/fs.hew drain line (verified by reverting it) — the Rust-side test pinned a precondition that already held, so this gives the drain itself teeth. 16/16 green on your head. Re-gating once CI cycles.

PR #2441, opened 2026-07-08
fix(runtime): cap oversized lines that arrive with an embedded newline

gertybotbot:
## Bug

`LinesStream::next()` (`hew-runtime/src/stream.rs`) checks for a complete
newline-terminated line *before* checking the `MAX_LINE_BUFFER_SIZE` cap. A
single upstream chunk that is already over the cap and happens to carry its
own trailing `\n` in that same chunk bypasses the cap entirely: the
newline-scan branch drains and returns the whole oversized line, because
the size-limit check below it only runs when no newline is found in the
buffered bytes.

Confirmed live before touching any code, with a temporary probe test
(reverted before this PR): a single chunk of `MAX_LINE_BUFFER_SIZE + 500`
bytes followed by `\n`, delivered all at once (`item_size = 0`), returned
`num_items=1, first_len=1049076` — the whole oversized line, uncapped.

The existing test suite doesn't cover this shape: `lines_single_oversized_chunk_capped_at_limit`
tests an oversized chunk with **no** newline, and
`lines_newline_immediately_after_limit_no_spurious_empty` /
`lines_crlf_immediately_after_limit_no_spurious_empty` deliver the `\n` in a
**separate** chunk (`chunk_size = MAX_LINE_BUFFER_SIZE`) after the cap has
already fired — neither covers a newline landing in the *same* over-cap
chunk.

## Fix

Gate the newline fast path on the delimiter position: only take it if the
line (through its delimiter) fits within `MAX_LINE_BUFFER_SIZE`. If a
newline exists but only past the cap, fall through to the existing
size-limit drain, which already returns exactly `MAX_LINE_BUFFER_SIZE`
bytes and sets `skip_next_delimiter` so the leftover newline is consumed
correctly (or, if it's no longer at the front of the remaining buffer,
simply becomes an ordinary line-delimiter on a later call — also correct).

No new field, no change to the public `Stream`/`lines()` surface, no
change to any existing passing case.

## Tests

- Added `lines_oversized_single_chunk_with_newline_at_cap_boundary_is_capped`
  (newline lands exactly at the cap) and
  `lines_oversized_single_chunk_with_newline_past_cap_is_capped_then_splits`
  (newline lands further past the cap, with trailing data after it).
- All 105 pre-existing `stream::tests` pass unmodified (107/107 total for
  the module after the additions).
- `cargo test -p hew-runtime --lib`: 2020/2020 passing.
- `cargo fmt -p hew-runtime --check`: clean.
- `cargo clippy -p hew-runtime --lib -- -D warnings`: clean.
- `tests/vertical-slice/run.sh`: exits 0, no failures.

## Context

Originally flagged in #323 (closed unmerged upstream) as a stream-buffer
cap edge case; re-verified this reproduces on current `main` before
implementing the fix — the mechanism is genuinely still present, unrelated
to whatever caused that PR to close.

### DIFF
--- a/hew-runtime/src/stream.rs
+++ b/hew-runtime/src/stream.rs
@@ -429,7 +429,18 @@ impl StreamBacking for LinesStream {
             // found) is bypassed entirely and an arbitrarily large line can
             // be returned whole, defeating the point of the cap.
             if let Some(pos) = self.buf.iter().position(|&b| b == b'\n') {
-                if pos < MAX_LINE_BUFFER_SIZE {
+                // `pos` is the index of the delimiter itself, not the
+                // content length: for a CRLF-terminated line, the '\r' just
+                // before it is a delimiter byte, not content. When pos ==
+                // MAX_LINE_BUFFER_SIZE and that preceding byte is '\r', the
+                // actual content (everything before the '\r') still fits
+                // within the cap, so this must still take the fast path —
+                // otherwise the '\r' gets misclassified as content and
+                // leaked into the returned line by the size-limit drain
+                // below.
+                if pos < MAX_LINE_BUFFER_SIZE
+                    || (pos == MAX_LINE_BUFFER_SIZE && self.buf[pos - 1] == b'\r')
+                {
                     let mut line: Vec<u8> = self.buf.drain(..=pos).collect();
                     // Strip the trailing newline delimiter (and \r for CRLF).
                     if line.last() == Some(&b'\n') {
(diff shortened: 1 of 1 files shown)


slepp:
Good catch on the bypass — the fix is right and I want it in. I found one boundary regression while gating it and pushed a follow-up commit to your branch (maintainer edit): the new length check measures the newline index, which counts a CRLF's `\r` as content — a single-chunk line of exactly cap−1 content plus `\r\n`, which round-trips clean today, came back as cap bytes with the `\r` leaked into the data. The pushed commit widens the boundary condition to tolerate the trailing `\r` and adds a revert-proven regression test for that shape; the LF at-cap/cap+1 and CRLF-at-cap cases were already exact. Stream suite is green (132/132) with the runtime lane at 2487. Re-gating now — thanks for the contribution.

PR #2442, opened 2026-07-08
fix(runtime): opportunistically reap finished deferred-teardown threads

gertybotbot:
## Problem

`defer_actor_free_on_background_thread` (`hew-runtime/src/actor.rs`) pushes every deferred self-free `JoinHandle` into `LiveActors::deferred_teardown_threads`, and two analogous deferred supervisor-stop paths in `scheduler.rs`/`supervisor.rs` do the same. The only drain, `drain_deferred_teardown_threads`, runs solely at runtime shutdown (via `cleanup_all_actors`/`hew_runtime_cleanup`).

In a long-lived native runtime with repeated actor self-terminate/free, completed background-free threads stay joinable — and their OS thread-handle resources retained — in that `Vec` for the runtime's entire lifetime, with no opportunistic reap of already-finished handles during normal operation. This is an unbounded-until-shutdown accumulation, not a leak in the traditional sense (everything is correctly joined and freed at shutdown), but real resource growth for any long-running native Hew process that frees many actors over its lifetime.

## Fix

`push_deferred_teardown_thread` now pulls out and joins any handles whose thread has already finished (`JoinHandle::is_finished()`, a non-blocking poll — just checks whether the thread's `Thread` handle has been marked dead) before/while pushing the new handle. `.join()` on an already-finished handle is itself non-blocking, so this cannot stall a hot actor-free/supervisor-stop call site waiting on some *other* still-running deferred-teardown thread.

The extraction (swap-remove under the registry lock) and the actual `.join()` calls (outside the lock) are deliberately split, so the lock is only held long enough to mutate the `Vec`, never across a join.

This does **not** change the shutdown join-before-free barrier: `drain_deferred_teardown_threads` is untouched and still takes + joins every remaining handle before any global actor sweep, exactly as before.

## Testing

- New regression test (`push_deferred_teardown_thread_reaps_already_finished_handles`) exercising: several already-finished handles being fully reaped by a subsequent push, a still-running gated handle being correctly retained across a later push (proving the reap never touches a genuinely still-running thread), and the shutdown drain still correctly joining that gated handle once released. Uses the existing `pub(crate)` `scheduler::NoWorkerSchedulerForTest` guard for a real, deterministic, worker-less runtime — no new test infrastructure needed.
- `cargo test -p hew-runtime --lib`: 2019/2019 passing (including the pre-existing `runtime_cleanup_sweeps_before_detach_and_drops_runtime_once` ordering oracle, which specifically exercises `push_deferred_teardown_thread`/shutdown-drain interaction and still passes unmodified).
- `cargo fmt -p hew-runtime --check`: clean.
- `cargo clippy -p hew-runtime --lib --tests -- -D warnings`: clean.
- `tests/vertical-slice/run.sh`: exits 0, full suite.

## Scope

Single file changed (`hew-runtime/src/lifetime/live_actors.rs`), no public API changes, no changes to the shutdown-path semantics.

### DIFF
--- a/hew-runtime/src/lifetime/live_actors.rs
+++ b/hew-runtime/src/lifetime/live_actors.rs
@@ -444,8 +444,8 @@ pub(crate) fn drain_all_for_cleanup() -> HashMap<u64, ActorPtr> {
 
 #[cfg(not(target_arch = "wasm32"))]
 pub(crate) fn push_deferred_teardown_thread(handle: std::thread::JoinHandle<()>) {
-    // Opportunistic reap: pull out and join any handles for threads that have
-    // already finished before growing the vector further. `JoinHandle::is_finished`
+    // Opportunistic reap: after pushing the new handle, pull out and join any
+    // handles for threads that have already finished. `JoinHandle::is_finished`
     // is a non-blocking poll (just checks whether the thread's `Thread` has been
     // marked dead), and `.join()` on an already-finished handle returns
     // immediately without blocking the caller — so this cannot stall a hot
@@ -479,7 +479,7 @@ pub(crate) fn push_deferred_teardown_thread(handle: std::thread::JoinHandle<()>)
             finished
         });
     for handle in finished {
-        // SAFETY/ordering: only handles that reported `is_finished() == true`
+        // Ordering: only handles that reported `is_finished() == true`
         // under the lock above reach here, so this join is non-blocking.
         if handle.join().is_err() {
             eprintln!("hew: warning: deferred teardown thread panicked");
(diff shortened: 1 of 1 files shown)


slepp:
Gated this thoroughly — the concurrency substance is clean: the reap only extracts finished handles under the lock and joins outside it, the re-defer re-entrancy path takes the same mutex briefly without nesting, and every handle is joined by exactly one of the opportunistic reap or the shutdown drain. The exact-count tests reading 4 on the pre-fix code are exactly the right shape.

I pushed one comment-only commit to your branch (maintainer edit): a doc-comment cited a review-notes reference that does not exist in this repo, so I dropped the citation and corrected three comment wordings (the ordering note on the join, the self-count description next to the == 0 assertion, and the push-then-scan phrasing). No code changes; your teardown tests still pass 14/14. Final review pass is running now.

PR #2450, opened 2026-07-08
fix(mir): stop empty exclusive descending ranges underflow-trapping

gertybotbot:
## Bug

`(a..b).rev()` lowers by computing `counter = raw_end - 1` (checked sub, trap on overflow) unconditionally in the pre-header of `lower_for_range`, before the loop header ever checks whether the range is empty. For an empty exclusive descending range ending at the counter type's representable minimum -- `(0u32..0).rev()` or `(i32::MIN..i32::MIN).rev()` -- `raw_end - 1` underflows the checked subtract and traps `IntegerOverflow` instead of running zero iterations, which is what every other empty-range shape in this function correctly does.

Confirmed live before touching code, against `origin/main` `03ff774d84f1830f3871816f05741c9b8cb34549`:

```
let lo: u32 = 0; let hi: u32 = 0;
for i in (lo .. hi).rev() { println(i); }
println(999);
// -> hew: trap in main context: IntegerOverflow   (should print only 999)
```

Both underflow-boundary shapes reproduce identically (unsigned ending at 0, signed ending at `i32::MIN`). Contrast cases already worked correctly before this fix -- a signed empty non-boundary range (`(5i32..5).rev()`) and a non-empty unsigned range (`(0u32..3).rev()`) both ran fine -- isolating the defect to specifically the empty-range-ending-at-the-representable-minimum case.

## Fix

Gate the `b - 1` decrement on emptiness first (`raw_start == raw_end`, width/signedness-agnostic bit-pattern equality via `CmpPred::Eq`). An empty range now branches straight to the loop's `exit_bb`, bypassing the header/body cycle entirely -- rather than routing a synthesized `counter` value through the header, since no single representable sentinel satisfies `counter >= bound` as false for every counter width without reintroducing the same underflow problem being fixed. A non-empty range takes the existing decrement-with-trap path completely unchanged.

The four `header_bb`/`body_bb`/`inc_bb`/`exit_bb` block-id allocations are hoisted earlier in the function to make this possible -- this is a pure id allocation with no emission side effect (`alloc_block` just returns a monotone id; every block in this function is already forward-referenced by id well before `start_block` is called on it, matching the existing pattern throughout). `exit_bb` already has multiple predecessors elsewhere in this same function (the header's false branch, `break`, and the descending decrement's own underflow-exit in `inc_bb`), so a third predecessor from the new emptiness gate isn't a new structural shape for this CFG.

No change to any non-empty range's lowering, no new field, no change to any public surface.

## Tests

Wired two previously-orphaned fixtures that existed on disk but were never referenced by `run.sh` (`for_range_rev.hew`, `for_range_rev_step_by.hew`) into the suite with new `.expected` files, and added two new regression fixtures covering both underflow-boundary shapes found live: `for_range_rev_empty_unsigned_zero` (`(0u32..0).rev()`) and `for_range_rev_empty_signed_min` (`(i32::MIN..i32::MIN).rev()`), both asserting exit 0 and stdout `999` (proving zero iterations ran, not a trap).

Verified before pushing:
- `cargo test -p hew-mir --lib`: 322/322
- `cargo fmt -p hew-mir --check`: clean
- `cargo clippy -p hew-mir --lib -- -D warnings`: clean
- full `tests/vertical-slice/run.sh`: exits 0, all fixtures pass including the 4 `for_range_rev*` fixtures explicitly
- manually re-ran every contrast case from the original diagnosis (signed empty non-boundary, unsigned non-empty, existing `for_range_rev`/`for_range_rev_step_by` fixtures) to confirm zero regressions


slepp:
Thanks for the fix here — the underflow trap on `(0u32..0).rev()` / `(i32::MIN..i32::MIN).rev()` was real and this closes it cleanly.

I found one gap while reviewing: the emptiness gate tests `raw_start == raw_end`, but an exclusive range `a..b` is empty whenever `a >= b`, not only when `a == b`. The `a > b` subclass — e.g. `(5u32..0).rev()` or `(0i32..i32::MIN).rev()` — is also empty, but the equality check doesn't fire for it, so it falls through to the same unconditional `raw_end - 1` decrement and hits the identical `IntegerOverflow` trap this PR sets out to fix.

I pushed `7e5778579` on top of your branch to close it: swaps the gate to the signedness-appropriate `>=` predicate (mirroring the existing `SignedGreaterEq`/`UnsignedGreaterEq` selection the loop header already uses), corrects the doc comment's `==` reasoning, and adds two regression fixtures for the `a > b, b == type_min` shape (unsigned and signed). Confirmed both new fixtures trap without the change and pass with it; full gate suite (hew-mir/hew-codegen-rs/hew-runtime tests, vertical-slice, test-hew-ratchet, checked-mir-verify, hew-check-all, workspace clippy, fmt) is green.

PR #2459, opened 2026-07-08
fix(dist): bind CTRL_LINK_DOWN to the handshake-authenticated peer

gertybotbot:
## Bug

`CTRL_LINK_DOWN` was delivered by `ref_id` alone: `handle_control_frame` dispatched it to `handle_link_down_frame(control)` with no connection or peer identity, unlike the `CTRL_SWIM` arm, which resolves and threads `peer_node_id_for_conn(mgr, conn_id)` before accepting attributed data. `dist_monitor::deliver_link_down_to_ref` matched solely on `watchers.get_mut(&ref_id)` + `Pending` + `WatcherAction::Link` and never compared the stored `entry.remote_node_id` to the frame's actual sender.

`WatcherEntry` already records the peer this node registered the link against, so the existing doc comment ("the EXIT fires exactly once and ONLY for a link entry this node registered") is an entry-**locality** guard, not a peer-**binding** guard: it stops a forged `ref_id` this node never linked, but does nothing to stop a *different*, genuinely handshake-authenticated peer from firing a link this node registered against a *third* node.

Cross-node link `ref_id`s are allocated from a monotonic, enumerable namespace (`DIST_REF_ID_BASE = 2^48`), not a secret, so any connected peer that guesses or learns a pending `ref_id` could send `CTRL_LINK_DOWN{ref_id, reason}` and force a `CrashLinked` `EXIT` (with an attacker-chosen reason) into a local actor whose real linked peer is still alive.

Verified live before fixing: a temporary unit probe modelling the delivery primitive (register a `CrashLinked` link against node 7, then `deliver_link_down_to_ref` as if the frame had actually arrived from node 99) fired the cascade unconditionally; reverted before committing.

## Fix

Resolve `peer_node_id_for_conn(mgr, conn_id)` in the `CTRL_LINK_DOWN` arm of `handle_control_frame` (mirroring the `CTRL_SWIM` arm exactly) and thread it through `handle_link_down_frame` → `hew_node::handle_inbound_link_down` → `dist_monitor::deliver_link_down_to_ref`, which now rejects the delivery unless the stored `entry.remote_node_id` equals the authenticated sender. `authenticated == 0` (no live/registered connection for `conn_id`) cannot match any real link, since node ids are never `0` elsewhere in this file, so it fails closed without a separate early-return.

Added `link_down_from_wrong_authenticated_peer_is_rejected`: confirmed load-bearing by reverting the peer check and watching the test fail with the exact forged-delivery symptom, then restoring the fix. Updated the three pre-existing `deliver_link_down_to_ref` call sites in `dist_monitor::tests` to pass each entry's own registered peer as the now-required `authenticated_peer` argument; behaviour of those tests is unchanged.

## Verification

- `cargo test -p hew-runtime --lib` — 2020/2020 (was 2019, +1 new)
- `cargo fmt -p hew-runtime --check` — clean
- `cargo clippy -p hew-runtime --lib --tests -- -D warnings` — clean
- `tests/vertical-slice/run.sh` — 427/427

## Scope

This PR covers only the LINK cascade (`CTRL_LINK_DOWN`). The sibling MONITOR family (`CTRL_MONITOR_REQ`/`CTRL_DEMONITOR`/`CTRL_MONITOR_DOWN`) has the identical peer-binding gap and is tracked separately (not touched here to keep this fix reviewable and scoped to one cascade).

### DIFF
--- a/hew-runtime/src/connection.rs
+++ b/hew-runtime/src/connection.rs
@@ -1247,10 +1247,13 @@ fn handle_link_down_frame(mgr: *mut HewConnMgr, conn_id: c_int, control: &Contro
     let mgr_ref = unsafe { &*mgr };
     // Peer-binding defence: resolve the handshake-authenticated identity of the
     // connection this frame arrived on. `authenticated == 0` (no active,
-    // registered connection for `conn_id`) still flows into the ref-id lookup
-    // below and simply cannot match any real link's `remote_node_id` (a valid
-    // node id is always non-zero), so it fails closed the same way an actual
-    // peer mismatch does — no separate early return required.
+    // registered connection for `conn_id`, e.g. mid-teardown) is NOT
+    // guaranteed to fail closed on its own — `handle_link_req_frame` performs
+    // no peer authentication on an inbound `CTRL_LINK_REQ`, so any connected
+    // peer can set `linker_node_id` to 0 and land a `WatcherEntry` with
+    // `remote_node_id == 0`. `deliver_link_down_to_ref` rejects
+    // `authenticated_peer == 0` explicitly, before it is ever compared to a
+    // stored `remote_node_id`.
     let authenticated = peer_node_id_for_conn(mgr_ref, conn_id);
     crate::hew_node::handle_inbound_link_down(payload.ref_id, authenticated, payload.reason);
 }
--- a/hew-runtime/src/connection.rs
+++ b/hew-runtime/src/connection.rs
@@ -1058,7 +1058,7 @@ fn handle_control_frame(
             return;
         }
         CTRL_LINK_DOWN => {
-            handle_link_down_frame(control);
+            handle_link_down_frame(mgr, conn_id, control);
             return;
         }
         other => {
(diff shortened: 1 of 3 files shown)


slepp:
Thanks for this fix — the peer-binding gap it closes is real and the mechanism is correct. I pushed one follow-up commit (5f48630) directly onto this branch to close a narrow hole in the fail-closed reasoning before merging.

The commit's claim that "node ids are never 0 elsewhere in this file" isn't true in this codebase. `handle_link_req_frame` (`connection.rs`) does zero peer authentication on an inbound `CTRL_LINK_REQ` — any connected peer can set `linker_node_id` to 0 in that payload, which flows straight through `handle_inbound_link_req` into `register_link_watcher(payload.linker_node_id, ...)` and lands a `WatcherEntry` with `remote_node_id == 0` in the same watcher table `deliver_link_down_to_ref` reads.

If `authenticated_peer` is also 0 at delivery time — `peer_node_id_for_conn`'s documented fallback when no ACTIVE connection matches the frame's `conn_id`, plausible mid-teardown — `entry.remote_node_id == authenticated_peer` becomes `0 == 0` and wrongly admits the delivery. I added an explicit `authenticated_peer == 0` reject in `deliver_link_down_to_ref`, before that comparison, and corrected the doc comments that relied on the false invariant.

Added `link_down_with_zero_authenticated_peer_is_rejected_against_zero_remote_node_id`, which registers a link entry with `remote_node_id == 0` and delivers with `authenticated_peer == 0`. Revert-proved it: disabling the new guard reproduces the exact forged-delivery symptom, restoring it passes.

Full gate re-run on the new head: `cargo test -p hew-runtime --lib` (2023/2023), `cargo clippy -p hew-runtime --lib --tests -- -D warnings` (clean), `cargo fmt -p hew-runtime --check` (clean), `tests/vertical-slice/run.sh` (427/427).

The `CTRL_LINK_REQ`/`CTRL_UNLINK` path having no peer authentication at all is a separate, pre-existing gap (not introduced by this PR, and not fixed by this follow-up) — I'll track that on its own.

PR #2562, opened 2026-07-10
fix(std): normalize regex string-return no-match to canonical empty string

### LINKED ISSUE #2529: regex.Pattern.find returns a no-match empty string that compares != to a string literal ""
The regex string-returning externs (`hew_regex_find`, and likely the other string-returning regex methods) return an empty string on no-match that reports `.len() == 0` but does NOT compare `==` to a string literal `""`. Two runtime-empties compare equal to each other, but a runtime-empty `!= ""`.

This silently breaks the idiomatic no-match check:
```hew
if pat.find(s) == "" {
    // treated as "matched" even when there was no match
}
```
The user must currently write `if pat.find(s).len() == 0` to be correct.

Root cause: a null-vs-empty string-representation mismatch at the regex FFI string-return boundary — the C side returns an empty buffer whose Hew string representation isn't the canonical empty-string literal. This is independent of method dispatch (find returns the correct len-0 value; the by-value dispatch bug is fixed separately).

Fix direction: normalize the FFI string-return empty to the canonical empty-string representation at the boundary (so runtime-empty `== ""`), or return `Option<string>` for the no-match case. Applies to the whole regex string-return FFI seam, not just find.

gertybotbot:
Fixes #2529.

## Problem

`Pattern.find` is typed `-> string` and documented as returning "the matched substring, or an empty string if no match", but `hew_regex_find` returned a raw **null** on no-match. A Hew `""` literal is a non-null, `malloc`'d, zero-length buffer, and `hew_string_equals` treats a null and a non-null empty buffer as **unequal**:

```rust
cabi_guard!(a.is_null() && b.is_null(), 1);
cabi_guard!(a.is_null() || b.is_null(), 0);
```

…even though `hew_string_length`/`hew_string_is_empty` both treat null as empty. So the idiomatic no-match check silently misfired:

```hew
if pat.find(s) == "" {
    // NOT taken even when there was no match — len 0 but != ""
}
```

Users had to write `pat.find(s).len() == 0` to be correct.

## Fix

Normalize the no-match / error returns of the two `-> string` regex FFI entry points to the canonical empty string (`str_to_malloc("")`) at the boundary, so the returned value compares `== ""`:

- **`hew_regex_find`** — no-match, null handle, and non-UTF-8 text now return the canonical empty string instead of null.
- **`hew_regex_replace`** — null handle and non-UTF-8 input now return the canonical empty string instead of null (its substitution path already returned a real string).

The `Vec<string>`-based capture seam (`capture`/`capture_named`/`find_all`) is unchanged: it deliberately models `Option<string>` via length 0/1 and never returned a bare-null string, so it never had this defect.

## Tests

- `cargo test -p hew-std --lib` regex suite **18/18** — 2 new: `find_no_match_returns_canonical_empty_not_null` (no-match + null-handle both return a non-null zero-length buffer) and `replace_invalid_handle_returns_canonical_empty_not_null`. Updated the existing `test_regex_find` no-match assertion to the new contract.
- New runnable vertical-slice fixture `regex_find_no_match_empty_string` asserts end-to-end that a positive match still compares by value, a no-match compares `== ""` and `!=` the matched string with `len() == 0`, and replace hit/miss both return the expected non-null string.
- Reproduced before/after against a from-source build: the repro printed a "treated as match" outcome (len 0 but `!= ""`) on `main` and the correct no-match outcome after.
- `cargo fmt -p hew-std -- --check` clean; full `hew-std --lib` **652/652** single-threaded.

### DIFF
--- a/hew-std/src/regex.rs
+++ b/hew-std/src/regex.rs
@@ -68,8 +68,18 @@ pub unsafe extern "C" fn hew_regex_is_match(re: *const HewRegex, text: *const c_
 /// Find the first match of the compiled regex in `text`.
 ///
 /// Returns a `malloc`-allocated, NUL-terminated C string containing the
-/// matched substring, or null if no match is found. The caller must free
-/// the returned string with `libc::free`.
+/// matched substring, or the canonical empty string (a non-null, zero-length
+/// buffer) when there is no match. The caller must free the returned string
+/// with `libc::free`.
+///
+/// The user-facing signature is `fn find(..) -> string` (not `Option<string>`),
+/// documented as returning "the matched substring, or an empty string if no
+/// match". Returning a raw null on no-match violated that contract at the string
+/// equality boundary: `hew_string_length`/`hew_string_is_empty` treat null as
+/// empty, but `hew_string_equals(null, "")` returns 0 (null and a non-null empty
+/// buffer are unequal), so the idiomatic `pat.find(s) == ""` no-match check
+/// silently reported a match. We normalize no-match to the canonical empty
+/// string here at the FFI boundary so the returned value compares `== ""`.
 ///
 /// # Safety
 ///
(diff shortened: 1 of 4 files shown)


slepp:
Thanks for catching this -- the no-match-returns-null bug is real and worth fixing. While reviewing, I found the fix here also converts the FFI's error paths (invalid handle, invalid UTF-8 input) into the same empty-string sentinel, which masks real caller errors as valid results. I've picked this up in #2567, which normalizes only the no-match path and keeps null as the error signal for actual failures. Closing this in favor of that -- appreciate the report.

PR #2563, opened 2026-07-10
fix(types): reject reserved type names in declarations

### LINKED ISSUE #2520: checker: reserve primitive and structural type names
The checker accepts declarations named `i64` and `tuple`:

```hew
type tuple<T> { value: T; }
type i64 { value: i64; }
```

`./target/debug/hew check` reports this program as OK. The type-fragment encoder intentionally leaves primitive fragments unchanged, so these names should be rejected rather than treated as a mangling concern.

Expected: declarations cannot use primitive names or structural encoder heads such as `tuple`.

Observed: both declarations are accepted.

This was found while fixing #2485; it is intentionally tracked separately from that symbol-encoding fix.

gertybotbot:
Fixes #2520.

## Problem

The checker accepted type declarations that reuse the compiler's own type vocabulary:

```hew
type tuple<T> { value: T; }
type i64 { value: i64; }
```

`hew check` reported both as OK. The type-fragment encoder intentionally leaves primitive fragments unchanged, and structural fragments are emitted under fixed heads (`tuple$x...$g`, `array$x...$g`, `fn$x...$r...$g`, ...). A declaration named after a primitive spelling or a structural head is therefore indistinguishable from a mangled structural fragment at the symbol layer, so these names must be rejected at declaration time.

## Fix

- New `ty::is_reserved_type_name(name)`: the primitive aliases (minus the non-identifier `()`/`!` spellings, which can never be declared) plus the structural encoder heads (`tuple`, `array`, `slice`, `fn`, `closure`, `ptr`, `ptrmut`, `borrow`, `dyn`, `task`, `unit`, `never`, `typeparam`). The head list is kept in lockstep with `monomorph::mangle_resolved_ty` and `resolved_ty::mangle_resolved_ty_segment`.
- New `ReservedTypeName` diagnostic + `TypeError::reserved_type_name` constructor, with a stable `as_kind_str` key for the LSP/WASM protocol.
- The gate lives in the two shared namespace-registration chokepoints, `register_type_namespace_name` and `register_machine_type_namespace_names`, so `type` / `record` / `actor` / `machine` declarations are all rejected uniformly and **fail closed** (return `false`) before any registration side effects.

Behaviour:

```
error: `tuple` is a reserved type name and cannot be used for a declaration
error: `i64` is a reserved type name and cannot be used for a declaration
```

## No over-fire

Only exact-match bare names are reserved. Look-alikes (`Tuple`, `MyString`, `I64`, `String`, `Array`) still declare cleanly. Module-qualified names are unaffected (the gate keys on the bare declared name).

## Tests

- `ty` predicate unit test over primitives, structural heads, and near-miss user names.
- Checker rejection tests for `type i64`, `type tuple<T>`, and `record`/`actor` kinds.
- A no-over-fire test asserting look-alike names produce no `ReservedTypeName`.

## Verification

- `cargo test -p hew-types --lib`: 1654 passed, 0 failed.
- `cargo fmt --all -- --check` clean; `cargo clippy -p hew-types` clean.
- Full stdlib `.hew` corpus re-checked with the built compiler: zero reserved-name regressions (no std type is named after a reserved word).

### DIFF
--- a/hew-types/src/check/registration.rs
+++ b/hew-types/src/check/registration.rs
@@ -2347,6 +2347,11 @@ impl Checker {
         name: &str,
         span: &Span,
     ) -> bool {
+        if crate::ty::is_reserved_type_name(name) {
+            self.errors
+                .push(TypeError::reserved_type_name(span.clone(), name));
+            return false;
+        }
         let owner_key = (module_short.map(str::to_string), name.to_string());
         if let Some(prev_span) = self.type_namespace_owners.get(&owner_key).cloned() {
             self.report_duplicate_type_namespace_name(name, span, prev_span);
@@ -2379,6 +2384,11 @@ impl Checker {
         machine_name: &str,
         span: &Span,
     ) -> bool {
+        if crate::ty::is_reserved_type_name(machine_name) {
+            self.errors
+                .push(TypeError::reserved_type_name(span.clone(), machine_name));
+            return false;
+        }
         let machine_key = (module_short.map(str::to_string), machine_name.to_string());
         if let Some(prev_span) = self.type_namespace_owners.get(&machine_key).cloned() {
             self.report_duplicate_type_namespace_name(machine_name, span, prev_span);
--- a/hew-types/src/check/tests/basic.rs
+++ b/hew-types/src/check/tests/basic.rs
@@ -1426,3 +1426,84 @@ fn typecheck_local_result_enum_not_qualified_to_sqlite() {
         }
     );
 }
+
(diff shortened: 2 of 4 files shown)


slepp:
Thanks for catching this -- reserved type names being silently shadowed is a real gap. While reviewing, I found this covers primitive names but misses the separately-emitted CancellationToken leaf, and the tests don't require the diagnostic exactly (a rejected declaration can still cascade into unrelated errors). I've picked this up in #2566, which covers the full type-head surface and adds a compile-time guard so a future type variant can't silently slip through unreserved. Closing this in favor of that -- appreciate the report.

PR #2564, opened 2026-07-10
fix(types): give lambda parameters a shadow span so nested locals warn, not error

### LINKED ISSUE #2503: Nested-scope shadow of a lambda parameter still hard-errors instead of warning
PR #2499 fixed the parameter-shadowing severity bug (a nested \`let\` shadowing a parameter was indistinguishable from shadowing a synthetic binding, since both had \`shadow_span: None\`) for function, receive-fn, actor-init, and actor-hook parameters — but explicitly scoped to those four surfaces.

Lambda/closure parameters (\`|x: i64| { ... }\`, bound via plain \`self.env.define(...)\` at \`hew-types/src/check/expressions.rs:5945\`) are a 5th surface with the same underlying bug class and are NOT covered by that PR. Verified on the post-#2499 branch: a nested \`let\` shadowing a lambda's own parameter still hard-errors instead of warning.

Found while gating PR #2499 — confirmed this is pre-existing behavior, not a regression introduced by that PR.

Fix direction: same shape as #2499's \`define_param_with_span\` — give lambda parameters a real shadow span at the binding site in \`expressions.rs:5945\`.

gertybotbot:
Fixes #2503.

PR #2499 fixed the parameter-shadowing severity bug for function, receive-fn, actor-init, and actor-hook parameters by switching their binding from `env.define` (fully synthetic, no span) to `env.define_param_with_span` (keeps `def_span: None` so `UnusedVariable` stays exempt, but adds `shadow_span` so a nested local shadowing the parameter is classified as outer-scope-shadows-a-user-binding — a warning — instead of outer-scope-shadows-a-synthetic-binding — a hard error). That PR explicitly scoped the fix to those four surfaces and named lambda parameters as a known 5th surface with the same bug class, still bound via plain `env.define` in `hew-types/src/check/expressions.rs` (`check_lambda`).

`LambdaParam` had no name span at all (only the type annotation carries one, and it's optional), so closing this required a real AST field: `LambdaParam::name_span`, populated by both lambda-parameter parsers (`try_parse_lambda_params` and the pipe-lambda variant `try_parse_pipe_lambda_params`) by capturing `peek_span()` right before consuming the identifier token. `check_lambda` now calls `check_shadowing` (previously never invoked for lambda params at all) and binds through `define_param_with_span` using that span, matching the #2499 pattern exactly.

## Verification

- Reproduced the described defect directly against a from-source pre-fix build: `fn main() { let f = |x: i64| -> i64 { let x = 2; return x; }; ... }` hard-errors (`variable `x` shadows a binding in an outer scope`).
- Post-fix build downgrades it to the same outer-scope-shadowing *warning* the other four surfaces get, with the correct originating span pointing at the lambda parameter's name token.
- Same-scope duplicate lambda params (`|x: i64, x: i64| ...`) still correctly hard-error — `check_shadowing`'s same-scope branch is unaffected.
- Added 2 regression tests mirroring #2499's free-fn/receive-fn pair (warn on nested-scope lambda-param shadow; unused lambda param still exempt from `UnusedVariable`), plus updated the one hand-constructed `LambdaParam` test fixture for the new field.
- `cargo test -p hew-types --lib`: 1651/1651 passing.
- `cargo test -p hew-parser -p hew-hir -p hew-analysis --lib`: 357 + 249 + 89 all passing.
- `cargo fmt` clean, `cargo clippy -p hew-parser -p hew-types --lib -- -D warnings` clean.

### DIFF
--- a/hew-parser/src/ast.rs
+++ b/hew-parser/src/ast.rs
@@ -708,6 +708,15 @@ pub struct TimeoutClause {
 pub struct LambdaParam {
     pub name: String,
     pub ty: Option<Spanned<TypeExpr>>,
+    /// Source span of the parameter *name* token, used only for outer-scope
+    /// shadowing classification (see `Checker::check_lambda` /
+    /// `Env::define_param_with_span`). Free/receive-fn/init/hook parameters
+    /// already carry this via `Param.ty.1`; lambda parameters had no name
+    /// span at all prior to this field, so a nested `let` shadowing a
+    /// lambda parameter was indistinguishable from shadowing a fully
+    /// synthetic binding and hard-errored instead of warning.
+    #[serde(default)]
+    pub name_span: Span,
 }
 
 #[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
--- a/hew-parser/src/parser/expressions.rs
+++ b/hew-parser/src/parser/expressions.rs
@@ -1469,6 +1469,7 @@ impl Parser<'_> {
         let mut params = Vec::new();
 
         while !self.at_end() && self.peek() != Some(&Token::Pipe) {
+            let name_span = self.peek_span();
             let name = self.expect_ident()?;
 
             let ty = if self.eat(&Token::Colon) {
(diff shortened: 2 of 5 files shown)


slepp:
Thanks for catching this -- the shadow-severity bug is real. While reviewing, I found the serde default on the new span field would silently manufacture a fake location for old serialized AST data. I've picked this up in #2568, which drops the fallback so every lambda parameter carries a real parsed span. Closing this in favor of that -- appreciate the report.

PR #2583, opened 2026-07-11
fix(types): reject opaque handles as receive-fn parameters with a clean diagnostic

### LINKED ISSUE #2511: Opaque FFI handle as an actor receive-fn message parameter fails closed with a raw codegen error, not a clean diagnostic
Found while fixing a doc example for #2446's groundwork PR (#2510).

Passing an opaque FFI handle type (e.g. `net.Listener`, `net.Connection`) as a receive-fn message parameter fails at codegen, not at type-check:

```
E_CODEGEN_FRONT_FAIL_CLOSED: ... wire CBOR serialize: named type 'Listener' is not a registered record layout
```

Actor message payloads are CBOR-serialized for mailbox dispatch, and `#[opaque]` handle types have no record layout, so they can't cross the actor message boundary. The compiler correctly fails closed rather than emitting garbage — this isn't a memory-safety or correctness bug. But the error surfaces as a raw codegen-front message naming an internal wire-serialization detail, not a clear type-check diagnostic pointing at the parameter and explaining why opaque handles can't be message payloads.

The equivalent restriction for storing an opaque handle in actor *state* (a field) is already documented with a clearer story (see `examples/net/http_await_service.hew`'s header comment on the supervisor-restart clone gate). This is the same underlying restriction, just manifesting for a message *parameter* instead of a *field*, without the same diagnostic clarity.

Worth a clean, early type-check-time diagnostic: "opaque type `T` cannot be used as a receive-fn parameter (message payloads must be CBOR-serializable); open/construct it locally in the handler instead."

gertybotbot:
Fixes #2511.

## Problem

Passing an opaque FFI handle type (e.g. `net.Listener`, `net.Connection`) as an actor **receive-fn parameter** failed at codegen, not type-check:

```
E_CODEGEN_FRONT_FAIL_CLOSED: ... wire CBOR serialize: named type `Listener` is not a registered record layout
```

Actor message payloads are CBOR-serialized to cross the mailbox dispatch boundary, and `#[opaque]` handle types are pointer-shaped runtime resources with no record layout, so they cannot cross it. The compiler correctly failed closed (no memory-safety/correctness bug), but the error surfaced as a raw codegen-front message naming an internal wire-serialization detail — even at `hew check` — instead of a clear type-check diagnostic pointing at the parameter.

## Fix

`check_receive_fn` now walks each parameter type through the existing transitive `ty_field_contains_opaque` classifier (the same one that gates opaque-in-actor-state clones). On an opaque leaf reached directly or nested inside a record/enum/tuple, it emits a clean, early `OpaqueMessagePayload` (`E_OPAQUE_MESSAGE_PAYLOAD`) diagnostic:

```
error: opaque type `net.Listener` cannot be used as a receive-fn parameter;
       actor message payloads must be CBOR-serializable, and opaque handles
       have no serializable layout. Open or construct the handle locally
       inside the handler instead [E_OPAQUE_MESSAGE_PAYLOAD]
 6 |     receive fn handle(listener: net.Listener) {
   |                                 ^^^^^^^^^^^^
```

The diagnostic mirrors the existing `OpaqueDirectConstruct` / `E_OPAQUE_CONSTRUCT` precedent for constructing opaque handles.

## Scope

Precise and additive:
- Only **receive-fn** (message) parameters are affected. Plain `fn` parameters taking opaque handles, and serializable payloads, are untouched (verified by negative-control tests).
- `ty_field_contains_opaque` is promoted `fn` -> `pub(super)` to reuse it from the items module; no logic change.
- The nested-in-actor-state clone gate (a separate, already-clear error) is unchanged.

## Tests

5 new tests in `hew-types` (`opaque_receive_fn_param_rules`):
- user `#[opaque]` type as a direct receive-fn param -> rejected (exactly one error, carries envelope code + names the type)
- registered stdlib-style handle as a direct param -> rejected
- opaque nested in a record field param -> rejected (one clean error at the parameter)
- ordinary serializable params (`i64`, `string`) -> accepted (negative control)
- opaque handle as a non-receive `fn` param -> accepted (negative control)

## Verification

- `cargo test -p hew-types --lib`: 1662 passed, 0 failed
- `cargo fmt -p hew-types -- --check`: clean
- `cargo clippy --workspace --tests -- -D warnings`: clean
- Built `hew` from source; reproduced the pre-fix raw codegen error on `main`, confirmed the clean diagnostic post-fix
- All `examples/net/*.hew` still check as before (the one pre-existing `probe_opaque_actor_state.hew` actor-state clone-gate error is unrelated and unchanged)


slepp:
Thanks for the original diagnostic and the channel-handle exemption — this is a real, well-scoped fix for #2511. The corpus ratchet caught one more case worth fixing before merge: `LocalPid<T>`/`RemotePid<T>` (actor process references) also need the exemption, the same way `channel.Sender`/`channel.Receiver` do, since passing a PID as a message payload is routine in the actor model. I couldn't push the additional commit to this branch (maintainer_can_modify is off, and only the PR author can restore it), so I've carried the fix forward in #2617 with the LocalPid/RemotePid exemption plus a regression test for it, verified clean against the full example corpus. Closing this in favor of that.

Issue #2340, opened 2026-07-02
Exhaustiveness false-positive: Some((a, b)) + None on Option<(i64, i64)> rejected as non-exhaustive


slepp:
```hew
match x {
    Some((a, b)) => a + b,
    None => 0
}
```

with `x: Option<(i64, i64)>` is wrongly rejected as non-exhaustive, even though `Some((a, b))` is irrefutable over the tuple payload (both binders are plain identifiers) and `None` covers the remaining variant.

The exhaustiveness checker appears not to credit a tuple-destructure payload subpattern as covering its variant when the destructure itself is irrefutable — the same coverage-crediting seam as the misleading diagnostic for record-destructure payloads (`Wrapper::W(Pair { a, b })` reporting `missing W`), but here the shape is fully supported and the rejection is simply wrong: the user must write a redundant wildcard arm to compile.

Distinct from the pattern-cluster feature work (rest patterns, or-patterns in payloads) — this is a soundness bug in coverage accounting for an already-supported shape.

PR #2590, opened 2026-07-11
fix(hir): destructure tuple-typed struct-variant fields in match arms (#2354)

### LINKED ISSUE #2354: Struct-variant fields typed as tuples cannot be destructured in match arms (E_HIR: identifier has no binding)
Found while fixing #2340. Given:

```hew
enum Packet {
    Data { value: (i64, i64) };
    Empty;
}
```

a match arm `Data { value: (a, b) } => a + b` passes the type checker — the bindings resolve and (with the #2340 fix) exhaustiveness credits the arm correctly — but HIR lowering rejects it with `E_HIR: identifier has no binding in resolved HIR` for `a` and `b`. The failure is independent of exhaustiveness: adding a `_ => 0` catch-all produces the identical error.

The gap is specific to destructuring inside struct-variant field position: the match-arm payload-binding path in hew-hir/src/lower.rs (the payload-predicate extractor around :121-150) only extracts literal predicates for struct-variant fields and has no recursive path into a field's own tuple sub-pattern. The `let`-binding desugar (`lower_pattern_value_into_stmts`, ~:11108) already recurses correctly, and a plain binder on the same field (`Data { value: v } => v.0 + v.1`) lowers and runs fine.

Before the #2340 fix this shape was unreachable (the match was wrongly rejected as non-exhaustive first), so the lowering gap was latent. Now it fails closed but with a confusing internal-flavoured diagnostic for a shape the checker accepts. The fix belongs with the match-payload destructure work already planned for the pattern cluster: give the struct-variant field path the same recursive destructure lowering the tuple-variant path has.

gertybotbot:
## Summary

Fixes #2354. A struct-variant match arm whose field carries an aggregate sub-pattern — e.g. `Data { value: (a, b) }` — passed the type checker (bindings resolve, exhaustiveness credits the arm) but HIR lowering rejected the inner identifiers `a`/`b` with `E_HIR: identifier has no binding in resolved HIR`.

## Root cause

The checker's `record_arm_resolution` only records a `PayloadBinding` for a **bare** field binder (`binding_name_for_pattern` returns `None` for a tuple/record sub-pattern), delegating the nested destructure to lowering. The tuple-**variant** path (`Data((a, b))`) already handled this via `lower_constructor_payload_aggregates` — bind a temp to the payload slot, recursively lower the sub-pattern against it. But that helper is only invoked for `Pattern::Constructor` arms, and its field-type lookup returns empty for `HirVariantKind::Struct`. A struct-variant field with a tuple sub-pattern therefore materialised no binding for the inner leaves, so each inner identifier failed closed at HIR verification.

Before #2340 this shape was unreachable (the match was wrongly rejected non-exhaustive first), so the lowering gap was latent; now it fails closed with a confusing internal-flavoured diagnostic for a shape the checker accepts.

## Fix

Struct-variant analogue of the existing tuple-variant machinery, HIR-only:

- **`struct_variant_payload_aggregate_subpatterns`** — detects a `Pattern::Struct` arm carrying a non-empty tuple / record field sub-pattern; extends the `has_payload_aggregate_subpatterns` gate that opens the arm scope.
- **`lower_struct_variant_payload_aggregates`** — resolves the variant's declared struct fields (applying the scrutinee's concrete type args to enum type params for generic instantiations), binds a temp to the aggregate field by its declaration-order index (the same positional index MIR's `EnumLayout` flattening uses — see `HirVariant::field_tys`), and recursively lowers the field's tuple/record sub-pattern against that temp via `lower_pattern_value_into_stmts`.
- The Match-arm site invokes it for `Pattern::Struct` arms, mirroring the existing `Pattern::Constructor` call.

## Scope

- Only tuple and record field sub-patterns are newly lowered; bare field binders already flow through the checker's `payload_bindings` and are untouched.
- Record-**in**-field (`Move { to: Pt { x, y } }`) remains rejected at the checker's pre-existing v0.5 payload-subpattern gate — a separate limitation outside this issue (verified identical rejection on `main`).
- Guarded aggregate arms continue to hit the shared "guarded match arm with nested aggregate payload destructure" fail-closed path, unchanged.

## Verification

- New vertical-slice fixture `struct_variant_field_tuple_destructure` exits **7** (3 + 4). Load-bearing A/B: reverting only the arm wiring reproduces the exact pre-fix `E_HIR` unbound-identifier failure, confirming the fixture pins the defect.
- Also verified end-to-end: generic-enum struct-variant tuple field, wildcard-in-tuple (`(a, _)`), and mixed aggregate+bare fields (`{ pair: (a, b), tag: t }`) all bind and run.
- `cargo test -p hew-hir --lib` 89/89, `-p hew-mir --lib` 367/367, `-p hew-types --lib` 1660/1660; `cargo fmt`/`clippy` clean on `hew-hir`.

### DIFF
--- a/hew-hir/src/lower.rs
+++ b/hew-hir/src/lower.rs
@@ -239,6 +239,34 @@ fn constructor_payload_aggregate_subpatterns(pattern: &Pattern) -> bool {
     })
 }
 
+/// True when a struct-variant match arm (`V { field: <aggregate> }`) carries a
+/// field-position sub-pattern that itself destructures — a nested tuple or
+/// record. Mirrors [`constructor_payload_aggregate_subpatterns`] for the
+/// struct-variant field path.
+///
+/// The checker binds the inner leaves into the arm-body environment so the body
+/// type-checks, but `record_arm_resolution` only records a `PayloadBinding` for
+/// a bare field binder (`binding_name_for_pattern` returns `None` for an
+/// aggregate sub-pattern). Without this signal the HIR arm would materialise no
+/// bindings for those leaves and every inner identifier would fail closed with
+/// `E_HIR: identifier has no binding` (#2354). Detecting the shape here routes
+/// the arm through `lower_struct_variant_payload_aggregates`, which rebuilds the
+/// destructure directly from the AST pattern.
+fn struct_variant_payload_aggregate_subpatterns(pattern: &Pattern) -> bool {
+    let Pattern::Struct { fields, .. } = pattern else {
+        return false;
+    };
+    fields.iter().any(|field| {
+        matches!(
+            field.pattern.as_ref().map(|(sub_pat, _)| sub_pat),
+            Some(Pattern::Struct { .. })
+        ) || matches!(
+            field.pattern.as_ref().map(|(sub_pat, _)| sub_pat),
+            Some(Pattern::Tuple(items)) if !items.is_empty()
+        )
+    })
+}
+
 /// Resolve the type of an `if`/`else` expression from its two branch types,
 /// mirroring the checker's `unify_branches` Never-handling.
 ///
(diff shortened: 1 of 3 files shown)


slepp:
Thanks for this fix -- a later, more complete version of the same #2354 fix landed in #2609 (which handles nested tuple sub-patterns more thoroughly). Closing this one as superseded; the credit for identifying and fixing #2354 stands.

PR #2610, opened 2026-07-11
fix(types): clean diagnostic for opaque-handle receive-fn parameter (#2511)

### LINKED ISSUE #2511: Opaque FFI handle as an actor receive-fn message parameter fails closed with a raw codegen error, not a clean diagnostic

gertybotbot:
Fixes #2511.

## Problem

Passing an opaque FFI handle type (e.g. `net.Listener`, `net.Connection`) as an actor receive-fn message parameter failed at **codegen**, not type-check, surfacing as a raw internal message:

```
E_CODEGEN_FRONT_FAIL_CLOSED: wire CBOR serialize: named type `Listener` (key `Listener`) is not a registered record layout
```

Actor message payloads are CBOR-serialized for mailbox dispatch, but opaque handles are pointer-shaped runtime resources with no record layout, so they can't cross the actor message boundary. The compiler correctly fails closed — this isn't a memory-safety bug — but the error named an internal wire-serialization detail instead of pointing at the offending parameter.

## Fix

Reject opaque-typed receive-fn parameters at the checker with a new `E_OPAQUE_MESSAGE_PAYLOAD` diagnostic pointing at the parameter type span:

```
error: opaque type `net.Listener` cannot be used as receive-fn parameter `l`;
actor message payloads must be serializable, but opaque handles carry no record
layout — open or construct it locally in the handler instead [E_OPAQUE_MESSAGE_PAYLOAD]
 6 |     receive fn handle(l: Listener) {
   |                          ^^^^^^^^
```

- New `TypeErrorKind::OpaqueMessagePayload { param_name, opaque_name }` variant.
- The check reuses `ty_field_contains_opaque` (promoted to `pub(super)`), so it is **transitive**: a `Vec<Listener>` / `Option<Handle>` payload is caught at the opaque leaf, mirroring the existing actor-state clone-gate story referenced in the issue.
- Lives in a small `reject_opaque_receive_fn_param` helper to keep `check_receive_fn` under the `too_many_lines` clippy gate.
- Serializable payloads (primitives, records) are unaffected.

## Testing

- 3 new checker tests: user-`#[opaque]` param rejected (exactly one error, correct param/type names, envelope code), transitive `Vec<Handle>` rejected, non-opaque `i64`/`string`/record params clean.
- Full `hew-types` suite green (1663 lib + integration), `cargo fmt` + `cargo clippy --tests` clean.
- Repro verified from-source on `main`: opaque param now errors at the parameter span; normal params `OK`.

### DIFF
--- a/hew-types/src/check/items.rs
+++ b/hew-types/src/check/items.rs
@@ -1955,6 +1955,7 @@ impl Checker {
         for p in &rf.params {
             self.check_shadowing(&p.name, &p.ty.1);
             let ty = self.resolve_type_expr(&p.ty);
+            self.reject_opaque_receive_fn_param(&p.name, &ty, &p.ty.1);
             self.env
                 .define_param_with_span(p.name.clone(), ty, p.is_mutable, p.ty.1.clone());
         }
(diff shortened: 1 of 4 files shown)


slepp:
This is the same underlying fix as #2583 (also targeting #2511), which needed one more exemption (LocalPid/RemotePid actor references, alongside the channel Sender/Receiver case) to pass the full example corpus. That combined fix landed as #2617. Closing this as superseded, with credit for the original report.

PR #2619, opened 2026-07-11
fix(types): bind package-local bare type refs to their own module identity (#2208)

### LINKED ISSUE #2208: Package-local same-bare-name type-body resolution collision (pre-existing, surfaced during #2202)
## Problem
When two type declarations share the same bare (unqualified) name within package-local scope, member-type resolution can resolve to the wrong `type_defs` entry. This is independent of import aliases — a no-alias control exhibits the same collision — so it is a broader pre-existing bare-`type_defs` keying issue, not introduced by #2202.

## Discovery
Surfaced by the GPT cross-eco review of #2202 (import alias in type-declaration member positions). The reviewer constructed a package-local same-bare-name body counterexample; a no-alias variant of the same shape also fails, confirming the fault is in the general bare-name → `type_defs` resolution, not the alias re-resolution pass.

## Notes
- Qualified keys are the cross-module authority and resolve correctly; this is specifically the package-local bare-name path.
- Low priority: unobservable in correctly-disambiguated programs. Tracked separately so #2202 (member-position alias resolution) can land.

## Relationship
Follow-up from #2202.

gertybotbot:
## Summary

Fixes #2208. A bare type reference inside a **non-root module** — e.g. a `Holder { w: Wrap }` field whose `Wrap` is declared in the same package-local module — resolved through the `is_local` arm of `resolve_type_expr` to the **bare** `type_defs` key. That key is last-write-wins across modules that declare the same bare name (`pre_register_type_decl` documents the qualified alias as the authority), so when a sibling module declares its own `Wrap`, the field silently binds to whichever module registered `Wrap` last.

## Observability (correcting the issue note)

The issue notes this as "unobservable in correctly-disambiguated programs." It is in fact observable. Two package-local modules, each with a divergent `Wrap` and a `Holder { w: Wrap }`, mis-resolve one holder's field to the *other* module's layout:

```
error: no field `a_field` on type `Wrap`
  = help: b_field
```

…and the winner flips with import order (a program that reads `a.w.a_field` / `b.w.b_field` compiles or fails purely by which module's `Wrap` was registered last).

## Fix

In the `is_local` arm, bind to the current module's **own** qualified identity (`{short}.{name}`) when that key exists — mirroring the sibling `published_bare_type_qualified` branch that already qualifies specifically to dodge this bare-key collapse. `pre_register_type_decl` always seeds the qualified key alongside the bare one, so:

- non-colliding locals resolve identically to before,
- root-module references (`current_module_short() == None`) are untouched,
- only the cross-module collision case changes — from wrong-module to own-module.

## Boundary / neighbours checked

- Root program (`current_module == None`): unchanged, still binds bare.
- Single-module local reference: unchanged (qualified key == own def).
- The `published_bare_type_qualified` and `handle_matches` arms below are untouched.

## Tests

New load-bearing regression `same_bare_name_member_field_binds_to_own_module_identity` asserts each holder's `w` field carries its **own** module's `Wrap` identity (`alpha.Wrap` / `beta.Wrap`). It **fails on revert** (both collapse to the single last-writer identity), verified locally.

- hew-types: 1665 lib + all integration suites green (registry_core 281→282).
- hew-hir: full suite green (field identity flows into HIR/MIR).
- `cargo fmt` + `cargo clippy --tests`: clean.
- e2e: the two-module repro (both import orders) now prints `42` / `hello` and exits 0.

### DIFF
--- a/hew-types/src/check/resolution.rs
+++ b/hew-types/src/check/resolution.rs
@@ -1638,7 +1638,32 @@ impl Checker {
                     return Ty::Error;
                 }
                 let resolved_name = if is_local && self.type_defs.contains_key(name) {
-                    name.clone()
+                    // A bare reference to a locally-defined type normally binds
+                    // to the bare `type_defs` key. But when checking a non-root
+                    // module, the bare key is last-write-wins across modules that
+                    // declare the same bare name (`pre_register_type_decl`
+                    // documents this: the qualified alias is the authority). If a
+                    // sibling module later overwrites the bare `Wrap` entry, a
+                    // bare local `Wrap` here would resolve to the WRONG module's
+                    // fields downstream (observable as `no field ... help: <other
+                    // module's field>` — #2208). Bind to this module's OWN
+                    // qualified identity (`{short}.{name}`) when it exists so the
+                    // local reference is immune to the cross-module bare-key
+                    // race, mirroring the `published_bare_type_qualified` branch
+                    // below. `pre_register_type_decl` always seeds the qualified
+                    // key alongside the bare one, so non-colliding locals resolve
+                    // identically to their own def.
+                    match self.current_module_short() {
+                        Some(short) => {
+                            let qualified = format!("{short}.{name}");
+                            if self.type_defs.contains_key(&qualified) {
+                                qualified
+                            } else {
+                                name.clone()
+                            }
+                        }
+                        None => name.clone(),
+                    }
                 } else if let Some(qualified) = self.published_bare_type_qualified(name) {
                     // Exactly one module published this bare name. Bind to that
                     // owner's QUALIFIED identity (`owner.Name`) so a sibling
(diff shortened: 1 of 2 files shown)


slepp:
Thank you for the #2208 report and the core own-qualified binding fix. I preserved your commit and authorship in #2626, which adds the private nested-module visibility fix and the full corpus proving needed to complete this change. Superseded by #2626.

slepp:
Thanks for this — the #2208 collision fix is correct in intent and the observability correction is right. Review found one boundary hole: the own-qualified binding trips the visibility gate for same-module references to `private` types in nested modules (it broke std/net/tls), because the gate compares the short decl-module against the full accessor path. I've carried your fix forward in #2626 with that boundary closed and a regression covering the private-local-in-nested-module case (which this PR's pub-only test didn't exercise). Closing in favour of #2626, with credit for the original fix and the sharper issue diagnosis.

PR #2627, opened 2026-07-11
fix(distributed): validate node identity and setup failures

slepp:
## Summary

I made the connection handshake NodeId authoritative across routing and SWIM membership, reject reserved, colliding, and explicitly mismatched peer identities, and removed placeholder membership for bare-address dials.

I also made cross-node monitor and link setup fail closed with typed errors. `monitor(RemotePid)` now returns `Result<MonitorRef, MonitorError>`, failed registrations are reclaimed immediately, and codegen preserves the exact runtime error variant instead of manufacturing a handle or reporting `Ok(())`.

I extended the compiled two-process fixture to prove exact node/location/incarnation identity, ALIVE membership, typed monitor setup failure after route loss, and the existing cross-node monitor, link, partition, gossip, stale-reference, and wire behavior.

## Validation

- `cargo test -p hew-runtime -- --test-threads=1`
- `cargo test -p hew-cli --test distributed_two_process_e2e -- --test-threads=1`
- `cargo check --workspace`
- `make hew-check-all`

### DIFF
--- a/hew-cli/tests/distributed_two_process_e2e.rs
+++ b/hew-cli/tests/distributed_two_process_e2e.rs
@@ -629,6 +629,21 @@ fn remote_monitor_down_on_connection_drop() {
     );
 }
 
+/// Cross-node monitor setup fails honestly after route teardown. The runtime
+/// must return `Err(MonitorError::Partition)`, never an inert `MonitorRef`.
+#[test]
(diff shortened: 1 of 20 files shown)


slepp:
Superseded by #2646, which preserves explicit peer identity across reconnects and closes the reconnect admission gap.

PR #2628, opened 2026-07-11
Retain string values when ownership is shared

slepp:
## Summary

- I added a MIR-driven `StringRetain` marker so genuine string co-owner mints retain through `hew_string_clone`, and matching drops are admitted from the same ownership decision.
- Single-owner last-use handoffs and consuming runtime, actor, fork, and explicit-release paths are preserved without unnecessary retains.
- Coverage spans aggregate stores and reads, duplicate fields, local and parameter shares, returned values, and escaping closure captures.
- Extern strings from package imports are now classified by provenance: the HIR-to-MIR classifier is total and fail-closed for invalid module provenance, so dropping an imported extern string no longer aborts through `hew_string_drop`.
- I corrected `hew_vec_set_str` ownership metadata to match its copy-in behaviour.

This closes the remaining string ownership gap: uncertain sole ownership now becomes counted sharing instead of leaking or rejecting the value.

## Verification

- `cargo nextest run --workspace --no-fail-fast` — green across the workspace
- `string_retain_share_leak_oracle` — 3/3 with real `leaks(1)` measurements, flat at 0→0 nodes
- New owned-payload channel-send leak oracle — 3/3, per-frame slope 0 (LOW=3/HIGH=50 frames; a 1-leak/frame bug would read as +47 nodes), including under the poisoned-allocator triple (MallocScribble/MallocGuardEdges/MallocPreScribble)
- Linux ASan gate — 9/9 fixtures passed with no ASan/LSan findings, including `string_cow_retain_share` and the bytes ownership fixture
- Checked-MIR goldens regenerated and mechanically verified — `scripts/checked-mir-corpus.sh verify` clean, 40 fixtures x 2 stages byte-identical
- `make test-pkg-import` — passes with MallocScribble and MallocGuardEdges
- `make test-vertical-slice` — passed after moving newly supported string-sharing fixtures into the accept corpus

## Out of scope

- No changes to the bytes ownership model beyond exercising its existing fixture under the ASan gate.
- No wire-format or serialization changes; this is confined to string ownership classification and retain/drop emission.


slepp:
Blocked on an extern-string ABI mismatch discovered by Linux package-import CI: imported package functions may return plain malloc C strings, while the retain/drop implementation assumes a Hew refcount header. The full Rust workspace and vertical-slice gates pass; the cross-module trait-method fixture aborts when the headerless result reaches hew_string_drop. I stopped rather than expanding this PR into an unplanned extern ABI migration.

PR #2629, opened 2026-07-11
test(runtime): de-flake destroy_wakes_blocked_recv (accept both wake outcomes)

### LINKED ISSUE #1945: Flaky test: destroy_wakes_blocked_recv races on slower runners (InvalidConn vs Partitioned)
`hew-runtime::sim_transport::tests::destroy_wakes_blocked_recv` (`hew-runtime/src/sim_transport.rs:1750`) is timing-flaky on slower/contended runners — it failed on macOS arm64 (`expected Partitioned wake from destroy, got InvalidConn`, 9470/9471 otherwise passing).

The test spawns a recv thread, sleeps only 20ms, then calls `sim_transport_free` expecting the recv loop to be *parked* so it observes the partition flag (`Partitioned`). On a slower runner the recv loop instead observes the conn table already cleared first and returns `InvalidConn`. Both outcomes legitimately prove the recv woke from `sim_transport_free` — the assertion is too strict about which one wins the race.

Fix options:
- Accept both outcomes: `matches!(outcome, SimRecvOutcome::Partitioned | SimRecvOutcome::InvalidConn)` — both prove the wake.
- Or add a synchronization point so the recv is provably parked before `sim_transport_free` (no sleep-based ordering).

This is the same macOS contention class as #1924 but a distinct test (#1924 cites `eval_wasm_file_runtime_failure...`). De-flake rather than blanket-retry. Milestone: v0.5.5.

gertybotbot:
Fixes #1945.

## Problem
`sim_transport::tests::destroy_wakes_blocked_recv` asserted the parked recv observed exactly `SimRecvOutcome::Partitioned` after `sim_transport_free`. But `shutdown()` sets the partition flag **and** clears the conn table before its `notify_all`, so a woken recv loop legitimately races between two views:

- the top-of-loop `st.partition` check surfaces `Partitioned`;
- re-taking the table lock over the now-empty `conns` surfaces `InvalidConn`.

Both prove the recv woke from the destroy rather than its own 5 ms watchdog timeout, so which wins is benign. On slower/contended runners the conn-table-cleared view wins first and the old assert flaked (observed on macOS arm64: `expected Partitioned, got InvalidConn`).

## Fix
Widen the assert to `Partitioned | InvalidConn` (maintainer's option 1). Any non-wake outcome (`PeerClosed`/`Frame`/`BufferTooSmall`) still fails, so the test keeps proving the destroy-wake it was written to guard — it is not blanket-retried or made unconditional.

## Verification
- 300 runs (200 serial + 100 under full-core `yes` saturation), zero failures.
- Full `hew-runtime` `sim_transport` suite green (16/16).
- fmt + clippy(--tests) clean.

### DIFF
--- a/hew-runtime/src/sim_transport.rs
+++ b/hew-runtime/src/sim_transport.rs
@@ -1744,12 +1744,22 @@ mod tests {
                 thread::sleep(Duration::from_millis(5));
             }
             let outcome = recv_handle.join().expect("recv thread panicked");
-            // After destroy the partition flag is set and the conn
-            // table is cleared; whichever the recv loop sees first
-            // surfaces as Partitioned.
+            // `shutdown()` sets the partition flag AND clears the conn
+            // table (then `notify_all`s) — a woken recv loop can legally
+            // observe either first: the partition check at the top of the
+            // loop surfaces `Partitioned`, while re-taking the table lock
+            // after the cleared conns surfaces `InvalidConn`. Both prove
+            // the recv woke from `sim_transport_free` rather than its own
+            // watchdog timeout; the race over which wins is benign, so
+            // accept both instead of asserting a fixed winner (this is
+            // the de-flake per #1945 — the old `Partitioned`-only assert
+            // was timing-flaky on slower/contended runners).
             assert!(
-                matches!(outcome, SimRecvOutcome::Partitioned),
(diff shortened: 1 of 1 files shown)


slepp:
Superseded by #2643, which proves the receiver was parked before destroy and wakes through the shutdown notification path.

PR #2816, opened 2026-07-26
test: run every ignored test or delete it

slepp:
An `#[ignore]`d test does not run, while its name and its file keep promising coverage. There were 21. Each is now either wired to something that executes it, or deleted.

Every `#[ignore]` that remains names both the make target and the CI step that invokes it, so the claim is checkable rather than aspirational:

```rust
#[ignore = "run by `make libhew-link-race-test`, invoked by the \"Run libhew bootstrap-race proof\" step in .github/workflows/ci.yml"]
```

## What was hiding behind the skips

**`make libhew-link-race-test` was reached by nothing** — not CI, not a preflight lane, not `make test`. Two multi-process proofs of the `libhew.a` bootstrap race were dead code. Now wired into CI.

**A test pointed at a fixture that was never committed.** `duplex_pair_send_skeleton.hew` has no git history at all, while the ignore reason described a compiler gap. Repointed at a real fixture.

**The QUIC round-trip blocker was fixed months ago.** The test passes — coverage that had been sitting behind a stale note.

**An MCJIT test works.** `machine_emit_exec` maps symbols by address instead of trusting the resolver, which is exactly why its 21 siblings SIGSEGV. Its ignore reason had been inherited from them and never measured. It now runs; the 21 that genuinely cannot work are deleted rather than left as a promise of coverage.

**One test was unrunnable by construction.** `run_timeout_kills_grandchild_process_tree` passes `--timeout 30` inside a harness with a 30 s deadline, so its assertion could never evaluate. This had been attributed to codegen.

**A comment claimed coverage that did not exist.** An ignored trait-coherence fail-open documented itself as "CI-visible via `cargo test -- --ignored`". `--ignored` appears twice in the repo, both crate-pinned elsewhere; nothing ran it. The test now asserts the fail-open we actually ship and fails when it is fixed. The claim is deleted.

## Where the stated blocker was real

Those tests now assert the behaviour we ship, so the gate fails forward when the defect is repaired rather than staying silent while it persists.

Investigating one widened a known bug: trap exit codes collapse to 1 on **native** as well as WASM. `hew run` prints `DivideByZero` and exits 1, while overflow correctly reports 201. The ignore reason had described this as WASM-only.

Un-ignoring the WASI-dependent test also exposed that Windows CI has no `wasmtime` — Linux installs it from a tarball and macOS via brew, but the Windows job never got an equivalent, and excluding `hew-wasm` does not help a test that lives in `hew-cli`. Installed there too, pinned to the same resolved version as the other platforms so the jobs cannot drift apart.

Nothing was re-ignored and no finding was deferred to an issue. One test was cut rather than kept: a handle-registration timing test asserting linear-not-quadratic scaling, because the neighbouring op-count test asserts the same lazy-fixpoint property deterministically instead of by wall-clock. Twenty-one MCJIT tests were also deleted — all of them already `#[ignore]`d, so nothing that had been executing stopped.


slepp:
Closing without merging — superseded rather than abandoned.

#2811 merged first and deleted 22 ignored tests in bulk, including several this branch had measured as working: the `machine_emit` MCJIT test (5/5), the QUIC service round-trip (4/4 once its bind probe is fixed), and some `eval_wasm_*` tests whose stated blockers were stale. That is my sequencing error — two changes on the same surface, and the coarser one landed first.

Rebasing this onto that produces conflicts in eight files where every resolution is a judgement about whether a deleted test should come back. Resolving them in a rebase is the wrong shape: a mistaken pick silently loses coverage with no reviewer able to see it.

Re-deriving the parts that stand on their own onto current main, with each restored test verified to pass rather than carried over on the strength of an earlier measurement: the working tests restored un-ignored, the QUIC wildcard-bind probe fix, wasmtime on Windows pinned to the same version as the other platforms, and the deletion of `try_require_wasi_runner`, which let tests pass while asserting nothing.

PR #2822, opened 2026-07-26
build: verify the archive identity once per archive, not once per link

slepp:
The build-identity check I added in #2812 must read the whole archive rather than stopping at the first stamp, because refusing an archive that carries two different identities requires knowing whether a second exists. On the 160 MB debug archive that is ~261 ms, and it ran on every resolution.

`tests/vertical-slice/run.sh` invokes the driver as a fresh process per case, so "once per resolution" and "once per link" are the same number, and the cost scaled with the number of programs linked. `Build & test (Linux)` on `main` went from 66 minutes to the 90-minute job ceiling, where it now cancels — taking every open PR with it, since a cancelled required check does not re-trigger itself.

## What changed

The verification answer is a property of the archive **file**, not of the link, so it is computed once and reused. The memo is keyed on the filesystem facts that pin down which bytes live at a path: length, mtime, ctime, device and inode.

Each field is there for a reason:

- **length and mtime** catch every ordinary rebuild — Cargo writes a new archive, `cp` writes a new archive, both move the mtime.
- **device and inode** catch replacement by rename or copy-onto-a-new-file, which is how Cargo and every install step actually put an archive in place: the path is the same and the file behind it is not.
- **ctime** catches the one case the others could be talked out of — an in-place rewrite that restores the old length and back-dates the mtime with `touch`. The kernel stamps ctime on any change to the inode or its contents and userspace cannot move it backwards, so a forged mtime does not buy a memo hit.

Off Unix there is no ctime and creation time stands in. That is weaker against deliberate back-dating and fully adequate against the failure this check exists for, which is an ordinary stale build artefact.

`make test-vertical-slice` goes from 4m26s to 2m9s.

## The guarantee is unchanged

Two different stamps in one archive is still an unconditional refusal; a missing, unreadable or unparseable stamp is still a refusal; and a replaced archive re-verifies rather than serving a cached pass.

That last one is the risk in caching a safety check, so it is the one to demonstrate. With a warm memo, replacing the archive with stamp-free bytes is refused before the linker sees it:

```
Error: libhew.a does not match this compiler — refusing to link.
  archive               : …/target/debug/libhew.a
  archive build identity: missing — no `HEW_BUILD_IDENTITY_V1=` stamp
  driver  build identity: bc890b6227bec6595438f8a10f845fe7096e29e2b41491873665c90a254e1a88
```

## Where the check belongs

It stays exactly where it is. The check already lives inside archive resolution rather than at the call site, so there is no per-link check to hoist. The axis was never the call site — it was process lifetime.

It should not move earlier either. A build-system step is the position `check-libhew-fresh.sh` already occupies, and its own header explains why that is not sufficient alone: the driver's check runs at the moment of link and therefore also covers direct `cargo build`, IDE builds and copied-in artefacts. Enforcing this anywhere but on the path to the link line reintroduces the holes #2812 closed.

## Remaining cost

The first link after any rebuild still pays the full scan. That is one-shot per archive build and does not scale with link count, so it is not addressed here — but if the debug archive grows substantially, the right answer is to attack the constant directly rather than to weaken the memo key.

### DIFF
--- a/hew-cli/src/build_identity.rs
+++ b/hew-cli/src/build_identity.rs
@@ -18,10 +18,30 @@
 //!
 //! Fail closed means fail closed: a missing, truncated or unreadable stamp is
 //! refused, never accepted as a match.
+//!
+//! # Cost
+//!
+//! Establishing the answer means reading the whole archive: refusing an archive
+//! that carries two different identities requires knowing whether a second one
+//! exists, so there is no early exit. On a 160 MB debug `libhew.a` that is
+//! roughly a quarter of a second, and the vertical slice and the ratchets link
+//! thousands of programs.
+//!
+//! The answer, though, is a property of the archive *file*. It cannot change
+//! between two links unless the file changes. So the scan is memoized on a
+//! fingerprint of the file itself (see [`ArchiveFingerprint`]) at two levels:
+//! an in-process map, so one driver process linking N programs scans once, and
+//! a marker file beside the archive, so the next `hew` process does not scan at
+//! all. Both are keyed strictly enough that a rebuilt or replaced archive
+//! misses and is scanned again, and both only ever record a *success* — the
+//! memo can assert "this exact file matched this exact driver" and nothing
+//! else. Anything the memo cannot vouch for falls through to the full scan.
 
 use std::collections::BTreeSet;
-use std::io::Read;
-use std::path::Path;
+use std::collections::HashMap;
+use std::io::{Read, Write};
+use std::path::{Path, PathBuf};
+use std::sync::{Mutex, OnceLock};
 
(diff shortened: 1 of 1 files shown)


slepp:
Closing in favour of removing the check rather than making it cheaper.

The per-link identity verification was the wrong shape: it polices what the build system should guarantee. The forward-engineered fix is already in #2812 — `libhew.a` is a real file target with source prerequisites, `make hew` is an alias of `hew-native` so a driver-only build cannot be produced, and the freshness check is an order-only prerequisite of its consumers. The graph rebuilds the archive because the graph says so.

The SHA stamp on top of that only adds coverage for direct `cargo build`, IDE builds and copied artefacts — cases better handled by using the build system than by checking up on it. It cost 66 min → the 90-minute CI ceiling and blocked the queue.

Removing the stamp and its verification; keeping the build-graph correctness.

PR #2866, opened 2026-08-08
feat(runtime): add HashMap.entries() -> Vec<(K, V)>

### LINKED ISSUE #2414: Add HashMap.entries() -> Vec<(K, V)>
`HashMap` has `keys() -> Vec<K>` and `values() -> Vec<V>` but no way to snapshot both sides in one pass. I want `entries() -> Vec<(K, V)>`: an eager snapshot of the occupied slots as tuples, with owned K/V cloned into the new vector (the snapshot owns independent copies, matching how `keys`/`values` clone their elements).

I attempted this once before and reverted it: the hard part is that `Vec<(K, V)>` is a tuple-element vector whose element can own heap (e.g. `HashMap<string, string>` gives `Vec<(string, string)>`), which needs the per-tuple layout descriptor plus the composite clone/drop thunk discipline — not the scalar/string element paths `keys`/`values` use.

Sketch for the next attempt:
- Build the tuple vec through the same layout-descriptor path the `hew_vec_*_layout`/owned family uses: codegen already synthesizes tuple layout descriptors with composite drop thunks, so the runtime kernel should receive the pair layout (and the V field offset, so the runtime never re-derives tuple layout rules) rather than inventing its own walk.
- Runtime kernel shape: `hew_hashmap_entries_layout(m, pair_layout, v_offset) -> *mut HewVec` — for each occupied slot, clone K into a scratch pair blob at offset 0 and V at `v_offset` via the map's own K/V descriptors, then `hew_vec_push_owned_move` the blob (move — no second clone, so each cloned pair has exactly one owner).
- Checker: an `entries` row next to `keys`/`values` with a `VecOfPair` return template instantiated as `Vec<(K, V)>`.
- Proof bar: order-agnostic set-equality with the inserted pairs at exact values, owned K/V leak-free under the poisoned-allocator oracle, and entries-then-drop-map with no double-free.

This is the remaining additive slice of the collection remove/sentinel reshape (the `Vec.remove -> T`, `HashMap.remove -> Option<V>`, and sentinel-to-Option changes); it is independent of those and lands on its own.

gertybotbot:
Closes #2414.

Implements the remaining additive slice of the collection remove/sentinel reshape:
`HashMap.entries()` returning an owned `Vec<(K, V)>`.

## Implementation

Wiring spans the full stack:

- **Checker** — dispatch, admissibility, and method resolution for `entries()`.
- **MIR** — lowering facts and runtime symbol registration.
- **Codegen** — layout, llvm, and runtime ABI.
- **Runtime** — `hew_hashmap_entries_layout` clones each key/value pair into a tuple
  element and transfers ownership into the result vector.

## Proof bar

The issue asked for three things; each is covered:

1. **Execution coverage** — `tests/hew/hashmap_entries.hew` plus additions to
   `hew-codegen-rs/tests/exec/hashmap_keys_values_exec.rs`.
2. **Owned-allocation balance** — `hew-runtime/tests/hashmap_entries_owned_balance.rs`
   asserts `allocations == frees` across map ingress, the entries clone, map free,
   and vec free.
3. **Leak detection under ASan** — `tests/vertical-slice/accept/hashmap_entries_asan.hew`
   wired into `scripts/asan-fixture-check.sh`.

### The balance test is load-bearing (verified by actual revert)

Reverting the ownership transfer so an extra owned clone escapes produces
**6 allocations against 4 frees** and fails the assertion:

```
assertion `left == right` failed: every owned entries allocation must be released
  left: 4
 right: 6
```

Scope note for reviewers: those counters instrument the test's own owned payloads,
so the test targets **element-ownership** balance. A leak of the runtime's internal
scratch buffer is outside its window — that case is what the ASan fixture covers.

## Verification

- `cargo test -p hew-runtime --lib` — 2240 passed, run 3x.
- `cargo test -p hew-types` — all suites green (1748 + 339 + 282 + 232 + 125 + 72 + 22 + 8).
- Rebased on `main` at the time of submission.

One flaky failure was observed in
`transport::tests::framed_send_to_broken_pipe_fails_closed_without_signal` (a
broken-pipe/signal-timing test unrelated to this change). It was **not** reproducible:
the full lib suite passes 2240/2240 on this branch across three consecutive runs, and
the test passes in isolation on both this branch and the pristine base.


slepp:
Thanks — this is a well-built change. The ownership story is the part that matters most and you got it right: clone into scratch, move into the Vec, deallocate without dropping. I checked it under `leaks` with dynamically built strings across bound, returned, iterated and argument-position shapes and found nothing, and your Rust balance test genuinely goes red when I break the value clone, which is exactly what I want from a test like that.

One thing worth knowing: the `.hew` fixtures all use string literals, and literals are never freed, so they can't actually observe an element-ownership bug. When I made the runtime alias the value instead of cloning it, every one of them still passed — including the ASan fixture and the test named for double-free. A version building its strings by concatenation aborted on a corrupted header immediately. I'm merging and tightening those fixtures in a follow-up so they earn the claim.

Separately, and not your problem: `m.entries().len()` on a non-empty map leaks the whole projection Vec. `m.keys().len()` does the same on main, so it's an existing gap in how projection temporaries get dropped — filing an issue. And now that `entries()` exists it's the obvious substrate for the `keys()`/`values()` zip that `for (k, v) in m` desugars to.

Issue #2828, opened 2026-07-27
mir: hew_process_last_error is audited fresh but plans zero releases


slepp:
`hew_process_last_error() -> string` is audited as returning a fresh owner, and the drop plan mints no release for it. The string leaks on `std/process.hew`'s error path.

Found while closing the ownership-authority seams in #2820, and deliberately left open there — the fix sits exactly on the fault line that work exists to hold.

## Why it is not a one-line fix

Closing it means minting a caller-side owner for a **pointer-bearing extern result**. That is the same shape as the fail-open #2820 closes, on the opposite polarity:

- Mint nothing, and the string leaks — where we are now.
- Mint an owner without establishing that the value is genuinely the caller's to release, and a double free becomes reachable. That is the defect the whole authority was built to make impossible.

The authority currently answers this by refusing to mint, which is the safe direction and the leaking one. Making it mint requires establishing that the returned pointer carries a transferable release obligation — which is a statement about the FFI contract of `hew_process_last_error`, not about the call site.

## What would settle it

Whether the extern's contract actually transfers ownership of the returned buffer to the caller, expressed somewhere the authority can read rather than inferred at the mint. If it does, the mint is licensed and the leak closes. If it does not — the runtime owns the buffer and hands out a borrow — then refusing to mint is correct and the leak is in the *runtime's* retention, not in the compiler.

Answering that question is the work here. Minting on the strength of "it is audited fresh" would be reasoning from the same kind of proxy that produced the seams #2820 spent ten rounds closing.

## Related

- #2820 — the authority and its seam closures
- #2821 — the by-value parameter trap surviving one `Option` deep

PR #2868, opened 2026-08-08
docs: state what the retention oracles cannot see

### LINKED ISSUE #2839: docs: state what the retention oracles cannot see
The `result-retention` axis added in #2833 establishes per symbol whether the
runtime transfers a returned buffer or lends it, and each row names the oracle
that established it. The oracles are black-box: two live results occupy distinct
addresses, the caller's handle is sole owner, and the slot survives the caller's
release.

Those three together do not catch one shape — a callee that allocates fresh,
hands the caller a genuinely sole-owned buffer, and separately stashes a raw
non-refcounted pointer to it that it frees at shutdown. Every probe passes and
the caller's release is still correct at the call site, but the buffer is freed
twice at process teardown.

Nothing in the family does this today; the TOML rows were corroborated against
the implementation sites. But the axis documents its rows as measured, and a
reader adding the sixteenth symbol should know exactly what "measured" covers
and what it does not, rather than inferring that a passing oracle is a proof of
sole ownership for all time.

One paragraph in the axis header in `scripts/jit-symbol-classification.toml`,
next to the existing documentation of the axis.

gertybotbot:
Closes #2839.

The `result-retention` axis documents its rows as *measured* and names the
oracles that establish them (`hew-runtime/tests/last_error_result_retention.rs`,
`hew-std/src/last_error_retention.rs`, #2828). Those oracles are black-box
probes over live results: distinct addresses for two live results, sole
ownership of the caller's handle at handoff, and callee state surviving the
caller's release.

Those three together do not catch one shape — a callee that allocates fresh,
hands the caller a genuinely sole-owned buffer, and separately stashes a raw
non-refcounted pointer to it that it frees at process teardown. Every probe
passes and the caller's release is still correct at the call site, but the
buffer is freed twice at shutdown.

Nothing in the family does this today and the rows were corroborated against
their implementation sites, so this is a documentation change only: it states
the boundary so a reader adding the sixteenth symbol knows what "measured"
covers, rather than inferring that a passing oracle proves sole ownership for
all time. It also names the concrete extra step (read the callee's teardown
path, not just run the oracle).

## Scope

One paragraph in the axis header of `scripts/jit-symbol-classification.toml`,
placed immediately after the existing sentence that names the oracles — the
location the issue asked for.

## Verification

- Comment-only: +13 lines, no contract row touched.
- `tomllib` parse clean; all 6 top-level tables present and all 494
  `[[ownership.contracts]]` entries intact after the edit.

### DIFF
--- a/scripts/jit-symbol-classification.toml
+++ b/scripts/jit-symbol-classification.toml
@@ -49,6 +49,19 @@
 # hew-runtime/tests/last_error_result_retention.rs and
 # hew-std/src/last_error_retention.rs (hew-lang/hew#2828).
 #
+# What those oracles CANNOT see: they are black-box probes over live results,
+# so they establish that two results occupy distinct addresses, that the
+# caller's handle is the sole owner at handoff, and that the callee's own
+# state survives the caller's release. Together those still miss one shape — a
+# callee that allocates fresh, hands back a genuinely sole-owned buffer, and
+# separately stashes a raw non-refcounted pointer to it that it frees at
+# process teardown. Every probe passes, the caller's release is correct at the
+# call site, and the buffer is freed twice at shutdown. No symbol in this file
+# does that today; the rows below were corroborated against their
+# implementation sites. But "measured" here means measured by those oracles,
+# not proved sole-owned for all time: anyone adding a row must read the
+# callee's teardown path as well as running the oracle.
+#
 # Three-tier model (rationale: Option A chosen over Option B for substrate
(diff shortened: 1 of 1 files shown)


slepp:
Checked the added paragraph against a few of the `transferred`-tier implementations and it holds — `hew_toml_last_error` allocates fresh with nothing stashed, exactly as described. Useful for the next person adding a contract row. Merged.

Issue #1451, opened 2026-04-21
WASM-TODO audit sweep across 10 feat(...) PRs missing parity markers


slepp:
## Summary

Ten recent `feat(...)` PRs introduced runtime features without either a WASM implementation or the `// WASM-TODO: <reason>` marker required by CLAUDE.md §4.

## Affected PRs

| PR | Title |
|---|---|
| #1225 | feat(runtime): PoisonSafe wrappers + migrate LINK_TABLE and ENV_LOCK |
| #1246 | feat(profiler): surface Hew actor type in pprof and observe output |
| #1257 | feat(bridge): record handler name in WASM metadata registry |
| #1266 | feat(stdlib): structured IoError replaces string-matched classification |
| #1293 | feat(observe): thread actor_type_id through MLIR emission, trace events |
| #1303 | feat(runtime): poison-safe wrappers for actor/monitor/node globals |
| #1305 | feat(runtime): poison-safe wrappers for TOP_LEVEL_SUPERVISORS + transport |
| #1306 | feat(runtime): poison-safe wrappers for scheduler and connection-manager |
| #1322 | feat(runtime): add per-TCP metrics counters |
| #1323 | feat(std/net/websocket): tie reader thread to actor/conn liveness |

## Proposed fix

Single sweep PR that, for each relevant function/struct introduced in these PRs:
- If the feature has a WASM analogue that was NOT implemented, add `// WASM-TODO: <brief reason + tracking issue>`.
- If the feature is genuinely native-only (e.g. poison-safe wrappers for state that doesn't exist in the single-threaded WASM model), add a short `// native-only: <reason>` comment for future readers.

## Risk

Without the tracking notes, future contributors cannot tell whether a feature was deliberately skipped on WASM or forgotten. Over time this erodes the WASM parity invariant.

## Evidence

Retrospective audit: `.tmp/orchestration/pr-audit-last-200.md` Section 3B.

slepp:
[...]

PR #2869, opened 2026-08-08
test(wasm): cooperative-timer parity probes for sleep and actor periodic timers

### LINKED ISSUE #1964: test(wasm): cooperative-timer parity probes for the sandbox-wasm suite
WASM timers are cooperative: `sleep` parks at the message boundary and time advances through the `hew_wasm_timer_tick` host call (`hew-runtime/src/scheduler_wasm.rs`). The v0.5 WASM parity suite (`hew-sandbox-wasm/tests/parity_ratchet.rs`) does not exercise this path.

Add parity probes that drive `hew_wasm_timer_tick` and assert the parked-then-resumed semantics — a plain `sleep` timeout and an actor periodic timer — so the cooperative-timer behaviour is covered against the native runtime.

Supersedes #1679, whose named tests (`wasm_e2e_timeout_timeout`, `wasm_e2e_actors_actor_periodic_timer`) were removed in the v0.5.0 rewrite.

gertybotbot:
Closes #1964.

## Venue correction

The issue asks for the probes in `hew-sandbox-wasm/tests/parity_ratchet.rs`. That crate can't host them: it has **no `hew-runtime` dependency** and compiles Hew to sandbox *bytecode* for the TypeScript VM (`hew-sandbox-vm`), where `sleep` is a virtual clock (`sleepVirtual`) with no host tick. `hew_wasm_timer_tick` is unreachable from there.

The real wasm32 execution harness is `hew-cli/tests/wasi_run_e2e.rs` (compiles to `wasm32-wasi`, runs under `wasmtime`, links the actual `hew-runtime`). The probes go there, matching that file's existing paired `wasm_*_run_pass` / `native_*_matches_wasi_output` idiom.

## What's covered

**(a) Plain `sleep` timeout — real parity.** A top-level `sleep(50ms)` parks at the message boundary and resumes; the wasm program terminates with stdout byte-identical to native. `wasm_cooperative_sleep_parks_and_resumes` + `native_cooperative_sleep_matches_wasi_output`.

**(b) Actor periodic timer — pinned divergence.** An `#[every(20ms)]` handler **does** fire under wasm (~249 ticks in 5s), but the program **never reaches quiescence**: with no host driving `hew_wasm_timer_tick`, the periodic queue keeps the scheduler permanently non-quiescent and `hew run` must kill it via `--timeout`. Native runs the same program to completion and exits 0.

This is a genuine native↔wasm divergence, so I pinned the divergence instead of asserting a parity that doesn't hold. The failure message instructs a future fixer to promote the probe to a stdout-parity case in the same commit that fixes quiescence — the test fails loudly if the runtime is fixed, so it can't rot into a stale allowance.

The probe deliberately avoids `await` inside the actor: that's still gated on wasm (`WASM-TODO(#1451)` — `hew_await_cancel_schedule_deadline_ms`), so an ask-based variant fails at codegen and proves nothing about the timer.

## Revert-control evidence

Both pairs were mutated to confirm they exercise the path rather than passing tautologically:

| Mutation | Result |
|---|---|
| `COOPERATIVE_SLEEP_EXPECTED` → `"before\n"` (as if the park never resumed) | both sleep tests **FAIL** |
| Remove `#[every(20ms)]` from the periodic actor | both periodic tests **FAIL** (`the native periodic handler never fired`; wasm reaches quiescence so the divergence assertion breaks) |

## Verification

- `cargo test -p hew-cli --test wasi_run_e2e` — **14 passed, 0 failed**; run twice, no flake.
- `cargo fmt --check -p hew-cli` — clean.
- `cargo clippy -p hew-cli --tests` — clean.

Test-only diff (+187 lines, one file). No runtime or compiler behaviour changed.

### DIFF
--- a/hew-cli/tests/wasi_run_e2e.rs
+++ b/hew-cli/tests/wasi_run_e2e.rs
@@ -540,6 +540,193 @@ fn native_hashmap_hashset_ownership_matches_wasi_output() {
     );
 }
 
+// ─────────────────────────────────────────────────────────────────────────
+//  Cooperative-timer probes (#1964)
+//
+//  WASM timers are cooperative: sleep parks at the message boundary and time
+//  advances through `hew_wasm_timer_tick` (hew-runtime/src/scheduler_wasm.rs).
+//  Nothing in the suite exercised that path, so both a working case and a
+//  divergent one could change silently. These two probes pin the observed
+//  behaviour of the cooperative timer against the native runtime.
+//
+//  Probe (a) — top-level `sleep`: parks and resumes, and the wasm program
+//  terminates with byte-identical stdout to native. This is real parity.
+//
+//  Probe (b) — actor periodic timer (`#[every]`): the timer DOES fire under
+//  wasm, but the program never terminates, because with no host driving
+//  `hew_wasm_timer_tick` the periodic queue keeps the scheduler permanently
+//  non-quiescent. Native runs the same program to completion. That is a real
+//  native↔wasm divergence, not a passing parity case, so probe (b) pins the
+//  divergence explicitly (bounded by `--timeout`) rather than asserting a
+//  parity that does not hold. If the runtime is fixed so the periodic actor
+//  reaches quiescence under wasm, this test fails and must be promoted to a
+//  stdout-parity case alongside the fix.
+// ─────────────────────────────────────────────────────────────────────────
+
+// A top-level `sleep` parks the main activation at the message boundary and is
+// resumed once the cooperative timer advances past the deadline. The prints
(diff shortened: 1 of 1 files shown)


slepp:
Nice venue call — `hew-sandbox-wasm` genuinely can't host these, since it has no `hew-runtime` dependency and its `sleep` is a virtual clock.

Ran all four probes clean, and confirmed the sleep pair actually fails when the expected output is wrong, so this is real coverage rather than a tautological pass. I took the liberty of flipping it out of draft to merge it — shout if that was premature.

PR #2872, opened 2026-08-09
cleanup(codegen): delete dead emit_struct_literal and two stale dead_code allows

### LINKED ISSUE #1885: Workspace cleanup: cross-crate duplication + dead code
A read-only audit of the non-stdlib workspace found a healthy structure (clean unidirectional frontend chain `parser → types → hir → mir`, genuine multi-consumer leaves, not over-split) with a concrete cleanup surface: five cross-crate duplications and three dead-code items. No crate should be merged mechanically.

## Cross-crate duplication

- **Native/wasm message-envelope ABI contract** (highest payoff, correctness-sensitive). The 10 `HEW_MSG_ENVELOPE_*` constants + `HewMsgEnvelope` struct/impls + `header_validate` are mirrored across `hew-runtime/src/mailbox.rs:165-183` and `mailbox_wasm.rs` (~80 lines of mirror code plus a bespoke per-constant parity test in `wasm_parity_tests.rs:358-398`). Extract into one shared `hew-runtime` module re-exported by both, keeping only a layout `const_assert`. This is a cross-target wire contract maintained by mirror + a hand-written parity test rather than a single source — same class as the native/wasm parity oracle in #1882.

- **Four parallel `HirExpr` walkers** (largest structural value). `layout_mono.rs:383/422`, `machine_mono.rs:932/1037`, and `lower.rs:3542/19539/19654` each re-implement a full `HirExpr`/`HirStmt` walk (~1500 lines of parallel match arms). A new `HirExprKind` variant fans out to all four today. Introduce one `HirVisitor` trait (`hew-hir/src/visit.rs`) and express the four as visitors.

- **Source-position helpers** (two duplications). A 1-based CRLF-aware `offset_to_line_col` and `compute_line_offsets` are duplicated across `hew-cli/diagnostic.rs:1154`, `hew-types/stdlib_loader.rs:172`, `hew-analysis/util.rs:17`, `hew-compile/lib.rs:214`. Promote one canonical pair into `hew-parser`/`hew-lexer` below all consumers. Keep the `hew-analysis` 0-based UTF-16 LSP family separate (intentionally different).

- **`TypeExpr` renderer.** Three renderers over the shared parser AST type (`hew-hir/lower.rs:860`, `hew-cli/doc/extract.rs:119`, `hew-analysis/hover.rs:949`) drift on new variants. Add one canonical renderer in `hew-parser` parameterized by a style flag, preserving the intentional surface divergence (`Option<T>` vs `T?`, pointer style) via the parameter.

## Dead code

- **Dead `emit_struct_literal` and stale suppression annotations** (near-free). Delete the dead test-only `emit_struct_literal` (`hew-codegen-rs/src/llvm.rs:31399-31416` + its test) and remove 8 stale `#[allow(dead_code)]`/`#[allow(unused_*)]` suppressions on now-live codegen helpers (`llvm.rs:31247/31263/31292/31349/31375/31434/289/295`). Restores the dead-code signal that would have caught the dead function itself.

- **`MembershipCallbackGeneration` epoch island** (decision-gated). The ~80-line `MembershipCallbackGeneration` epoch island in `hew-runtime/cluster.rs` (struct+impl `278-307`, accessors `1062-1072`, FFI entry `1895-1904`) is unreached — its sole consumer `remote_sup.rs` was deleted. Default to deletion per no-speculative-retention, after confirming no imminent supervision-protocol need. Do **not** touch `MembershipCallbackBinding` / `replace_membership_callback` (live write path).

## Crate boundaries — no mechanical merges

- `hew-compile` is the one arguable artificial split, but the higher-value direction is the opposite: promote it to the shared frontend driver that `hew-lsp`/`hew-analysis` currently reimplement around.
- `hew-capability-gen` is a deliberate forward seam — defer, do not merge.
- The testkit, `hew-cabi`, the frontend chain, and the product-surface binaries are all genuine boundaries and stay.
- The hypothesized "vestigial msgpack helpers from the dead Rust→C++ AST path" do not exist — every `hew_msgpack_*` symbol is live FFI for the user-facing msgpack stdlib crate.

The four-parallel-walkers and `TypeExpr`-renderer work are multi-file refactors; the rest are focused lanes. The position-helpers and dead-code items are good low-risk starters.

gertybotbot:
Picks up the "dead code" lane of #1885 (the "near-free" starter), re-derived against current `main` (`d4793f894`) rather than taken from the audit text — the line numbers there have drifted and the claimed "8 stale suppressions" is now **2**, but the substantive finding holds and one part of it turned out to be the opposite of what was filed.

### Genuinely dead: `emit_struct_literal`

No production consumer. Its only caller was a unit test of the helper itself, and its own docstring already recorded why: `monitor()` builds `MonitorRef { ref_id }` through `Instr::RecordInit` → `lower_record_init` directly, reaching the same underlying path without the wrapper.

Deleted the function, that test, and `fixture_monitor_ref_layout` — the fixture had no other caller and clippy flagged it the instant the test was removed, which is exactly the restored dead-code signal this lane is about.

### Stale, and masking *live* code

The other two `#[allow(dead_code)]` attributes in `llvm.rs` were annotated `W2.004 Stage 1 substrate; consumers in Stage 2/3 + W2.005`. Those consumers **have landed**, so the suppressions were no longer describing dead code at all:

| helper | production call sites |
|---|---|
| `emit_result_err` | `suspend.rs:6095,6152`; `runtime_abi.rs:1005,1151` (9 total) |
| `emit_enum_variant_literal` | `layout.rs:4717,4725,4850,4858` (14 total) |

Both removed. `cargo clippy -p hew-codegen-rs --lib --all-features --all-targets` is clean, which is the proof neither was load-bearing — had either still been dead, `-D warnings` would have caught it.

This is worth noting for #1885 generally: the audit's dead-code section assumed the suppressions were stale *because the helpers were still dead*. For two of three, they were stale because the helpers came alive and nobody removed the annotation.

### Also

Updated the substrate header comment (`llvm.rs:~24885`) from "Four substrate helpers" to three, dropped the `emit_struct_literal` precedent bullet, and removed the now-orphaned signature-rationale block that explained its `FieldOffset`/`struct_name` choices.

### Verification

- `cargo clippy -p hew-codegen-rs --lib --all-features --all-targets` — clean, no warnings
- `cargo test -p hew-codegen-rs --lib` — **239 passed, 0 failed**

No behavior change; deletions only (+2/−106).

Does not close #1885 (the duplication lanes and the `MembershipCallbackGeneration` decision-gated item remain).

### DIFF
--- a/hew-codegen-rs/src/llvm.rs
+++ b/hew-codegen-rs/src/llvm.rs
@@ -24882,8 +24882,8 @@ pub(crate) fn emit_node_lookup_call<'ctx>(
 // Composite-return spine — Lane W2.004 Stage 1
 // ----------------------------------------------------------------------------
 //
-// Four substrate helpers that synthesize user-visible composite return values
-// (`Result<T, E>`, struct literals, enum-variant literals) into a destination
+// Three substrate helpers that synthesize user-visible composite return values
+// (`Result<T, E>`, enum-variant literals) into a destination
 // `Place`. Each helper resolves the destination's registered layout
 // (`EnumLayout` via `machine_layouts`, `RecordLayout` via `record_layouts`)
 // and stores the tag + per-field payload through the existing
@@ -24895,21 +24895,8 @@ pub(crate) fn emit_node_lookup_call<'ctx>(
 //   - `emit_remote_pid_send_call` Err-arm `llvm.rs:~7553-7641` — `emit_result_err`
 //   - `emit_node_lookup_call`     Ok-arm `llvm.rs:~7783-7819` — `emit_result_ok`
 //   - `emit_node_lookup_call`     Err-arm `llvm.rs:~7747-7781` — `emit_result_err`
-//   - `Instr::RecordInit` `lower_record_init` — `emit_struct_literal`
-//     (no in-scope refactor; helper is the substrate for future Cluster-2
-//     consumers per audit §C.3)
 //   - `emit_enum_variant_literal` — no precedent; new substrate for Stage 3
 //     MIR producers (W2.005).
-//
-// Helper signatures (push-back per A259 #3(b) vs the plan):
-//   - `emit_struct_literal` uses `&[(FieldOffset, Place)]` (the MIR
-//     primitive used by `Instr::RecordInit`) rather than the plan's
-//     `&[(FieldName, Place)]`. `FieldName` is not a type in the MIR model;
-//     name → offset resolution happens at MIR-producer time per the
-//     `RecordInit` docstring (`hew-mir/src/model.rs:1747`).
-//   - `emit_struct_literal` does not take `struct_name`: the registered
-//     record name is recovered from `place_resolved_ty(dest)`'s
-//     `ResolvedTy::Named { name, .. }`, identically to `lower_record_init`.
 //
 // Constraints (CLAUDE.md customs):
 //   - **#2 Fail-closed Codegen**: any missing layout / mismatched type /
@@ -25064,7 +25051,6 @@ pub(crate) fn emit_result_ok(
 /// NodeRoutingNotWired`), the caller constructs the source local with
 /// the discriminant already set; this helper does not synthesise
 /// variant tags for the payload type.
-#[allow(dead_code)] // W2.004 Stage 1 substrate; consumers in Stage 2/3 + W2.005
 pub(crate) fn emit_result_err(
     fn_ctx: &FnCtx<'_, '_>,
     dest: Place,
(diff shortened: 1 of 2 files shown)


slepp:
Verified `emit_struct_literal` has no remaining callers anywhere in the tree, and that `emit_result_err` and `emit_enum_variant_literal` do have real production call sites now — so removing those two `#[allow(dead_code)]` was correcting a suppression that had started hiding live code, not just tidying. Clean and useful, merged.

PR #2873, opened 2026-08-09
feat(lsp,wasm): surface MIR-stage lints as warnings

### LINKED ISSUE #2176: Surface MIR-stage lints (dead_store / clean_counter) in the LSP and wasm/playground
## Background

The semantic lint pass has two surfacing tiers:

- **Checker-stage lints** (`needless_range_loop` and the M2 checker lints) are emitted through the type checker's `warnings` vector, which already flows to the CLI, the LSP, and the wasm/playground with no extra plumbing.
- **MIR-stage lints** (`dead_store`, `clean_counter`) require dataflow (backward liveness) that only exists after MIR lowering.

Today the LSP (`hew-lsp/src/server/analysis.rs`) and the wasm front end (`hew-wasm/src/lib.rs`) stop their analysis at the HIR stage — neither lowers to MIR. As a result, MIR-stage lints surface **only through the CLI** (`hew check` / `build` / `run` / `compile`), not in editors or on the website.

## Goal

Surface MIR-stage lint diagnostics (as warnings) in the LSP and the wasm/playground, so editor and web users see `dead_store` / `clean_counter` (and future MIR lints) the same way they see the checker-stage lints.

## Notes / constraints

- This needs a MIR-lowering + lint step added to the LSP `analyze_document` pipeline and the wasm `parse_and_type_check` pipeline, plus a diagnostic-conversion path that renders MIR lints as **warnings**. Note that the HIR diagnostics in both surfaces are currently mapped to LSP `ERROR` / wasm `"error"` unconditionally — the MIR lint path must not reuse that error mapping.
- The wasm path runs in-browser; lowering to MIR there has bundle-size and runtime-cost implications that should be measured (consider gating it or running it lazily).
- The lints themselves and their CLI surfacing land first; this issue tracks the deferred editor/web surfacing.

See `docs/design/lint-pass.md` (§ surfacing) for the architecture.

gertybotbot:
Closes #2176.

## What changed

Both `hew-lsp/src/server/analysis.rs` and `hew-wasm/src/lib.rs` stopped at HIR, so MIR-stage lints reached only the CLI. Both now lower to MIR and run `run_mir_lints` on the module **they already lowered for HIR diagnostics** — no second lowering — and apply the CLI's suppression policy (`// hew:allow(...)`, `LintLevel::Allow`, and skipping out-of-range spans that belong to imported modules).

Findings render as **warnings on a dedicated path**. The existing HIR conversions map every diagnostic to LSP `ERROR` / wasm `"error"` unconditionally; reusing them would make a style finding fail an editor buffer the native compiler accepts.

Lowering failures degrade silently. `IrPipeline::diagnostics` (the hard `E_MIR_*` move/init errors) are dropped rather than promoted into editor errors, and the already-computed HIR diagnostics are never lost. MIR lints run only when HIR lowering is clean.

## Scope correction

The issue names `dead_store` *and* `clean_counter`. **`clean_counter` does not exist at `d4793f894`** — `run_mir_lints` calls `detect_dead_stores` and nothing else. This PR therefore surfaces `dead_store`, and the plumbing is keyed off `IrPipeline::lint_warnings` rather than any specific lint, so `clean_counter` and any future MIR lint surface in both places with no further work.

## Verification

Exit codes captured bare, not through a pipe.

| command | rc |
|---|---|
| `cargo test -p hew-lsp -p hew-wasm` | 0 (351 passed, 0 failed) |
| `cargo clippy --all-features --all-targets` | 0 |
| `cargo build -p hew-wasm --release --target wasm32-unknown-unknown` | 0 |

Six new tests, each proving **discrimination** rather than presence — three in the LSP (one at the unit seam, three via `protocol_smoke`) and three in wasm:

- the dead store produces the finding, and its severity really is `WARNING` / `"warning"`, not error;
- a near-identical `for i in 0..n` accumulator, where every store is read, stays **silent**;
- `// hew:allow(dead_store)` suppresses it on both surfaces.

Ground truth confirmed at the CLI before any code was written: the positive fixture warns, the accumulator control is silent.

## Bundle cost — please read before merging

The issue asks for this to be measured, and it is not free:

| | bytes | |
|---|---|---|
| `main` @ `d4793f894` | 4,771,978 | 4.55 MiB |
| this branch | 6,482,389 | 6.18 MiB |
| **delta** | **+1,710,411** | **+1.63 MiB (+35.8%)** |

Measured on raw `cargo build --release --target wasm32-unknown-unknown` from a clean worktree of `main` in a separate target dir. **This is pre-`wasm-opt`**; the shipped playground artifact goes through `wasm-pack` with `-O1`, which I could not run locally, so the delivered delta will be smaller than this — but the direction and rough magnitude stand.

The *runtime* cost is already gated (MIR lowering runs only on buffers that parse, type-check and lower cleanly, so it stays off mid-edit buffers). The **code-size** cost is from linking `hew-mir` into the bundle and no gating removes it — that would need a split artifact or lazy second module, which is a bigger design decision than this issue.

If +1.6 MiB pre-opt is unacceptable for the playground, the LSP half stands alone and I'm happy to split the wasm half out into its own PR behind whatever loading strategy you prefer.


slepp:
Merged. Keeping the LSP/wasm severity mapping on its own path instead of reusing the HIR-to-error conversion is the right call, and the controls make the tests prove something — silent on the accumulator loop where every store is read, suppressed by the allow directive.

On the bundle size you flagged: accepted, parity with the CLI's diagnostics is worth it. Thanks for surfacing it explicitly rather than burying it.

Issue #1924, opened 2026-06-15
Flaky: macOS Build & test workspace-tests step (subprocess/eval harness contention)


slepp:
The macOS arm64 Build & test job's 'Run Rust workspace tests' step fails intermittently on subprocess-heavy tests, with the failure surfacing in the test support harness rather than an assertion. Most recently eval_wasm_file_runtime_failure_exits_with_child_exit_code panicked at hew-cli/tests/support/mod.rs:37 on a PR that does not touch hew-cli or wasm-eval; the same step flaked on unrelated PRs this cycle while Linux and Windows pass the identical tests.

Likely cause: contention/timing on the slower macOS runner for tests that spawn the compiled hew and link libhew.a (cf. the warm-binary work in #1887). Fix the harness setup race and/or warm a settled binary on macOS before the bounded exec/eval runs, rather than relying on re-runs.

slepp:
Still real — I checked the last 10 completed (non-cancelled) macOS arm64 CI runs on `release/v0.6.0-rc1`, one per distinct commit SHA, spanning 2026-08-03 through 2026-08-07:

| SHA | Failing test |
|---|---|
| fd097c670 | `websocket::tests::repeated_bind_accept_close_drains_all_accept_authority` (websocket.rs:3347) |
| 0487a3a3ad | `debugger_renders_only_active_enum_variant_payload` (dwarf_debugger_locals_e2e.rs:985), then `websocket::tests::repeated_bind_accept_close_drains_all_accept_authority` (websocket.rs:3318) |
| c25c7345a5 | `debugger_renders_only_active_enum_variant_payload`, then `machine_snapshot_select_watch_matches_state_variants` (machine_transition_watch_e2e.rs:103), then a third failure |
| f4b65582ed | same two-test pattern as above |
| 8939957a51 | same two-test pattern as above |
| 06c2f19b4c | `debugger_renders_only_active_enum_variant_payload`, then `supervisor_child_i32_fixed_sleep_oracle_misses_delayed_turn` (supervisor_init_args_e2e.rs:217) |
| ef1a820791 | `debugger_renders_only_active_enum_variant_payload`, then `machine_snapshot_select_watch_matches_state_variants` |
| eca907394c | failure (older run, same class) |
| 5f4849702e | failure |
| 76df5732be, 7851454e01, d321307595 | failure |

[...]

Issue #1887, opened 2026-06-14
Exec tests false-timeout under target/ contention via in-deadline `cargo run` fallback


slepp:
## Symptom

`hew-codegen-rs` exec integration tests (`generator_exec`, `ffi_int_width_exec`, `float_return_exec`, and the wider `*_exec` family) intermittently FAIL or TIMEOUT when another `cargo`/`nextest` invocation shares the same `target/` directory — e.g. a concurrent `cargo nextest run --workspace --profile ci`. The failure presents as `panicked ... exceeded 60s wall-clock deadline; process tree was killed`, not an assertion.

## Root cause

`hew_command()` (defined identically in ~20 exec test files, e.g. `hew-codegen-rs/tests/generator_exec.rs`, `float_return_exec.rs`, `hashmap_keys_values_exec.rs`) resolves the `hew` binary like this:

```rust
fn hew_command(repo: &Path) -> Command {
    let bin = hew_bin(repo);          // target/debug/hew
    if bin.exists() {
        return Command::new(bin);     // warm: direct exec, ~0.14s
    }
    // cold fallback: compile hew INSIDE the test
    let mut command = Command::new(cargo);
    command.current_dir(repo).args(["run", "--quiet", "-p", "hew-cli", "--bin", "hew", "--"]);
    command
}
```

The command is then run under a 60s `hew_testutil::run_command_bounded` deadline. When `target/debug/hew` is absent (cold) **and** another process holds the cargo build-lock on the shared `target/`, the fallback `cargo run` blocks on `Blocking waiting for file lock on build directory` and burns the entire 60s budget → deadline trip → false FAIL. With a warm `target/debug/hew` the same fixture compiles in 0.14s and passes.

This is the structural mechanism behind the documented "a 600s-budget gate on a cold target is a false red" lesson — except here the per-test budget is only 60s and the cold cost is a full `hew` link, so it is far easier to trip.

## Fix

Warm `target/debug/hew` once, before the bounded runs, instead of letting a cold `cargo run` happen inside the deadline. The pattern already exists right next to `hew_command`: `ensure_hew_runtime_lib()` builds `libhew.a` once via a `OnceLock` if absent. An analogous `ensure_hew_bin()` (build `-p hew-cli --bin hew` once under a `OnceLock`, outside any per-test deadline) makes the exec corpus robust to `target/` contention. Alternatively, resolve the binary via `env!("CARGO_BIN_EXE_hew")` so cargo guarantees it is built before the test binary runs.

[...]

PR #2874, opened 2026-08-09
test(hew-cli): serialize wasm stdlib archive builds under the shared bootstrap lock

gertybotbot:
Refs #1924.

## The asymmetry

`bootstrap_wasi_runner` calls two builders in sequence:

```rust
build_wasi_runtime_serialized(&target_dir, build_profile)?;
for (package, archive) in WASI_STDLIB_ARCHIVES {
    build_wasi_stdlib_archive(&target_dir, build_profile, package, archive)?;
}
```

The first takes a cross-process `fd_lock` before invoking cargo, and its doc comment
already names the hazard precisely:

> without it, multiple test binaries each see the artifact absent and each launch
> `cargo build` concurrently; Cargo writes the staticlib non-atomically, so races
> corrupt or transiently delete the file and wasm-ld then fails with
> "cannot open `libhew_runtime.a`".

The second builds the same kind of wasm32-wasip1 staticlib, from the same loop, into
the same directory — and had the identical `is_file()` fast path with **no lock at
all**. `libhew_std.a` was left exposed to exactly the race its sibling documents.
`hew-testutil`'s `ensure_hew_lib_built` locks too, so the unlocked stdlib builder was
the only one of the three that didn't.

That matches #1924's signature: the failure surfaces in the support harness rather
than an assertion, appears on PRs that touch neither hew-cli nor wasm-eval, and is
far likelier on the slower macOS runner where the build window stays open longer.

## The change

Hoist the lock into `open_wasi_bootstrap_lock` and take it in both builders, so every
wasm32-wasip1 staticlib build serializes on one lock file regardless of which artifact
a given process found missing. A per-builder lock would not do: the point is that
different builders contend with each other.

Both keep their unlocked fast path and their post-lock re-check, so the common
all-artifacts-present case (CI pre-build, or a sibling finished first) is unchanged.

## Verification

Mutual exclusion is proven directly rather than assumed, with a control:

| | order observed |
|---|---|
| **with shared lock** | `ENTER runtime / EXIT runtime / ENTER stdlib-hew-std / EXIT stdlib-hew-std / ENTER stdlib-2 / EXIT stdlib-2` |
| **control, lock removed** | `ENTER runtime / ENTER stdlib-hew-std / ENTER stdlib-2 / EXIT... ` |

Three processes on the real `fd_lock` + real lock filename. Locked: perfectly nested,
no interleaving. Unlocked: all three inside the critical section at once — i.e. three
concurrent `cargo build` invocations racing the same non-atomic staticlib write.

- `cargo clippy -p hew-cli --tests` → RC=0 (captured bare, not through a pipe)
- `cargo test -p hew-cli --test eval_e2e wasm` → **16 passed, 0 failed**, RC=0,
  including `eval_wasm_file_runtime_failure_exits_with_child_exit_code`, the test
  named in the issue.

## What this does not claim

A flake fix cannot be proven by a green run — passing was always the common case. The
evidence here is structural: a documented race, a sibling that guards against it, an
unguarded third path, and a control showing the guard is what prevents concurrency.
If #1924 recurs after this, the remaining suspect is the warm-binary/settling issue
in the body's second hypothesis (cf. #1887), which this does not address.

Issue #1924 is left open deliberately — closing it needs runner evidence over time,
not a merge.

### DIFF
--- a/hew-cli/tests/support/mod.rs
+++ b/hew-cli/tests/support/mod.rs
@@ -128,7 +128,10 @@ fn build_wasi_runtime_serialized(target_dir: &Path, build_profile: &str) -> Resu
 
     let mut lock = open_wasi_bootstrap_lock(target_dir)?;
     let _guard = lock.write().map_err(|error| {
-        format!("failed to lock WASI bootstrap in {}: {error}", target_dir.display())
+        format!(
+            "failed to lock WASI bootstrap in {}: {error}",
+            target_dir.display()
+        )
     })?;
 
     // Re-check after acquiring the lock: a sibling may have built while we waited.
--- a/hew-cli/tests/support/mod.rs
+++ b/hew-cli/tests/support/mod.rs
@@ -83,6 +83,28 @@ fn bootstrap_wasi_runner() -> Result<(), String> {
     Ok(())
 }
 
+/// Open and return the shared cross-process WASI bootstrap lock.
+///
+/// Every wasm32-wasip1 staticlib built by this harness shares one lock file, so
+/// concurrent nextest processes serialize against each other regardless of
+/// which artifact each one happened to find missing.
+fn open_wasi_bootstrap_lock(target_dir: &Path) -> Result<RwLock<std::fs::File>, String> {
+    let lock_path = target_dir.join("hew-cli-wasi-bootstrap.lock");
+    let lock_file = OpenOptions::new()
+        .create(true)
+        .read(true)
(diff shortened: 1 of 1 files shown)


slepp:
Good catch on the asymmetry — the stdlib archive builder really was the only one of the wasm32-wasip1 staticlib builders with no cross-process lock, while its sibling already had one. Having both contend on the same lock file is the right fix. Merged.

PR #2876, opened 2026-08-09
fix(runtime): bound worker joins during scheduler teardown

### LINKED ISSUE #2508: teardown_workers joins parked workers with no timeout (unreachable for compiled main today; blocks a future host-embedder path)
Found while investigating #2446. `teardown_workers` (`hew-runtime/src/scheduler.rs:569`) hard-joins every worker thread with no timeout. A worker genuinely parked in a blocking syscall (e.g. `accept()`) never reaches the park-recheck, so this join is unbounded.

This does NOT affect any compiled `.hew` binary today — `hew_runtime_cleanup` (the only route to `teardown_workers`) is never emitted by codegen for native `main`, confirmed by direct process sampling during the #2446 investigation (the actual exit path for a compiled binary goes through `shutdown_orchestrate`'s own bounded drain timeout instead). It's latent, currently-unreachable hardening for a not-yet-wired host-embedder path (`runtime_handle.rs`).

Before that embedder path is wired up for real use, this join needs a bound — otherwise a worker parked in a runtime-owned blocking op at that point could hang the embedding host's shutdown forever.

gertybotbot:
Closes #2508.

## The defect

`teardown_workers` (`hew-runtime/src/scheduler.rs`) hard-joined every worker thread with no timeout:

```rust
if let Some(h) = handle.take() {
    if h.join().is_err() { ... }
}
```

A worker genuinely parked in a blocking syscall (e.g. `accept()`) never reaches the park-recheck, so `join()` never returns and shutdown hangs forever.

## Scope correction

The issue says this "does NOT affect any compiled `.hew` binary today" and is latent hardening for the not-yet-wired host-embedder path. **That understates it.**

`teardown_workers` has three callers, and one is `hew_sched_shutdown` — which codegen emits *directly into native `main`* for supervisor programs (`hew-codegen-rs/src/llvm.rs`, `emit_immediate_shutdown_epilogue`, alongside the `hew_sched_shutdown` ABI declaration in `runtime_abi.rs`). Supervisor programs deliberately bypass the generic idle drain and go straight to closing and joining workers, so they reach the unbounded join today. `hew_runtime_cleanup` being unreachable is true but is not the only route.

The fix is the same either way; the reachability is just wider than filed.

## The fix

Poll `is_finished()` against a deadline, and `join()` only once the thread is known finished so the join itself cannot block.

Two deliberate choices:

- **One deadline for the whole worker set**, not per-handle — N workers each parked in a blocking syscall must not multiply the bound by N. Covered by a dedicated test.
- **Detach, do not force-kill.** A worker past the deadline may still be touching scheduler-owned memory and there is no safe way to interrupt it, so the handle is leaked via `mem::forget` and reported on stderr. Bounding shutdown must not introduce a use-after-free; leaking a thread is the fail-closed choice. This mirrors the existing task-scope reaper precedent.

`WORKER_JOIN_TIMEOUT` is 5s, matching `shutdown::DEFAULT_DRAIN_TIMEOUT_MS`. By the time teardown runs, the drain phase has already had its own budget, so this is a backstop against a worker that *cannot* return — not a second drain window. The polling shape follows the house convention in `shutdown::drain_until_idle`.

The stale doc comment on `hew_sched_shutdown` (which promised an unconditional join of every worker) is updated to state the bound.

## Verification

Three tests, proving discrimination rather than presence:

| test | proves |
|---|---|
| `teardown_joins_workers_that_exit` | **CONTROL** — a normally-exiting worker is still *joined* (its write is visible on return), and teardown returns well before the timeout |
| `teardown_abandons_a_worker_that_never_exits` | a stuck worker is bounded and teardown returns |
| `teardown_deadline_is_shared_across_workers` | 3 stuck workers cost ~1x, not 3x, the timeout |

**Revert-control:** with the join loop reverted to `h.join()` but the tests kept, `teardown_abandons_a_worker_that_never_exits` runs past 60s and has to be killed — i.e. it reproduces exactly the hang #2508 describes. With the fix it passes in ~5s. The test is a real discriminator, not a tautology.

- `cargo test -p hew-runtime --lib` — **2243 passed**, RC=0
- `cargo clippy -p hew-runtime --all-targets -- -D warnings` — RC=0
- `cargo fmt -p hew-runtime -- --check` — RC=0

The pre-existing `spawn_failure_teardown_joins_partial_worker_set_and_drops_scheduler` regression test still passes.

## Unrelated flake found while verifying

`transport::tests::framed_send_to_broken_pipe_fails_closed_without_signal` fails intermittently in full-suite runs. I initially suspected my change and ran it down:

- **unmodified `main`, 6 full-suite runs: 5 ok / 1 FAILED** — so it is pre-existing and not caused by this PR.
- It passes in isolation on both branches; only concurrent full-suite runs trigger it.
- The assertion failure reports `left: 4096` — a *fully successful* 4096-byte write where `-1` was expected. So the send did not hit EPIPE at all: after `drop(far)` closes the peer fd, a concurrent thread elsewhere in the suite opens a new fd that reuses that number, and the write lands on a live socket. Classic fd-reuse race inherent to the test, not a defect in the SIGPIPE suppression it is meant to guard.

My diff contains no fd-touching code (only sleeps, `is_finished()`, `join`, and `mem::forget`), though the added ~5s of wall-clock does widen the window. Filing separately rather than folding an unrelated fix into this PR.

### DIFF
--- a/hew-runtime/src/scheduler.rs
+++ b/hew-runtime/src/scheduler.rs
@@ -28,7 +28,7 @@ use std::sync::atomic::{AtomicBool, AtomicU64, Ordering};
 use std::sync::atomic::AtomicPtr;
 use std::sync::{Condvar, Mutex};
 use std::thread::{self, JoinHandle};
-use std::time::Duration;
+use std::time::{Duration, Instant};
 
 use rand::RngExt;
 
@@ -805,6 +805,17 @@ pub extern "C" fn hew_sched_init() -> c_int {
 /// scheduler — deques, parkers, stealers — as the final teardown step). The
 /// scheduler is owned by the runtime, so detaching the runtime is what frees
 /// the scheduler.
+/// Maximum total time worker teardown will wait for ALL workers to exit.
+///
+/// Chosen to match `shutdown::DEFAULT_DRAIN_TIMEOUT_MS` (5s): by the time
+/// teardown runs, the drain phase has already had its own budget, so this is a
+/// backstop against a worker that cannot return (parked in a blocking syscall),
+/// not a second drain window.
+const WORKER_JOIN_TIMEOUT: Duration = Duration::from_secs(5);
+
+/// Poll interval while waiting for a worker to finish.
+const WORKER_JOIN_POLL_INTERVAL: Duration = Duration::from_millis(2);
+
 fn teardown_workers(
     scheduler: Option<*const Scheduler>,
     handles: Option<Vec<Option<JoinHandle<()>>>>,
@@ -832,6 +843,10 @@ fn teardown_workers(
         })
     });
     let current_id = thread::current().id();
+    // ONE deadline for the whole set, not per-handle: N workers each parked in a
+    // blocking syscall must not multiply the bound by N.
+    let deadline = Instant::now() + WORKER_JOIN_TIMEOUT;
+    let mut abandoned = 0usize;
     for handle in &mut handles {
         if let Some(ref h) = handle {
             if h.thread().id() == current_id {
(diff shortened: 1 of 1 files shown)


slepp:
The bounded-join fix is solid — sharing one deadline across the whole worker set instead of per-handle, and detaching rather than force-killing a stuck worker, are the right calls here. I reverted the fix locally to confirm the test actually reproduces the hang before trusting it, and it does.

Appreciated the honest scope correction on where this is reachable today, and filing the unrelated transport flake separately instead of folding it in. Merged.

PR #2878, opened 2026-08-10
feat(mir): implement clean_counter via faint-variable analysis

### LINKED ISSUE #2178: Implement clean_counter lint (dead loop-counter detection via faint-variable analysis)
## Goal

Implement the `clean_counter` lint: flag a loop-carried counter/accumulator local whose value is **never read after the loop exits** (the counting is dead work), suggesting the counter be removed.

## Why it isn't implemented yet

The semantic lint pass (M3) added a backward **liveness** analysis and the `dead_store` lint, but `clean_counter` was deferred because plain liveness is insufficient to make it precise:

1. **Shape recovery.** A manual `c = c + 1` lowers through a temp + a checked `IntArithChecked` (which can trap) + a `Move c = temp`, not a single self-update instruction — so the "counter" shape is not directly recognizable in MIR.
2. **Liveness can't separate the populations.** A dead accumulator `c` and a legitimate `for i in 0..n` counter `i` are *both* "live inside the loop, dead at the loop exit". Distinguishing "the counting is dead work" from "a normally-scoped loop variable" needs **faint-variable / strong-liveness** analysis (whether a *value*, not merely the local, influences an observable), which is a second dataflow pass beyond the liveness M3 added.
3. **Checked arithmetic defeats the naive premise.** The increment's overflow flag feeds a trap branch, so `c` is *strongly live* — its value decides whether the program traps. Removing `c = c + 1` would be semantics-changing exactly where the naive lint looks applicable.

It would also overlap/double-fire with `dead_store` on the straight-line `c = c + 1` case.

## What's needed

A faint-variable (strong-liveness) analysis in `hew-mir` layered on the existing backward liveness (`hew-mir/src/liveness.rs`), plus counter-shape recognition that accounts for the checked-arith lowering, so the lint fires only when removing the counter is provably semantics-preserving.

## Notes

- The lint name `clean_counter` is intentionally **not** registered in `hew_types::LintId` until detection exists, so `-D clean_counter` fails closed ("unknown lint") rather than silently no-opping.
- Design rationale lives in `docs/design/lint-pass.md` (§10, M3).

gertybotbot:
Closes #2178.

Implements the `clean_counter` MIR lint via faint-variable analysis: a counter that is
updated but never observed is reported as removable.

## Scope correction (important)

The lint ships **scoped to float accumulators**, not integers. This is narrower than
#2178 implies, and the narrowing is a soundness result rather than a shortcut:

- Integer counters under checked arithmetic are **genuinely strongly live**. The
  overflow flag feeds `Branch` → `Trap`, so the counter is observable through the trap
  edge and removing it would change program behaviour. These correctly never fire.
- Float accumulation has no trap edge, so an unobserved float counter is soundly
  removable.

`IntArithChecked` is therefore deliberately kept **off** the purity allowlist.

## Discrimination, not presence

Tests are built as controls rather than assertions that the lint exists:

- two identical accumulators in one function — only the **unobserved** one fires
- integer counters: silent (per the trap-edge result above)
- for-range counters: silent
- returned counters: silent
- printed counters: silent

`clean_counter` is registered as a real lint, so `-D` / `-W` / `-A` all work.

## Verification (run bare, exit codes read directly)

- `cargo test -p hew-mir` → RC=0
- `cargo test -p hew-cli --test lint_pass_e2e` → 40/40, RC=0
- `cargo clippy --workspace --all-targets` → RC=0
- `cargo fmt --all --check` → RC=0

## Relationship to #2873

#2873 wires MIR lint findings into the LSP and wasm/playground surfaces keyed off
`IrPipeline::lint_warnings`. That plumbing is already written such that `clean_counter`
surfaces on both surfaces for free once this lands — the two are independent but
complementary, and this one is the substrate.


slepp:
Thanks — this is a nice piece of work. What I appreciated most is that the checked-arithmetic problem became the *scope* of the lint rather than something to work around: integer counters are genuinely observable through the trap edge, and keeping `IntArithChecked` off the purity allowlist says so plainly.

I mutation-tested the soundness argument rather than taking it on trust. Deleting the faintness gate turns four tests red, and widening both the type guard and the shape allowlist to admit integers leaves the lint silent — so faintness really is the guard doing the work, and the other two are belt-and-braces. Ran the compiled Hew suite and scanned all 1863 `.hew` files: green, no spurious firings.

Two small followups, neither worth holding this up. In `recover_counter_shape`, the backward scan returns as soon as it finds the temp's defining instruction, so if the first `Move` into a local isn't an accumulate, later writebacks to that local in the same block never get examined — a `continue` would widen it. And an accumulator updated in both arms of an `if` inside a loop emits one warning per arm; defensible, since each line is dead work, but worth a deliberate call.

Issue #2864, opened 2026-08-07
Machine-embedding aggregate layouts: record-wrapped machines and machine-in-machine ordering


slepp:
## Summary

Two valid machine-embedding aggregate shapes do not yet compile. Both currently **fail closed** with a clear diagnostic that names the offending member — they do not miscompile or corrupt memory — but they are valid shapes that should eventually be supported. Backstopping them is the opaque-member layout guard added alongside the direct `Option<machine>` sizing fix, which turns any layout-registration-ordering gap into a compile error instead of silent under-sizing.

The direct and nested/generic cases (`Option<Conn>`, `Envelope<Result<GenericConn<i64>, string>>`, and similar) are fixed and size correctly.

## Gap 1 — an enum payload containing a record that contains a machine

```hew
type Wrapper { conn: Conn; tag: i64; }
let wrapped: Option<Wrapper> = rx.recv();
```

The enum machine-dependency classifier follows the resolved-type spine but does not descend into a named record's fields, so the enum is sized before `Wrapper` (and its embedded machine) is complete. Today this fails closed:

```
tagged-union `Option$$Wrapper` variant `Some` field of type `Wrapper` resolves to a still-opaque LLVM struct
```

Fix: treat a record field's machine dependency as an enum-deferral dependency.

## Gap 2 — a machine embedded in another machine, declaration-order-dependent

```hew
machine Outer { state Holding { inner: Inner } }
machine Inner { state Ready { payload: string } }
```

With `Inner` declared first this compiles and sizes correctly; with `Outer` first it fails closed, because machine layouts are registered in source order rather than dependency order (unlike enum layouts, which are topologically sorted).

Fix: dependency-sort machine layout bodies so the embedded machine is complete before its container is sized.

## Acceptance

- A positive IR test for the record-wrapped-machine shape (payload sized to the record's full size).
- Positive IR tests for machine-in-machine under **both** declaration orders.
- Keep the opaque-member layout guard after both fixes — it is the safety backstop that prevents any future registration-ordering regression from becoming heap corruption.

Target: v0.6.0 RC2.

gertybotbot:
Took the **Gap 2** half (machine-in-machine declaration order). PR: #2879. One correction worth making before anyone plans the rest of this issue.

### Gap 2 is not fail-closed — it silently miscompiles

The description says both gaps "**fail closed** with a clear diagnostic ... they do not miscompile or corrupt memory." That holds for Gap 1, but not for Gap 2.

[...]

PR #2879, opened 2026-08-10
fix(codegen): dependency-sort machine layouts before sizing

gertybotbot:
Addresses **Gap 2** of #2864 (machine embedded in another machine, declaration-order-dependent). Gap 1 is untouched.

## Correction to the issue

#2864 states both gaps "**fail closed** with a clear diagnostic ... they do not miscompile or corrupt memory." That is true of Gap 1, but **not of Gap 2**. Verified at `d4793f894`:

Two programs with **byte-for-byte identical token multisets**, differing only in declaration order:

| declaration order | result |
|---|---|
| `Inner` (embedded) first | prints `hello-world-payload`, rc=0 |
| `Outer` (container) first | compiles clean, **SIGSEGV**, rc=139 |

```
$ diff <(sort of_main.hew) <(sort if_main.hew) && echo IDENTICAL-TOKENS
IDENTICAL-TOKENS
$ ./of_main.bin
Segmentation fault (core dumped)
```

No diagnostic is emitted, and the opaque-member layout guard does not catch it: the container's payload array is sized `0` and the embedded machine's `string` field is read from unallocated memory. So this is a **silent miscompile**, which raises its priority relative to how it is filed.

## Cause

`register_machine_layouts` iterated `machine_layouts` in **source order**. `build_tagged_union_layout` sizes each variant payload via `TargetData::get_abi_size`; if a state payload holds another machine whose named struct is still opaque, `get_abi_size` returns 0.

`register_enum_layouts`, in the same file, already documents this exact hazard and solves it with a stable Kahn's topological sort. Machines simply never got the same treatment — the issue's own framing ("unlike enum layouts, which are topologically sorted") is correct.

## Fix

Apply the same stable Kahn's sort before the registration loop:

- Dependencies collected from **both** state and event payloads — the `<Name>Event` companion is built in the same pass and needs the same guarantee.
- Keyed by **short name**, matching the `machine_short_names` convention in `llvm.rs`.
- Recurses through generic args, tuples, arrays, and slices.
- Stable within each tier, preserving input order for unrelated machines (required by the first-registration-wins dedup below it).
- A layout cycle **fails closed** with a named diagnostic rather than falling back to input order, which would silently size a participant against an opaque member.

## Tests

`tests/hew/machine_in_machine_test.hew` covers **both** declaration orders — a fix handling only one is half a fix. Each asserts the exact embedded payload string, so a zero-sized or misaligned payload fails rather than reading back an accidental value.

**Verified discriminating (revert-control):** with the sort reverted and the tests kept, the container-first test fails and the embedded-first test still passes. The tests detect the defect rather than merely asserting presence.

## Verification

- `cargo test --release -p hew-codegen-rs` — **755 passed, 0 failed**, RC=0
- All **158** files in `tests/hew/` — 153 report `test result: ok` with 0 failed; the other 5 are `main`-style oracles with no `#[test]` functions and all exit rc=0
- `cargo fmt --all --check` — RC=0
- `cargo clippy --release -p hew-codegen-rs --all-targets -- -D warnings` — **zero warnings in `layout.rs`**

⚠️ That clippy invocation does exit non-zero, but on two **pre-existing** findings in `hew-hir/src/lower.rs` (`:6021`, `:6231` — `unused_self` and an `allow` without a reason). Confirmed present on unmodified `main` with this branch's changes stashed, so they are not introduced here and are left alone rather than folded into an unrelated fix.

### DIFF
--- a/hew-codegen-rs/src/layout.rs
+++ b/hew-codegen-rs/src/layout.rs
@@ -277,11 +277,31 @@ pub(crate) fn host_data_layout_string() -> &'static str {
 
 /// Collect machine-layout indices that `field_ty` depends on for sizing.
 ///
-/// Mirrors `collect_named_enum_deps`, but keyed by SHORT name because machine
-/// layouts are registered and looked up by short name (see the
-/// `machine_short_names` set in `llvm.rs`). Recurses through generic args,
-/// tuples, arrays, and slices: any of them can carry an embedded machine whose
-/// body must be set before the container is sized.
+/// Mirrors `collect_named_enum_deps`. Keyed by [`machine_dep_key`] because a
+/// `MachineLayout.name` is a canonical class-tagged mono key (`mc$$Inner$$`)
+/// while a payload field spells its type nominally (`Inner`); both sides are
+/// normalized to the same projection so the lookup can resolve. Recurses
+/// through generic args, tuples, arrays, and slices: any of them can carry an
+/// embedded machine whose body must be set before the container is sized.
+/// Normalize a machine name to the key used for dependency lookups.
+///
+/// Two spellings must meet here. `MachineLayout.name` is the canonical
+/// class-tagged monomorphization key produced by
+/// `hew_hir::mono::machine_layout_key` — `mc$$Inner$$`, or
+/// `mc$$control.Lifecycle$$i64$$` once module-qualified and generic. A state
+/// payload field, by contrast, spells its type nominally: `Inner`.
+///
+/// Keying the graph on `short_name` alone left the `mc$$` tag on one side and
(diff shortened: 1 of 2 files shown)


slepp:
Both of these hold up. The machine-in-machine sort mirrors the enum one closely enough that it reads as the obvious missing half, and I confirmed the cycle branch actually fires and fails closed with a named diagnostic rather than being dead code.

I reproduced the pre-fix defect by reverting the sort to input order — on macOS/arm64 it comes out as a silent wrong exit rather than the segfault you saw, so the symptom is platform-dependent, but the silent-miscompile framing is right and worth having on the record. Merged.

PR #2880, opened 2026-08-10
fix(codegen): descend through record fields when sorting machine layouts

gertybotbot:
Related to #2864, but does **not** close it — corrected during review.

This fixes a third shape the issue does not name: a machine state payload reaching a machine *through a record field*. "Gap 1" as filed is the enum-side classifier (`resolved_ty_references_machine`, `hew-codegen-rs/src/llvm.rs:2593`), which does not descend records either and is untouched here — the issue's own example still fails on this branch with `E_NOT_YET_IMPLEMENTED: indirect-enum free synthesis`. #2864 stays open for the enum side.

⚠️ **Stacked on #2879.** The fix lives *inside* the dependency sort that #2879 introduces, so this branch contains that commit as its parent. Review/merge #2879 first; this diff is the 43-line descent on top of it.

## The gap

`register_machine_layouts` sorts machines so a container is never sized against a still-opaque embedded machine. That walk looks for payload field types whose name resolves to a machine — but a **record is not a machine**, so it missed the lookup and stopped. A machine reached only *through* a record field was never registered as a dependency:

```hew
machine OuterG { state HoldingG { w: WrapG } ... }
type WrapG { inner: InnerG; }
machine InnerG { state ReadyG { payload: string } ... }
```

## It fails SILENTLY — worse than Gap 2

Gap 2 segfaulted. This one produces **no diagnostic, no output, and exit 1**. Two programs with byte-for-byte identical token multisets, differing only in declaration order:

```
$ diff <(tr -s ' \n' '\n\n' <gap1.hew|sort) <(tr -s ' \n' '\n\n' <gap1_ctl.hew|sort)
IDENTICAL-TOKENS

$ hew run gap1_ctl.hew    # embedded declared first
gap1-payload              # rc=0

$ hew run gap1.hew        # container declared first
                          # rc=1, no stdout, no diagnostic
```

This is worth noting against #2864's claim that both gaps "fail closed ... do not miscompile". A silent empty exit-1 is arguably harder to diagnose than the segfault, since nothing points at layout at all.

## Fix

`collect_named_machine_deps` now descends into a named type's record field types, keyed off `pipeline.record_layouts` (already threaded into codegen). A `visiting` stack guards recursion rather than trusting acyclicity — a malformed or indirect layout terminates instead of overflowing the stack.

## Evidence

⚠️ **Load-bearing: the tests are revert-verified to discriminate.** With the record descent disabled and the tests kept, the container-first test **FAILS** while embedded-first still passes — so they test the fix, not merely its presence.

Both declaration orders are covered, each asserting the exact payload string so a zero-sized or misaligned payload fails rather than reading back an accidental value.

- `hew-codegen-rs` — 755 tests, RC=0
- all **159** `tests/hew/` files — 159 ok / 0 fail
- `cargo fmt --all --check` — RC=0 (caught a reflow pre-push; the lane that cost #2874/#2878 a red run)
- `cargo clippy -p hew-codegen-rs --all-targets -- -D warnings` — RC=0, zero findings
- pre-existing `machine_in_machine_test.hew` and `machine_in_record_test.hew` both still green

### DIFF
--- a/hew-codegen-rs/src/layout.rs
+++ b/hew-codegen-rs/src/layout.rs
@@ -281,31 +281,60 @@ pub(crate) fn host_data_layout_string() -> &'static str {
 /// canonical class-tagged monomorphization key (`mc$$control.Lifecycle$$i64$$`),
 /// while a payload field carries the nominal owner plus its concrete type
 /// arguments. Project the latter through [`hew_hir::machine_layout_key`] so
-/// both sides meet without discarding module or generic identity. Recurses
-/// through generic args, tuples, arrays, and slices: any of them can carry an
-/// embedded machine whose body must be set before the container is sized.
+/// both sides meet without discarding module or generic identity. Records use
+/// the parallel [`hew_hir::layout_key_for_named`] authority before the walk
+/// descends through their fields. Recurses through generic args, tuples,
+/// arrays, and slices: any of them can carry an embedded machine whose body
+/// must be set before the container is sized.
 fn collect_named_machine_deps(
     field_ty: &ResolvedTy,
     index_of: &HashMap<String, usize>,
+    record_fields: &HashMap<String, &[ResolvedTy]>,
+    visiting: &mut Vec<String>,
     out: &mut Vec<usize>,
 ) {
     match field_ty {
-        ResolvedTy::Named { name, args, .. } => {
+        ResolvedTy::Named {
+            name,
+            args,
+            builtin,
+            ..
+        } => {
             let machine_key = hew_hir::machine_layout_key(name, args);
             if let Some(&idx) = index_of.get(&machine_key) {
                 out.push(idx);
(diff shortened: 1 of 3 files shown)


slepp:
Merged behind #2879. The record-descent tests are genuinely discriminating — I ran them against the #2879 binary alone and the container-first case goes red, so the descent is load-bearing. I also ran the O0-vs-O2 differential gate on the stacked tip, which is the one I'd most want on a layout change, and it's clean.

One correction I applied to the PR body before merging: this doesn't close Gap 1 of #2864. Gap 1 is the enum-side classifier (`resolved_ty_references_machine`) which also doesn't descend records and is untouched here — I ran the issue's own example against this branch and it still fails the same way. What you fixed is a third shape the issue never names, which is a real gap and a good fix. #2864 stays open for the enum side.

Small followup for later: the record lookup is keyed by short name, while records are keyed by full name elsewhere in that file, so a `Wrap` and a `mod.Wrap` could be conflated.

PR #2887, opened 2026-08-11
refactor(build): centralize the build gate graph in xtask

slepp:
## Summary

Build and CI entry points now derive gate dependencies, required tools, platform setup, and timeout policy from xtask. The Makefile remains a thin facade, while compiled Hew test caching preserves identities, successful verdicts, and failure evidence across fixture changes. Review the gate dependency graph and the cache invalidation boundaries.

## Verification

- `cargo test --workspace --no-run`: workspace test targets compile
- `cargo xtask gate cutover-contract`: cutover and timeout contracts pass
- `cargo xtask gate freebsd-contract`: centralized FreeBSD setup contract passes
- `cargo xtask tools --gates ci --field tools`: the selected gate set resolves its required tools

## Out of scope

- Workflow lint and local workflow runner provisioning remain in a dependent change.


slepp:
I do not think this branch is safe to land as-is. The first deterministic failure is the sandbox parity contract: the Makefile no longer exposes the recipe shape the check requires, and the Linux preflight also reports four failed commands. Current main has since changed the preflight dispatcher in #2944 and hardened tool provisioning in #2927, while this branch conflicts across a 186-file, 72k-line build rewrite. With the dotted-path cutover imminent, I recommend restating only the still-wanted build changes as fresh, bounded PRs rather than attempting a wholesale rebase.

slepp:
Closing this in favour of #2984.

I measured the conflict surface against current main before deciding: 93% of this diff is checked-MIR corpus churn (a snapshot adding ~65k lines against 117 goldens removing about the same), and the substantive design is the ~2,240-line declarative artifact graph in xtask.

Rebasing the whole branch would have been actively harmful rather than merely expensive. Main has since made the diff dispatcher the standard preflight, moved release machinery to the workflow ref, and completed the dotted-path cutover; the four remaining conflicts were semantic ones in workflow and action files, and applying this wholesale would have discarded the newer fail-fast dispatcher guarantees and the migration corpus gate.

So the artifact graph is restaged cleanly on current main in #2984. The Makefile and workflow gate cutover from this branch is intentionally not carried forward — it needs its own design pass to preserve the current dispatcher behaviour, and I would rather it land deliberately than arrive as rebase fallout.

PR #2888, opened 2026-08-11
feat(build): provision workflow tools from selected gates

slepp:
## Summary

Depends on #2887 and should merge after it.

Selected workflow gates now declare `actionlint` and `act`, and shared setup installs the pinned tools on supported hosts. Workflow lint covers both workflows and local actions, while local workflow discovery uses the same gate-derived provisioning. Review the split between action-installed tools and Go-installed tools across hosted and FreeBSD setup.

## Verification

- `cargo test --workspace --no-run`: workspace test targets compile
- `cargo xtask tools --verify workflow-lint,workflow-local`: both pinned tools resolve and verify
- `cargo xtask gate workflow-lint`: workflows and local actions pass actionlint
- `cargo xtask ci-local --list`: the local workflow job is discovered through act

## Out of scope

- The local Linux workflow is listed but not executed as a containerized job on macOS.

### DIFF
--- a/.actrc
+++ b/.actrc
@@ -1,1 +1,2 @@
 -P ubuntu-24.04=catthehacker/ubuntu:act-22.04
+--container-daemon-socket -
--- a/.github/actionlint.yaml
+++ b/.github/actionlint.yaml
@@ -1,0 +1,4 @@
+paths:
+  .github/workflows/deploy-docs.yml:
+    ignore:
+      - 'constant expression "false" in condition'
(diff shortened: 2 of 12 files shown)


slepp:
This remains blocked on #2887. Its own first deterministic failure is gate reachability: `make workflow-lint` is reached by no CI workflow step. It should not land independently; if this provisioning work is retained, it needs a fresh base after the #2887 decision and an explicit CI edge for workflow lint.

slepp:
Closing this alongside #2887.

This depends on the Makefile and workflow gate contracts from that branch, which I deliberately did not carry forward when restaging the artifact graph into #2984 — they conflict with main's newer dispatcher behaviour and need their own design pass. So this branch cannot simply retarget the restaged foundation: the contracts it provisions against are not there.

The provisioning work itself is still wanted. It should be restaged once the gate and workflow integration is settled, against whatever those contracts turn out to be, rather than kept open against a base that no longer exists.

PR #2905, opened 2026-08-11
fix(mir): preserve conditional resource drop authority

### LINKED ISSUE #2897: Resource drop plan aborts on the documented close() pattern
`hew build` aborts with exit 134 on the resource lifecycle pattern v0.6 documents as the correct one.

```hew
import std::stream;

fn main() {
    if let Ok(sink) = stream.to_file("/tmp/repro.txt") {
        sink.write("rc1");
        sink.close();
    }
}
```

`hew-mir/src/lower/drop_plan.rs` panics with `build_lifo_drops invariant`.

This is the shape the release notes tell users to write: `impl Drop` is rejected, and handle-like types are expected to use `#[resource]` with an explicit `close(self)`. A compiler abort on the recommended pattern outranks the documented refusals, since a refusal is honest and an abort is not.

Found while migrating `applications/file-io.hew` in hew-lang/examples to rc1.

slepp:
Closes #2897.

Binding an owned resource in an `if let` aborted the compiler with a `build_lifo_drops` invariant failure in `hew-mir/src/lower/drop_plan.rs`, exit 134.

The release notes direct handle-like types to use `#[resource]` with an explicit `close(self)`, since `impl Drop` is rejected — so the compiler aborted on the pattern users are told to write.

The trigger is wider than the reported case. An owned `if let` binder aborts with an empty body and no close at all; `while let` and returning the resource abort the same way. An equivalent `match` was the nearest control that did not.

The invariant: every owned binding retained in the drop ledger must keep a concrete backend Place until drop elaboration. `if let` and `while let` restored the lexical mapping while leaving the owner registered, so the ledger held an owner with no place to drop. Deferred drop-place mappings now survive without influencing retain derivation, and projection binders are suppressed only when their carrier drop exists on the same edge.

## Verification

Closure proven in both directions rather than argued. The implicit-close path reports zero leaks at 3 and 50 frames with exact output counts; the explicit-close path completes 200 allocator-scribble iterations without abort, so there is no second close.

The fixture covers all five shapes: the empty-body reduction, explicit close, match, while-let, and a returned resource. The MIR lowering suite passes 566 tests, and the compiled Hew ratchet reports zero expected and zero actual failures.


slepp:
I rebased this onto current main and force-pushed `57732b6ab52f`. The old release-count refresh was dropped because #2962 already changed that topology and the refreshed counts no longer matched current output; the functional MIR fixes remain. The focused MIR and release-count differential tests pass, and CI is rerunning.

PR #2906, opened 2026-08-11
fix(tasks): reject cancelled value results

slepp:
A value-returning task cancelled between allocation and entry published a zeroed value as a successful result. The caller received `x=0 y=0` from a body that returns `Point { x: 11, y: 22 }`, with no indication anything had gone wrong.

An ABI-only zero looked complete because the wrapper published every return, and await consumed the payload bytes without first honouring terminal status. Completion was inferred from readable memory rather than established by the task.

Await now takes terminal status as its authority, and publication is suppressed when the task did not reach the point of producing a value.

## Behaviour at each cancellation window

- **Before allocation** — no result authority exists.
- **Allocation to entry**, and **mid-body** — publication is suppressed and the join surfaces cancellation.
- **After the value is published** — the value stands and is released by typed teardown.

## Verification

A genuine zero return must remain distinguishable from a cancelled task, since that is the pair a naive fix conflates. Both are covered: cancellation surfaces as cancellation, and a task that really returns zero still succeeds with that value.

An owned result belonging to a cancelled task is released exactly once, checked with the leak oracle. Runtime task tests pass 66/66, task IR emission 2/2, the forced-cancel compiled probe passes, and the vertical-slice suite is green. New runtime exports are classified and `make verify-ffi` matches.


slepp:
I rebased this onto current main and force-pushed `70361c6b4b46`, preserving the cancellation result authority and updating the FFI ownership ratchet against the current baseline. The focused codegen, runtime, forced-cancel, and FFI checks pass, and CI is rerunning.

PR #2907, opened 2026-08-11
fix(types): derive structural clone and equality from members

slepp:
Two operations a user reaches for immediately did not work.

`Option<i64>` had no `clone`. Two `(i64, string)` tuples could not be compared — the attempt failed with `IntCmp aggregate operands must be named structural-equality types` and emitted under-release warnings alongside it.

Both now follow one rule: an aggregate has a capability when every member has it, and is refused with a diagnostic naming the offending member otherwise. Refusals identify the member by path, such as `item.Some.1`. Tuples, `Option`, `Result` and records compose, including through nesting.

The two share that rule but not their lowering — equality goes through comparison thunks, clone through snapshot plans that carry drop obligations.

## Ownership

An affine or resource-bearing value never gains `Clone`. Manufacturing one would produce two owners of a single resource, which is the double free the release exists to prevent. Cloned aggregates own their contents independently.

The under-release warnings on the comparison path were not noise. `IntCmp` aggregate reads were being treated as ownership escapes, which removed the normal drop plans for the compared values. A comparison borrows its operands, and is now modelled that way.

## Verification

Accepted and refused shapes are pinned as fixtures. Leak oracles pass in both directions, so a cloned aggregate is released exactly once per owner and a refused clone never partially copies. The compiled Hew suite, core matrix, vertical slice and compiler tests all pass.


slepp:
The actual first failure is infrastructure, not the structural clone/equality tests: `wasm-pack` failed while downloading Binaryen 117. Current main contains `08b49afc8`, which prefetches Binaryen with retries. This needs a refresh onto current main and a clean CI rerun before review.

PR #2908, opened 2026-08-12
ci: verify Cloudflare docs deployment target

slepp:
## Summary

- record the verified Vocti Corp account for the `hew-docs` Pages project serving `docs.hew.sh`
- fail closed when generated stdlib documentation is empty or does not cover every source module
- update the release runbook with the account selector and current module-count check

## Verification

- generated v0.6.0-rc1 documentation for all 74 stdlib modules
- deployed the output to `hew-docs` and confirmed `docs.hew.sh` matches the generated index
- validated the guarded workflow with actionlint

## Out of scope

The workflow remains disabled until an operator adds the `CLOUDFLARE_API_TOKEN` repository secret with Pages edit access for the Vocti Corp account.

### DIFF
--- a/.github/workflows/deploy-docs.yml
+++ b/.github/workflows/deploy-docs.yml
@@ -9,9 +9,9 @@ on:
 jobs:
   deploy:
     # Disabled until CLOUDFLARE_API_TOKEN secret is configured in repository
-    # settings and the operator has verified the deploy command against the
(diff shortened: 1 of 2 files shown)


slepp:
Superseded by #2910. This branch's history embedded a Cloudflare account identifier, which would have been permanent once merged; #2910 reaches the same end state without it appearing at any point.

PR #2909, opened 2026-08-12
ci: enable the docs deployment

slepp:
Stacked on #2908.

The `CLOUDFLARE_API_TOKEN` repository secret is now configured with Pages edit access to the Vocti Corp account, which was the only thing #2908's guard was waiting for.

A tag push now regenerates the standard-library reference and deploys it to the `hew-docs` Pages project serving docs.hew.sh.

The validation step from #2908 derives the expected module count from the sources rather than pinning a number, so a partial generation fails the job instead of publishing a thinner site. On the current tree that is 74 source modules and 74 generated pages.

### DIFF
--- a/.github/workflows/deploy-docs.yml
+++ b/.github/workflows/deploy-docs.yml
@@ -8,11 +8,9 @@ on:
 
 jobs:
   deploy:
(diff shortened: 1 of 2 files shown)


slepp:
Superseded by #2910, which carries the same change on a clean history.

Issue #2207, opened 2026-06-25
Lift fail-closed carry restriction on owned non-record fields in functional-record-update


slepp:
## Summary

Functional-record-update `{ ..base, field: new }` over a fresh, uniquely-owned base destructively drops the overridden owned fields in place and **carries** (shallow-moves) the non-overridden owned fields out of the consumed base into the result. The carry lowering only soundly handles owned field types that are either:

- a single-pointer COW value with an inline drop symbol — `string`, `bytes`, `Vec<_>`, `HashMap<_,_>`, `HashSet<_>`; or
- a nested owned **record** (handled recursively via the record's clone/move machinery); or
- `BitCopy` / `View` (no destructor).

Every **other** owned field type in carry position is currently **fail-closed rejected** at `hew check` with:

```
E_NOT_YET_IMPLEMENTED: ... functional-update carry of owned non-record field ...
```

This issue tracks lifting that restriction so those types can be carried soundly instead of rejected.

## Field types that currently reject (carry position)

| Owned field type | Why no sound shallow carry today |
| --- | --- |
| closure / `fn`-typed field | The closure pair is a heap-boxed capture env with **no retain/release carry spine** in the funcupdate lowering. A shallow pointer copy makes the consumed base's drop and the result's drop both release the same env → double-free at teardown (reproduced under Guard Malloc), even when the closure captured only fresh locals. |
| tuple-of-owned (`(string, Inner)` …) | No recursive element-wise carry spine; shallow copy aliases the tuple's owned leaves. |
| `Option<owned>` / enum-with-heap-payload | No payload-aware carry spine; shallow copy aliases the active variant's owned payload. |
| `CancellationToken` / `Task` and other owned runtime handles | No carry spine; release semantics differ from the COW retain-set. |

The override (non-carry) position for these types is already handled or independently rejected; this issue is specifically the **carry** path.

## What "lift" requires

- **Closure / `fn` fields:** carry through the closure env's existing retain/release env handling (the same spine drop/clone uses), so the carried closure is transferred exactly once rather than shallow-aliased.
- **Tuple / `Option` / enum owned composites:** recurse into the composite and carry each owned leaf via its existing clone/move machinery (mirroring how the nested-record carry was made sound).

## Acceptance criteria

For each lifted field type:

[...]

PR #2916, opened 2026-08-12
fix(mir): apply one carry rule at every tuple nesting depth

gertybotbot:
Lifts the fail-closed NYI in #2207 for heap-owning tuple fields, under one carry rule that holds at every nesting depth.

## The rule

`carry_transfers_field_ownership` (`hew-mir/src/lower/expr.rs`) is the single authority for which field types the functional update's shallow `RecordFieldLoad` can carry out of the consumed base:

- `BitCopy` / `View` — no heap ownership; nothing to transfer.
- a single-pointer COW / handle leaf (string, bytes, `Vec`, `HashMap`, `HashSet`, `Generator`) — `project_field_inline_drop_symbol` is `Wired`, so the one allocation is released exactly once.
- an owned user record — the nested record's heap leaves ride its `RecordInPlace` spine.
- a heap-owning tuple **each of whose elements is itself carry-sound by this same rule**.

The tuple is not a transfer boundary in its own right. A tuple admits exactly the element types the same field admits bare, so there is no shape that is rejected as a field and accepted one layer down inside a tuple.

## Why the tuple is not its own boundary

The first cut of this change admitted any tuple carrying a drop obligation. That let an `Option<owned>` element through:

```hew
record T { pair: (Option<Inner>, i64), tag: string }
```

with `..b` compiled, excluded the consumed base from its `RecordInPlace` drop, and left nothing owning the optional's payload. `leaks --atExit` counted exactly one leaked allocation per frame — both when the result escaped its constructing frame and when base and result died in the same scope — while the identical record shape built *without* a functional update leaked nothing. `Option<string>` leaks the same way, so it is the optional's shape that has no carry, not the payload's.

Heap-payload user enums in a tuple measure clean today, and are still rejected: lifting the enum carry should lift the bare and in-tuple forms together, under one soundness argument and one leak oracle, rather than leaving a bare-rejects/tuple-admits split whose only justification is that nobody has measured the bare form yet.

## What now carries

Tuples of string, bytes, `Vec`, nested tuples, and owned records — `(string, i64)`, `(string, (string, i64))`, `((string, i64), (string, i64))`, `(Vec<string>, i64)`, `(Inner, i64)`.

## Verification

- `cargo test -p hew-cli --test funcupdate_consume_semantics` — 51 passed / 0 failed. Every accept fixture reads a carried leaf back out, so a corrupted carry cannot pass on the overridden field alone.
- `cargo test -p hew-cli --test funcupdate_owned_field_drop_leak_oracle` — 22 passed / 0 failed, including three new macOS `leaks(1)` slope oracles for the tuple carry: `(string, i64)` in escape drop order, `(string, i64)` in same-frame drop order, and `(Inner, i64)`. Zero leaks at both 3 and 50 frames in every case.
- `make preflight` — clean.

The two drop orders are measured separately because they exercise different halves of the transfer: escape order (the result outlives the frame that consumed the base) exposes a carry that never hands the tuple over; same-frame order (result released first, base's teardown immediately after) exposes a transfer that failed to exclude the base.

Boundaries pinned by named tests: `reject_carry_tuple_of_option_field`, `reject_carry_tuple_of_enum_field`, `reject_carry_option_of_owned_field`, `reject_carry_closure_field`, `reject_carry_non_heap_tuple_field_is_conservative`.

## Out of scope

Bare `Option` and enum fields, closures, `fn` values, and handles remain fail-closed, as do their in-tuple forms. A tuple with no heap fields (`(i64, i64)`) still rejects — conservative rather than wrong, and pinned by a named test.

Refs #2207


slepp (changes requested):
The lowering change itself held up under everything I threw at it — Guard-Malloc with scribble/guard-edges, a 100k-iteration reassign-loop carry, and 50k/100k/200k frame sweeps with zero RSS slope. Two things need addressing before this can land:

1. **Admitted scope is wider than the stated scope.** `ty_is_heap_owning_tuple` is a pure shape predicate (`Tuple(_)` + any drop obligation), so the carry now also admits a tuple *containing* an `Option` payload (`record O { maybe: (Option<Inner>, i64), tag: string }`) and a tuple containing a heap-payload user enum — both pass `hew check` today, while the bare-field forms are still rejected. The description says Option/enum are untouched, and there's no fixture or soundness argument for either newly admitted family. My probes for both ran clean under the malloc oracles, so I don't have evidence of unsoundness — but they need fixtures and a stated argument, or the predicate needs narrowing to what the PR argues for.

2. **The soundness comment names a prover that doesn't run on this path.** `derive_tuple_composite_drop_allowed`'s candidate set is tuple *bindings*; in these fixtures the carried tuple is a record field, never a binding, so that prover has no candidate. What actually protects the base is `derive_owned_record_drop_allowed`'s escape rule — the tuple-typed `RecordFieldLoad` dest counts as heap-owning and its escape into the result's `RecordInit` drops the base root from `RecordInPlace`. The mechanism is real; the committed explanation attributes it to the wrong prover. Please repoint the comment at the escape rule.

One smaller note: `MALLOC_CHECK_`/`MALLOC_PERTURB_` are glibc-only, so it's worth stating the platform in the test plan — the suite's own module doc names Guard-Malloc + `MallocScribble` as the authoritative oracle, and that's what I verified with.

The `(i64, i64)` conservative-reject follow-up you propose sounds right; happy to see that as a separate PR.

slepp:
The current head is green, and the requested scope and soundness corrections are present in `42ca548ee`. I consider the code ready for a fresh review verdict; the existing changes-requested state is the remaining gate.

slepp:
Superseded by #3020, which carries this change rebased onto the current ownership seam (after #3012 and the projection-read work); the fixtures here still fail to compile on main without the rule, so the fix stands and continues there.

PR #2917, opened 2026-08-12
fix(mir): drop owned method receiver temporaries

### LINKED ISSUE #2885: Method call on a projection temporary leaks the whole Vec
Calling a method directly on a collection projection leaks the projection Vec and everything it owns.

```hew
let n = m.keys().len()      // leaks the Vec and every key it owns
let e = m.entries().len()   // same, and strands two owned elements per pair
```

Under `leaks --atExit` with a 3-entry map of dynamically built strings, `entries().len()` reports 8 leaks / 368 bytes rooted at `hew_vec_new_with_elem_size` under `hew_hashmap_entries_layout`, retaining all 6 owned strings. `keys().len()` leaks 5 / 240 bytes the same way.

This is not specific to `entries()` — it reproduces with `keys()` on main and predates that method. `entries()` only widens the blast radius, since each leaked pair strands a key and a value rather than one element.

Only method-call-on-temporary leaks. All of these are clean:

```hew
for k in m.keys() { }        // clean
takes_ownership(m.keys())    // clean
let ks = m.keys()            // clean
```

So the gap is in drop planning for a temporary that is consumed by a method call rather than bound, iterated, or passed. The receiver is released, but the projection Vec created for the call is never scheduled for release.

Reported during review of #2866 and verified against a base build predating it.

slepp:
Closes #2885.

Calling a method directly on an owned temporary leaked the temporary and everything it owned: `m.keys().len()` leaked the projection Vec and every key.

Iteration, consuming arguments and bindings all transfer the published owner, but a borrowing method receiver lacked the equivalent typed-owner completion — its provisional generation stayed outside drop planning, so no release was ever scheduled.

The fix is general to owned temporaries used as method receivers, not specific to HashMap projections.

## Verification

All proofs use dynamically built strings, since literals are immortal and can observe neither a leak nor a double free. The memory oracle reports zero leaks at low and high frame counts, and 200 frames exit cleanly under the poisoned allocator, so there is no second release. The `for`, argument-position and `let`-bound shapes still release exactly once. MIR unit tests pass 750; the full compiled Hew suite passes including the four new fixtures.


slepp:
Closing in favour of #2986, which contains this PR's entire diff plus the piece that makes the chained shape actually work.

The delta is load-bearing rather than cosmetic. With this PR's MIR fix present but #2986's codegen catalog entries reverted, `make_vec(7).clone().len()` fails closed:

    E_NOT_YET_IMPLEMENTED: fail-closed: typed crash cleanup for Local(47) (Vec<string>)
    has no lifecycle hook for its producer

So merging this alone would ship a compiler that hard-fails an ordinary receiver chain. #2986 adds the missing lifecycle hooks and pins the shape with a fixture and a leak oracle measuring zero at both frame counts, against 84 and 1400 nodes on the previous lowering.

The ownership work here is preserved in full — it is the foundation #2986 builds on, not something being discarded.

PR #2920, opened 2026-08-12
fix(mir): allow moving owned payloads out of direct call-result matches

### LINKED ISSUE #2914: json.try_parse cannot be matched: heap-owning payload rejected as possibly aliasing caller storage
Matching on `json.try_parse` fails:

    E_MIR_CHECK: cannot move the heap-owning payload ... out of a match on a
    value that may alias caller storage

So the modern JSON parsing API cannot be used in the natural way — matching its `Result` to get either the value or a typed error.

The workaround is the legacy path: `json.parse` plus a `type_of() < 0` check. That works, but discards the native JSON error detail that `try_parse` exists to provide, so callers lose the ability to report *why* a document was invalid.

This is the shape a user reaches for first, on an API the documentation presents as current. Found while parsing JSON Lines log records, where per-line error detail is the whole point.

Reported from dogfooding rc1.

slepp:
Closes #2914.

Matching a `Result` returned by value refused to move its heap-owning payload:

    E_MIR_CHECK: cannot move the heap-owning payload ... out of a match on a
    value that may alias caller storage

So `json.try_parse` could not be used the natural way, and callers fell back to the legacy parse path that discards native error detail.

A completed call's ownership facts identify the result as a fresh owner, so direct matches now trust them and admit the move. The fix is general to owned call results, not JSON-specific.

Borrowed and forwarded results keep the refusal — a match on a parameter alias still rejects, pinned by a reject fixture, so the aliasing rule is narrowed only where the scrutinee is provably fresh.

## Verification

The vertical slice passes including the new `json_try_parse_direct_match_payload` accept fixture (success and error arms carrying payloads) and the `match_result_param_alias_payload` reject fixture. The compiled Hew ratchet passes. MIR lowering-call provenance tests extended.


slepp:
The only red check is the macOS periodic-timer quiescence test: the handler never fired before shutdown. Current main contains `befa21ed4`, which fixes that exact test by awaiting tick delivery and documenting the shutdown semantics. This needs a refresh onto current main and a clean rerun.

PR #2923, opened 2026-08-12
feat(concurrency): support multi-statement fork bodies

### LINKED ISSUE #2912: A multi-statement fork body is unsupported, so accepted connections cannot be handled concurrently
Building a TCP key-value service, the natural shape is to accept a connection and move it into a concurrent child:

```hew
fork {
    let conn = listener.accept();
    handle(conn);
}
```

The compiler does not support a multi-statement `fork { ... }` body, so a connection cannot be moved into a concurrent child. The server was forced to process requests **sequentially**, one connection at a time.

That is a hard limit on any network service written in Hew: concurrent connection handling is the defining requirement of a server, and the language advertises structured concurrency as a strength.

Related, and hit in the same program: opaque handles cannot be `receive fn` parameters, failing with `E_OPAQUE_MESSAGE_PAYLOAD`. So a connection also cannot be handed to an actor to own. Together these leave connection ownership handler-local with no supported concurrent handoff.

Reported from dogfooding rc1.

slepp:
Closes #2912.

A `fork { ... }` body was limited to a single direct call: the parser accepted full blocks, HIR rejected multi-statement and non-call bodies, and MIR lowered only the direct-call shape. A server accepting a connection could not move it into a concurrent child, forcing sequential handling.

Multi-statement fork bodies now lower through the scoped-children machinery. A value created in or moved into the body is owned by the child, and its drop obligation follows the child through completion, crash, and cancellation while suspended — all three paths release exactly once, proven by the lifecycle oracles. A borrow of parent state that would dangle across the fork boundary is refused with a diagnostic naming the binding, pinned by a reject fixture.

Concurrency is proven rather than asserted: the accept fixture starts two slow children and observes both begin before either completes, under O0 and O2.

## Verification

Checker, HIR task-gate, MIR cancellation-scope, codegen structural, and runtime lifecycle suites pass. The O2 differential reruns 1,328 outcomes identically. The compiled Hew ratchet passes. Suspending expressions directly inside fork-body closure shims remain outside the current codegen ABI; a coroutine-driving task wrapper is the follow-up before admitting them.


slepp:
I rebased this onto current main and force-pushed `385f46cf397f`. The branch contained two patch-identical copies of the feature and inventory commits; the rebased history keeps one copy of each. The focused HIR, MIR, codegen, and runtime tests pass, and CI is rerunning.

PR #2924, opened 2026-08-12
feat(format): add structural aggregate rendering

slepp:
Implements the ratified aggregate-formatting decision (V06-06). Stacked on #2919 — the alias-scoped Display customization depends on alias identity.

Aggregates could not be printed at all: interpolating a HashMap, record, tuple, Option or Result was refused, so debugging a collection meant writing Display by hand.

`f"{value:?}"` renders any aggregate structurally: records as `Point { x: 3, y: 4 }`, tuples, `Option`/`Result`, `Vec`, `HashMap`, and nesting thereof. Opaque and resource members render as type-plus-identity rather than raw bits.

Ordinary interpolation `f"{value}"` is unchanged — it still requires `Display`, so internal structure never leaks into user-facing strings by accident. An explicit `Display` on a concrete type or a named alias always wins over the structural rendering.

The member walk reuses the structural equality/hash layout machinery rather than adding a third walker.

## Verification

Exact-output fixtures pin the format per shape. The ownership oracle proves rendering has zero ownership effect: unchanged identity, exactly-once resource close, zero leaks. Ratchet and vertical slice pass.


slepp:
This needs a focused fix before rebasing. The first product-local failure is the checked-MIR coverage guard: `hew_structural_format` is no longer covered and must gain a fixture or an explicit justified pin. The Windows and Linux LLVM download failures were transient infrastructure errors. #2919 is now on main, so the repair should rebase onto current main, restore the coverage contract, and rerun the ownership and rendering checks; if it lands after the dotted-path cutover, its Hew fixtures also need migration.

PR #2935, opened 2026-08-13
docs: fill hew test, modules/imports, pub const, and wire/crypto/path doc gaps

slepp:
## Why

A maintainer-side review of a real port project (goobers-hew) surfaced a set of docs gaps where a diligent user hand-rolled things that already exist because the docs never mentioned them: `hew test` (the entire harness — discovery, `#[should_panic]`, `#[ignore]`, `#[serial]`, `--filter`, `--format junit`, `--timeout`) had zero coverage anywhere in `docs/`; `pub const` had zero mentions despite working cross-module; the crypto module's own example built a `bytes` buffer by hand instead of using `to_bytes()`; the `std::encoding::wire` module page described only removed legacy helpers instead of `#[wire]`; and `path.absolute()` doesn't warn that it leaves `..` unresolved.

## What

- **language guide**: new "Testing — `hew test`" chapter; new "Modules and imports" chapter (`import "file.hew"`, `src::` dotted imports, selective imports, the current `machine`-as-path-segment restriction); `pub const` documentation; lint-suppression documentation (`// hew:allow(...)`, `-A`/`-W`/`-D`) with the rationale for why `dead_code` still fires on `pub` functions; a rationale note on `process.run` (shell) vs `process.run_argv` (no shell); a rationale note on why a machine's `default` arm is all-or-nothing per machine, not per state; a one-line "coming from Rust/Go" nit on `type` vs `struct`.
- **std/crypto/crypto/crypto.hew**: module-doc example now uses `"A".to_bytes()` instead of hand-built `bytes::new()` + `.push()`.
- **std/encoding/wire/wire.hew**: module doc (source of the `docs.hew.sh/std.encoding.wire` page) now leads with `#[wire]`, `@N` tags, `to_json`/`from_json`, `hew wire check`, and states the current same-module-only scope of the codec, instead of describing only removed legacy HBF helpers.
- **std/path.hew**: `absolute()` doc now states plainly that it does not resolve `..`, with a warning against using it alone for containment checks.

## Test

Every new/changed `docs/hew-language-guide.md` fence was verified with `scripts/extract-doc-fences.sh` against `target/debug/hew` built from this branch — 305 fences extracted, ratchet unchanged (87 expected failures, all pre-existing; zero new failures). Every behavioural claim in the new prose (test-runner flags, `pub const` cross-module access forms, `#[wire]` round-trip, `crypto.sha256("A".to_bytes())`, `process.run` vs `process.run_argv`, `path.absolute()` leaving `..` unresolved, the `machine`-keyword import restriction, `-A dead_code` / `// hew:allow(dead_code)` suppression) was run manually against the built `hew` binary before being written down.

## Quality Checklist
- [x] No new `.ok()?` or `unwrap_or_default()` in codegen without `// JUSTIFIED` comment
- [x] New allocations have cleanup paths for sync, async, and actor shutdown contexts
- [x] Serialization changes include round-trip encode/decode tests
- [x] New runtime features have a WASM implementation or an exact `// WASM-TODO(<stable-backlog-id>):` marker whose id is defined in `wasm-capability-manifest.toml`, and new `hew_*` exports are classified in `scripts/jit-symbol-classification.toml`
- [x] No duplicated logic — checked for existing helpers before adding new ones

### DIFF
--- a/docs/hew-language-guide.md
+++ b/docs/hew-language-guide.md
@@ -56,6 +56,7 @@ for the documented resolver precedence.
 - Negate a bool with `!`: `!x` is `true` when `x` is `false`.
 - `break` and `continue` work in both `loop {}` and `while` loops.
 - Iterate all HashMap keys with `.keys()` → `Vec<string>`; all values with `.values()` → `Vec<V>`.
+- Coming from Rust/Go/TS: there is no `struct` keyword — `type` declares both plain data records and generic containers, so one keyword covers what Rust splits into `struct`/`enum` shapes. Writing `struct` is a parse error with a `write \`type Name { ... }\`` hint.
 
 ## Primitives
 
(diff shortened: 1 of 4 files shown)


gertybotbot (commented):
Two small accuracy fixes in the new `hew test` section:

- The `_test.hew` / `tests/` layout restriction applies to directory traversal only. An explicitly passed `.hew` file is scanned regardless of its name or location, so `hew test scratch.hew` does discover `#[test]` functions.
- `--allow-empty` does not cover a path that does not exist: path canonicalization rejects it before discovery. The example should use an existing generated/optional directory that may contain no tests, rather than one that “may not exist.”

PR #2939, opened 2026-08-13
fix: close port-review friction smalls

slepp:
## Summary

Small fixes surfaced by reviewing a real port of an external project onto Hew:

- `path.normalize()`: new pure-Hew lexical path cleaner (Go `filepath.Clean` semantics); doc-warns `path.absolute()` does not resolve `..` segments.
- Credit selective imports (`import mod::{Thing};`) as used when the imported name appears via record/variant construction, not just annotation position — fixes a false-positive unused-import warning.
- `hew build --release`/link driver: find the debug `libhew.a` archive under a custom `CARGO_TARGET_DIR`, not just the literal `target/` path.
- Dead-code lint: for a `pub` function, suggest suppression (`-A dead_code` / `// hew:allow(dead_code)`) instead of a `_`-prefix rename, since renaming a public symbol is not the fix a caller wants.
- Bump the structural-authority lint TSV counts to match the above changes.

## Deferred

One finding from the same source is not in this PR: a false-positive leak advisory (`E_MIR_CHECK ObligationUnderReleased`) that `std::iter::any`/`filter`/`collect` combinators emit at every use site from their own monomorphized bodies, even though the runtime already discharges correctly. It needs its own change in the MIR drop-plan/obligation-balance modelling and carries a distinct, real leak finding worth fixing alongside it — tracked separately, not folded into this PR.

## Verification

- `make lint`: pass (exit 0)
- `make test-hew-ratchet`: pass, 0 expected failures, 0 actual failures

## Out of scope

- The false-positive leak advisory noted above (deferred).
- Everything else that came out of the same review — separate PRs.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -556,7 +556,45 @@ jobs:
         if: env.RUN_CODE_PATH != 'true'
         run: echo "No code changes detected; there are no compiled Hew shards to aggregate."
 
+      - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10  # v6.0.3
+        if: always() && env.RUN_CODE_PATH == 'true'
+
+      - name: Download compiled Hew bundle
+        if: >-
+          always() && env.RUN_CODE_PATH == 'true' &&
+          needs.compiled-hew-linux.result == 'success'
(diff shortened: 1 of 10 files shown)


slepp:
I rebased this onto current main and force-pushed `44a2a07d7739`, preserving the path, selective-import, target-directory, and diagnostic fixes against the newer authority counts. The focused type-checker, link, structural-authority, and path tests pass, and CI is rerunning.

PR #2943, opened 2026-08-13
test: add consume-param transfer regression coverage

slepp:
## Summary

The consume-parameter builder mechanism and the value-tree close()/free() disposal contract ship on main without dedicated regression guards. This adds a leak oracle and vertical-slice fixtures covering both the canonical close() and the consuming free() alias, and pins the UseAfterConsume rejection surface. Test-only — no compiler or stdlib changes.

## Verification

- Leak oracle: 2/2 passing, 0/0 leaks
- Vertical slice (`tests/vertical-slice/run.sh`): green, including the new consume-param builder accept fixture and use-after-move rejection fixture
- Ratchet: 0/0
- Preflight: ready-to-push

## Out of scope

- No changes to the consume-parameter mechanism or disposal contract themselves — this only pins the behaviour already shipped on main.
- No compiler or stdlib changes.


slepp:
The actual failure is the fuzz oracle compiling the new accept fixture as a standalone runnable program: its declared native symbols `hpush` and `hclose` have no linked implementation. The fixture needs either a self-contained native harness or explicit frontend-only classification before the oracle can pass; then the leak and vertical-slice coverage should be rerun.

PR #2945, opened 2026-08-13
fix(cli): resolve test imports from the project root

slepp:
## Summary

`hew test` compiled each harness from a staged copy in a temporary directory, so the compiler inferred that directory as the project root and no cross-file import resolved inside any test. Harnesses now compile in memory against the original test path with the project root threaded through, making test-time import resolution identical to `hew run`. A multi-file fixture and a string-path regression pin the behaviour.

## Verification

- `cargo test -p hew-cli --test test_runner_e2e`: 30 passed, 0 failed
- `cargo test -p hew-cli --bin hew test_runner`: 27 passed, 0 failed
- Full CLI preflight: pass
- Compiled ratchet: 0 new failures
- Preflight: ready-to-push

## Out of scope

- Test-runner discovery rules and layout conventions are unchanged; only where harnesses compile from is different.
- Package-level import surface for `*_test.hew` peers is handled separately and untouched here.

### DIFF
--- a/hew-cli/src/main.rs
+++ b/hew-cli/src/main.rs
@@ -1038,32 +1038,6 @@ fn compile_build_binary_with_hew_lib(
     }
 }
 
-pub(crate) fn compile_test_binary_with_paths(
-    input: &Path,
-    output_path: &Path,
-    paths: &NativeBuildPaths,
-    extra_libs: &[String],
-) -> Result<(), ()> {
-    let target = target::TargetSpec::from_requested(None).map_err(|error| {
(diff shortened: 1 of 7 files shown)


slepp:
The red check is deterministic: `hew-check-all` compiles `project_root_test.hew` standalone from its `tests/` directory, so `test_project_root::answer` cannot resolve. Current main still reproduces the underlying `hew test` project-root failure, so the feature is not superseded, but this branch now conflicts with #2958. The fix needs a rebase that preserves the package-FFI test path plus a corpus-safe fixture arrangement.

PR #2954, opened 2026-08-14
fix(types): model scope tails as subsuming values

### LINKED ISSUE #2898: Internal checker error on the documented scope { spawn ... } form
`hew build` reports an internal checker error rather than type-checking or diagnosing this structured-concurrency form:

```hew
actor Worker {
    receive fn run() {
    }
}

fn main() {
    scope {
        let worker = spawn Worker();
        worker.run();
        0
    };
}
```

    checker produced-value graph is incomplete:
    identity dependency changes type from I64 to Unit

Either the form is valid and should check, or it is not and should get a user-facing diagnostic. An internal graph error is neither.

Found while migrating `actors/structured-concurrency.hew` in hew-lang/examples to rc1.

gertybotbot:
## Summary

- model `scope` and `scope after(...)` tails as subsumed values rather than identity edges
- allow a scope's trailing expression to differ from the outer scope's `Unit` type
- cover the reported `spawn` + actor call + trailing value form with a regression test

## Root cause

The produced-value graph treated the scope node and its tail as type-identical. A scope expression is always `Unit`, while its tail is evaluated for side effects and may have another type, so the validator rejected valid code as an internal graph error.

## Verification

- `cargo test -p hew-types scope_body_with_spawned_call_and_trailing_value_checks_cleanly -- --nocapture`
- `cargo test -p hew-types native_scope_no_platform_error -- --nocapture`
- `cargo test -p hew-types`
- `cargo fmt --all --check`
- reported reproducer via `cargo run -p hew-cli -- check`: `OK`

Closes #2898.

### DIFF
--- a/hew-types/src/check/expressions.rs
+++ b/hew-types/src/check/expressions.rs
@@ -8280,7 +8280,13 @@ impl Checker {
             }
             Expr::Scope { body } | Expr::ScopeDeadline { body, .. } => {
                 body.trailing_expr.as_deref().map(|tail| {
-                    ProducedValueDependency::Identity(SpanKey::in_module(
+                    // A scope expression is always Unit at the outer surface,
+                    // but its tail expression may produce any type and still be
+                    // evaluated for side effects. Model that as a subsuming
+                    // publication rather than an identity edge so the checker
+                    // does not require the scope node and its tail to have the
+                    // same type.
+                    ProducedValueDependency::Subsumes(SpanKey::in_module(
                         &tail.1,
                         self.current_module_idx,
                     ))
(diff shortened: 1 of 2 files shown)


slepp (approved):
Reviewed the produced-value graph end-to-end, including the recent dyn join-coercion paths. `Subsumes` is the right edge here: it preserves the child ownership/provenance while keeping the Unit-valued scope as the materialized publication, without weakening the `Identity` type-equality invariant. The regression is focused and the checker suite is clean.

slepp:
Thanks, Gerty — I particularly liked that this fixes the graph model at its semantic boundary rather than special-casing the actor call. Nice tight regression too.

PR #2989, opened 2026-08-20
fix(runtime): isolate actor faults

slepp:
A program whose actors panicked exited 0. The panic reached stderr but never the exit code, so a run in which two actors died was indistinguishable from a clean one to any caller reading $?. A shell pipeline, CI job or supervisor gating on exit status read a crashed run as success.

An actor panic now produces a nonzero process exit. A clean run still exits 0 — the fix reports the fault rather than making every run look failed.

Fire-and-forget sends to a stopped local actor are now no-ops. Previously a send to an already-crashed actor panicked the *sender*, and since those sends return nothing the caller had no result to guard with, so one dead client cascaded into the actor broadcasting to it. A broadcast loop now survives a dead peer without pre-checking liveness on every send.

Found by dogfooding: a developer building an actor chat room hit both on an ordinary program. Covered by a fixture that crashes one client mid-broadcast and asserts the exit code in both the faulting and the clean case.


slepp:
I recorded the fault at the Running->Crashing transition before any wake, made exit() quiesce through the shared helper before reading the exit-status authority, and added the host executable suffix to compiled binaries.

PR #3000, opened 2026-08-21
fix(mir): transfer Rc ownership at value-aggregate ingress

slepp:
## Summary

An `Rc` or `Weak` handle moved into a value aggregate — a tuple element, an
`Option`/`Result` payload, a record field, a machine payload, an array element —
was byte-copied into the aggregate with no retain, while the source binder kept
its own `DropKind::RcRelease` obligation. Both owners released the one strong
count and the second release underflowed it, so the runtime aborted with
`Rc double-free: strong refcount is already 0`. No `clone` was needed to reach
it:

```hew
type Node { id: i64; }

fn main() -> i64 {
    let shared: Rc<Node> = Rc.new(Node { id: 7 });
    let pair: (Rc<Node>, string) = (shared, "tag");   // aborts at exit
    pair.1.len() - 3
}
```

Every other owning class already answers this question. `Vec`, `HashMap`/`HashSet`,
`bytes`, CoW `string` and owned records each gate their scope-exit release on a
per-class escape-scan allow-set that removes a handle proven to have escaped into
an aggregate — which is why `let pair = (v, "tag")` over a `Vec` was always
correct. `Rc`/`Weak` have no such allow-set: the `AffineResource` arm of
`build_lifo_drops` emits their release unconditionally. So the transfer is now
recorded where it happens, in the single aggregate-ingress funnel
(`alias_moved_owned_operand`), and every aggregate shape is covered by that one
edit.

What it records is the path-sensitive drop-flag store a by-value
`Use { Consume }` already emits. The flag is minted for every `Rc`/`Weak` local
at its introducing `let`, zero-initialised so the init dominates the store and
every exit, and codegen already gates the release on `flag == 0`. A conditional
ingress therefore stays exactly-once on both arms: the arm that placed the handle
skips the binder's release, the arm that did not keeps it.

The dataflow lattice is deliberately untouched. `AliasedIntoAggregate` remains
`Live` for every escape-scan, alias and drop reader, so the fail-closed aggregate
refusals keep their current posture for the classes that do have allow-sets, and
using the handle after the move is still `UseAfterConsume` at check time. Only
the runtime transfer record is added.

**What is now guaranteed:** moving an `Rc`/`Weak` into a value aggregate transfers
its strong count exactly once, on every control-flow path — no over-release, no
leak — for tuple, `Option`, `Result`, record, nested record, machine payload and
array-element ingress, in both scope-exit release orders.

## Verification

- `cargo test -p hew-mir` — 1724 pass
- `cargo test -p hew-codegen-rs` — 848 pass
- `cargo test -p hew-cli --test rc_aggregate_ingress_ownership_oracle` — 22 pass
  (new; every one of the 11 shapes aborts on `main`)
- `scripts/ll-corpus.sh verify` — 15 fixtures × 2 targets byte-identical, no IR
  delta. The corpus carries `rc_weak_lifecycle` and `rc_weak_graph`, neither of
  which places a handle into a value aggregate, so the change is inert for them.
- `make test-hew-ratchet` — PASSED
- `make test-vertical-slice` — 567 pass
- `make lint` — clean

The new oracle pins both directions per shape. Over-release runs the probe under
the poisoned allocator (`MallocScribble`/`MallocPreScribble`/`MallocGuardEdges`),
where a release against an already-zero count kills the process, so a clean exit
is the assertion. Under-release takes the per-iteration leak slope with an exact
line-count work witness — that is the half that fails if the transfer record is
not path-local, because the arm that never placed the handle would then leak one
allocation per frame. Both scope-exit release orders are covered per shape:
aggregate-first (both locals of one frame) and binder-first (the aggregate is
returned and the caller releases it).

## Out of scope

`clone`-ing a `Vec<Rc<T>>` in a function that picks between the clone and the
original with a `match` leaks. Measured on this branch and on `main`, and it is
NOT an `Rc` defect: the identical shape over `Vec<i64>` leaks the same way, and
so does a two-`Vec` version with no `clone` at all. The losing arm's `Vec` is
never released; the `Rc` inside it is only a passenger. Separate defect, separate
lane.

`clone` over a tuple or an `Option` payload is not supported on `main` at all
(`no method 'clone' on '(Rc<Node>, string)'`), so the structural-clone surface is
untouched here.


slepp:
Rebased onto the merge tip; both blockers and both advisories are closed at `7067a72`.

## B1 — array-literal ingress

Reproduced, but only in its **conditional** form. Straight-line `[shared]` is
clean on every spelling I tried (single element, two elements, `Rc<i64>`,
`Weak`, `Option<Rc<T>>`, `Vec<Holder>`, returned, in-loop) — all exit 0 with 0
leaks. `b1_cond_array` aborts:

```hew
match flag {
    true => { let v: Vec<Rc<Node>> = [shared]; v.len() }
    false => 0,
}
```

The mechanism is not a missing funnel call. That route
(`consume_owned_vec_move_array_element`) records the move as a dataflow
`Consume`, which IS sufficient straight-line — `filter_drops_by_state` excludes
a `Consumed` binding. On a join with an arm that did not move, the meet is
`MaybeConsumed`, which the same filter admits as **live**, so the binder's
guarded release fires on the arm that already handed the handle to the Vec,
guard still reading 0.

Fixed at `mark_binding_moved`, which is already the general consume seam and
already sets two sibling path-sensitive flags there on exactly this reasoning
("every consume site, not just the primary `Use{Consume}` lowering"). The affine
release flag was simply missing from that list. Putting it there rather than at
each call site is what closes the class: a transfer surface cannot be added
without funnelling through this seam, so it cannot be added without the transfer
being recorded — no per-site enumeration to keep in sync.

Ingress sites, each probed in its conditional form:

| site | route | verdict |
|---|---|---|
| tuple / record / `Option` / `Result` / nested record | aggregate-alias funnel | fixed, first commit |
| array literal, owned-move element push | consume seam | **was the B1 abort**, fixed |
| array literal, other push callees | aggregate-alias funnel | clean |
| user `Vec.push` | consume seam | clean |
| function argument by value | consume seam | clean |
| closure capture by move | consume seam | clean |
| returned binding / returned aggregate member | consume seam | clean |
| machine state payload | aggregate-alias funnel | no over-release; see below |
| `var` reassignment | generation boundary | **was B2**, fixed |
| actor message payload | — | unreachable: `cannot send Rc<Node> to actor: type is not Send` |

Machine payload does not over-release. It leaks one state value per frame
because the machine local has no scope-exit composite drop — the compiler emits
its own `ObligationUnderReleased` advisory naming it, and the identical machine
with a `string` payload leaks at the same rate. That is a machine-composite drop
gap, not a refcount defect, so the oracle pins the over-release half only and
says why.

## B2 — move-then-reassign

Reproduced exactly: 16 leaks / 1280 bytes over 16 frames, now 0.

A `var` reassignment is a generation boundary where two obligations meet. The
old generation must be released iff this frame still owns it — which it does not
when the handle was moved into an aggregate, and that is what the flag records —
so the release is gated on `flag == 0` rather than on the binding's static
`owned_locals` disposition, which an aggregate ingress deliberately leaves
untouched. The unguarded static arm would have released a handle the aggregate
now owns, the same double-free one step along. The new generation is owned
outright, so the flag resets to 0 after the store and the scope-exit disposition
is restored. This mirrors the arm `overwrite_guard_flags` already runs, and keeps
the aliasing posture — `shared = shared` skips the release rather than freeing
storage the incoming value points at.

**Adjacent gap, unchanged and pre-existing:** `var shared = Rc.new(a); shared =
Rc.new(b);` with no aggregate anywhere still leaks generation `a` (16/1280 over
16 frames, same on `main`). `project_field_inline_drop_symbol` has no `Rc`/`Weak`
arm, so `emit_local_overwrite_release` takes its documented fail-open path.
Adding that arm reaches the Vec-element, record-field and generator-yield release
pickers, so it is not something to land inside a double-free fix.

## A1 / A2

Oracle is now 31 tests. Added array-element ingress (straight-line and
conditional), machine payload (over-release half, with the measured reason the
leak half is excluded), reassign-after-ingress, and reassign-in-loop. The
conditional shapes assert their exact computed total instead of a sign check, so
an arm that silently stops running fails rather than passes.

## Gates

- `cargo test -p hew-mir` — 1724 pass
- `cargo test -p hew-codegen-rs` — 848 pass
- the oracle — 31 pass
- `make test-hew-ratchet` — PASSED, 0 actual / 0 expected failures
- `make test-vertical-slice` — 567 pass
- `make lint` — clean
- `scripts/ll-corpus.sh verify` — 15 × 2 byte-identical after one explained
  recapture

### The IR delta

One instruction, `rc_weak_graph` only, identical on native and wasm32:

```
store i64 1, ptr %local_9, align 8
```

`%local_8` is `weak`, the handle `build()` returns; `%local_9` is its
path-sensitive release flag, zero-initialised at the introducing `let`. The store
lands immediately before `weak` moves into the return slot — the transfer record
now set on the consume seam, which `mark_returned_binding_moved` routes through.
No release, call, block or branch changed. On this straight-line return the store
is redundant (the dataflow already marks `weak` `Consumed` at that exit, so the
`WeakRelease` was excluded either way, and both stores to the flag are dead
here); it is load-bearing on the conditional analogue. Golden recaptured in its
own commit with that justification.

slepp:
Delta closed at `15ab9f1`.

## B — reassignment with `flag == 0` leaked the old generation

Reproduced: 50 frames alternating move-then-reassign / nonmove-then-reassign,
25 leaks (the nonmove half). **Two holes stacked**, and both had to go — closing
either alone leaves the leak.

`Rc.new(..)` is an `RcIntrinsic` with no arm in the reassignment alias
predicate, so it fell to the conservative tail and every `shared = Rc.new(..)`
was treated as possibly embedding an un-retained alias of the outgoing value.
That vetoed the release before it was ever reached. It cannot alias:
`hew_rc_new` copies the payload into a fresh block, so the handle is never the
old one. `New` now takes the ordinary construction rule — aliasing iff its
payload operand aliases, which keeps `Rc.new(Node { v: old.get().v })`
fail-closed. `Clone` / `WeakClone` / `Downgrade` / `WeakUpgrade` are retained
mints: they may hand back the same allocation, but the RHS is lowered before the
release is emitted, so `+1` strictly precedes the `-1` and the count cannot reach
zero across the overwrite — releasing the outgoing generation is exactly the
balancing `-1`. `GetCopy` shares the payload's un-retained leaves and `Set` /
the count and uniqueness queries own nothing, so all keep the fail-closed tail.

Behind that veto, `emit_local_overwrite_release` had no `Rc`/`Weak` arm either,
so even an admitted release emitted nothing. The symbols are the ones
`generator_yield_drop_symbol` already wires and codegen already admits as inline
`Instr::Drop` rituals (`is_known_cow_heap_drop_symbol`). They are selected at the
overwrite seam rather than in `project_field_inline_drop_symbol` — that picker is
shared with the Vec-element, record-field and generator-yield paths, and widening
it would change all three.

Safety is from ordering, not type: the release runs only through the flag-gated
arm, so a handle already transferred into an aggregate (`flag == 1`) is never
released there.

This also closes what the previous round reported as pre-existing.
`var shared = Rc.new(a); shared = Rc.new(b);` with no aggregate anywhere now
releases `a` at the store (was 16 leaks / 16 frames), and
`shared = shared.clone()` no longer leaks a generation per iteration.
`shared = shared` still skips the release — a bare `BindingRef` is the
predicate's aliasing leaf.

## A2 — single-sited by construction

Confirmed: `set_owned_local_consumed` is the lower seam, and several production
paths reach it without passing through `mark_binding_moved` (the synthetic
produced-value owner handoffs). The record now lives inside it, so the
`ConsumedAt` disposition and the runtime flag are written by one statement —
a caller cannot record one without the other, and `mark_binding_moved`'s own
store is gone rather than duplicated.

Two callers legitimately cannot take it, and I did not paper over that: the
finalized-MIR `string` / `bytes` retain-site derivations discover their handoffs
*after* block building closes, so a `push_instr` there appends to a stream
nothing emits — and they are the wrong authority anyway, resolving CoW-carrier
handoffs whose release accounting belongs to the retain-site derivation. They
call an explicitly named `set_owned_local_consumed_post_lowering` whose doc
states that condition, rather than silently getting disposition-only behaviour
from a shared entry point.

## A1 — oracle

37 tests. Added the alternating move/no-move reassign over `Rc`, the same over
`Weak` (sibling release symbol and drop kind), and `shared = shared.clone()` —
the shape that must be admitted *and* must not double-free, where vetoing leaks
per iteration and running unretained frees a live allocation. All assert exact
totals at both frame counts.

## Gates

- `cargo test -p hew-mir` — 1724 pass
- `cargo test -p hew-codegen-rs` — 848 pass
- the oracle — 37 pass
- `scripts/ll-corpus.sh verify` — 15 × 2 byte-identical, **no further golden
  change**: no corpus fixture reassigns an affine `var`, so this round is inert
  for all 15
- `make test-hew-ratchet` — PASSED, 0 actual / 0 expected
- `make test-vertical-slice` — 567 pass
- `make lint` — clean

One note on the vertical-slice runs, since two intermediate runs went red and I
do not want that read as a silent retry. Two different deadline-bearing actor
fixtures failed on two consecutive runs — `supervisor_lifecycle_phase_writes`
(expected 22, got 2) then `await_accept_shutdown_repark_loop` (timeout, 124).
A deterministic miscompile does not move between runs. Root cause is host
contention: 16 CPUs against a load average of 18 / 44 / 40 from concurrent
builds. Both fixtures are deterministic and correct when run directly —
6/6 at exit 22 and 5/5 at exit 44 respectively — and neither contains `Rc`,
`Weak` or `#[resource]`, so all three changes are gated off for them
(`affine_release_flags` is empty, the overwrite arm is type-gated, the alias arm
is `RcIntrinsic`-gated), which the byte-identical corpus corroborates. Re-run at
load 6.8: 567 pass, exit 0.

slepp:
Follow-up done at `835528839532fc3a512e3cbad3f731976a6cf9ac`.

## The blind spot was real, and it was the only one in the file

The shape allocated once above the loop and self-cloned in place, so a skipped
overwrite release only bumped that single allocation's refcount — nothing new is
ever allocated, the leak NODE count stays flat, and the slope sees nothing. The
allocation now happens inside the frame, once per iteration, so the same missing
release orphans one allocation per frame.

Measured both ways with the overwrite-release arm deleted, tolerance is 5 nodes:

| shape | 3 frames | 50 frames | slope | verdict |
|---|---|---|---|---|
| hoisted (before) | 1 leak / 80 B | 1 leak / 80 B | 0 | **passes despite the bug** |
| per-frame (after) | 3 leaks / 240 B | 50 leaks / 4000 B | 47 | **fails** |

With the arm restored the shape reports 0 leaks at both counts, and with it gone
the harness names the failure precisely:

```
retained_self_reassign: per-iteration leak SLOPE — low_frames=3 low_leaks=3,
high_frames=50 high_leaks=50. Excess of 42 NODES over the tolerance of 5
indicates a per-iteration allocation is not being released.
```

Note the split under the counterfactual: `1 passed; 1 failed` — the over-release
half still passes, which is correct, since a *missing* release cannot double-free.
Only the under-release half moves, which is the half that was blind.

The property under test is unchanged: the RHS still retains the same allocation
the slot holds, so the release must be admitted and must not double-free. The
expected total is now the sum of the frame indices rather than a constant.

I re-ran the counterfactual against every other shape in the file while the arm
was disabled; this was the only one that failed to notice, so the rest were
already measuring what they claim to.

## Merge and gates

`git merge origin/main` — clean, no conflicts, nothing to resolve (`c541c83`
Windows release-candidate validation, `7e92413` lint/CI coverage contract).

- `cargo test -p hew-mir` — 1724 pass
- the oracle — 37 pass
- `make lint` — clean. Not on your list, but `7e92413` changes `make lint`
  itself by folding in the lint/CI coverage contract, so the green I reported
  last round no longer covered the gate as it now stands. It does now.

PR #3039, opened 2026-08-26
feat(mir): add verified virtual Raw values

slepp:
## Summary

Introduces the first durable value-only lane inside Raw MIR. This is not another general-purpose IR: it separates virtual `RawValueId` values from addressable `Place`s for a deliberately small, no-drop tuple proof.

- Adds `RawValueId`, typed value definitions, `RawValueOp::{Param, ConstI64, ConstBool, TupleMake, TupleGet}`, and the sole materialization boundary: `ReturnAbi -> Place::ReturnSlot`.
- Preserves tuple semantics through SIR and Raw MIR; LLVM alone realizes it with `insertvalue` / `extractvalue`.
- Adds one canonical Raw verifier used by both strict SIR lowering and Raw-to-LLVM codegen. Virtual bodies must be scalar-ABI, non-coroutine, non-synthesized, storage-free, and have exact parameter facts plus matching Checked and zero-drop Elaborated MIR.
- Fails closed for semantic type aliases, aggregate return ABI, place/storage mixing, stale scheduler/context facts, missing ladder stages, coroutine metadata, and synthetic actor/machine provenance.
- Adds strict source E2E proving scalar parameters remain virtual through SIR -> Raw -> Checked -> Elaborated -> LLVM, with no legacy fallback or tuple/parameter allocas.

## Intentionally deferred

Raw CFG value edges/block arguments, tuple ABI, calls, ownership-bearing values, records/enums, and layout-dependent lowering. The existing `ValueId -> Place::Local` bridge remains a temporary compatibility route outside this new admitted lane.

## Validation

- `git diff --check`
- `make structural-lint`
- `cargo clippy -p hew-mir -p hew-codegen-rs -p hew-sir --tests -- -D warnings`
- `cargo test -p hew-codegen-rs virtual_raw --lib` (8)
- `cargo test -p hew-mir sir::tests:: --lib` (17)
- `cargo test -p hew-sir` (35)
- strict CLI source -> SIR -> Raw -> Checked -> Elaborated -> LLVM tuple/parameter E2E
- affected machine-transition group rerun with the in-tree stdlib path: 17 passed

A broad pipeline run was also attempted; its failures were environmental (an ephemeral-port startup race, then a manual test invocation without `HEW_STD`), not a compiler failure. CI remains the full clean-environment gate.

### DIFF
--- a/hew-cli/tests/sir_shadow_e2e.rs
+++ b/hew-cli/tests/sir_shadow_e2e.rs
@@ -45,6 +45,30 @@ fn increment(value: i64) -> i64 {
 }
 ";
 
+/// The smallest source-level vertical slice for Raw MIR virtual values. The
+/// tuple must remain semantic/value-only through strict SIR lowering; only
+/// its selected scalar field reaches the existing return ABI slot.
+const VIRTUAL_TUPLE_PROJECTION: &str = r"
+fn main() -> i64 {
+    let pair = (0, 42);
+    pair.0
+}
+";
+
+/// Extends the virtual tuple proof to the only ABI shape admitted by the
+/// first Raw-value slice: `BitCopy` scalar parameters. The tuple itself remains
+/// internal and semantic; `pair_second` returns only a selected scalar.
+const VIRTUAL_TUPLE_SCALAR_PARAMS: &str = r"
+fn pair_second(x: i64, y: i64) -> i64 {
+    let pair = (x, y);
+    pair.1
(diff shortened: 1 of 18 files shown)


slepp:
Superseded by #3062. Verified by range-diff: this commit was replayed and then repaired there — a follow-up closes the virtual Raw value admission it left open — so the content is carried in a corrected form rather than copied.

PR #3040, opened 2026-08-26
runtime: share COW envelope lifecycle

slepp:
Summary:
- Make cow_envelope the single implementation of the ABI-stable COW envelope representation and lifecycle.
- Retain native and WASM mailbox modules as thin ABI wrappers with their existing allocator hooks and target-specific null-error behavior.
- Remove the mirrored WASM layout assertion because both targets now use the same Rust type.

Validation:
- cargo fmt --all -- --check
- cargo clippy -p hew-runtime --tests -- -D warnings
- cargo test -p hew-runtime --lib --no-fail-fast (2,438 passed)
- cargo check -p hew-runtime --target wasm32-wasip1
- git diff --check

### DIFF
--- a/hew-runtime/src/cow_envelope.rs
+++ b/hew-runtime/src/cow_envelope.rs
@@ -1,0 +1,386 @@
+//! Target-neutral COW mailbox envelope representation and lifecycle.
+//!
+//! Native and WASM mailboxes deliberately keep different queue and scheduling
+//! implementations. The payload envelope, however, is ABI-identical and has
+//! one lifecycle, so this module is its sole implementation authority.
+
+use std::ffi::c_void;
+use std::ptr;
+use std::sync::atomic::{AtomicU32, AtomicUsize, Ordering};
+
+use crate::mailbox_header::{
+    header_validate, HEW_MSG_ENVELOPE_ALIAS_ACTIVE, HEW_MSG_ENVELOPE_FORKED,
+};
+
+/// Drop glue invoked once, before the envelope releases its payload allocation.
+pub type HewMsgEnvelopeDropFn = unsafe extern "C" fn(*mut c_void);
+
+/// ABI-stable, refcounted COW container for an actor message payload.
+#[repr(C)]
+pub struct HewMsgEnvelope {
+    /// Number of live observers.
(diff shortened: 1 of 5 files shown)


slepp:
Rebased onto current `main` (`b35773d76`) and squashed to one commit, `398b66559`.

The C ABI is unchanged: `make cabi-surface-check` passes with no delta to the
generated surface, which was the condition that would have meant the extraction
had altered the contract rather than just relocated it.

The duplication is genuinely collapsed rather than moved — `mailbox.rs` loses 284
lines and `mailbox_wasm.rs` loses 198, against a 386-line shared core. Both edges
keep their own allocator hooks, their own null-error strings, `cabi_guard!` on
native and `wasm_no_mangle!`/`cfg_attr` on WASM, and the WASM compatibility
re-exports.

`scripts/structural-authority-audit.py` now fails if the `repr(C)` envelope or any
of the five lifecycle operations gains a second definition, or if either wrapper
stops delegating to the shared core. That is the property this change exists to
establish, so it is enforced rather than left to review.

Verification on the rebased head:

- `make cabi-surface-check` — pass, no surface delta
- `cargo clippy -p hew-runtime --tests -- -D warnings` — pass, 0 warnings
- `cargo test -p hew-runtime --lib --no-fail-fast` — 2,447 passed
- `cargo check -p hew-runtime --target wasm32-wasip1` — pass
- Actor envelope send/drain/close oracles — 16 passed
- Native/WASM mailbox parity suite — 38 passed
- Shared-core lifecycle coverage (layout fingerprint, refcount/alias bit, null payload, exactly-once drop, OOM rollback, forked payload independence, inherited-bit mask, reserved-bit refusal) — 7 passed
- `cargo fmt --all -- --check`, `git diff --check` — pass

Behavioural drift: none.

`make preflight` does not complete on this base for an unrelated reason:
`scripts/generate-cabi-surface.py` fails to parse on Python older than 3.12, which
takes down `cabi-surface-check` and `ffi-ownership-ratchet` before they read a
symbol. That is fixed separately in #3050; this branch will rebase onto it and the
full preflight will be rerun before this is considered ready.

PR #3045, opened 2026-08-26
ci: make pull request validation change-aware and build-once

slepp:
## Summary

Pull request validation now skips work a change set doesn't touch and builds
the Linux test surface once instead of four times. A single archive producer
compiles and packages the Linux nextest binaries; the four Linux shards
extract that read-only archive instead of each running its own `cargo build`.
sccache on pull requests is read-only (only `main` writes to the shared
cache), platform routing is Linux-first, dispatcher/reachability gains
dedup and subsumption so an owner-declared gate no longer also runs as a
duplicate raw step, and the nightly/scheduled tier now has an explicit
owner, a failure reporter, and a freshness advisory instead of running
silently with no one accountable for its result.

Measured baseline: 413 runner-minutes, 69.1 min wall (run `32966803389`,
push→main, success). Conservative projection with this change: ~342
runner-minutes, ~59 min wall. Those projected figures are a projection, not
an observation — the first hosted comprehensive run on this branch is what
actually establishes the runner-minute and critical-path result.

## Verification

- `python3 scripts/tests/test_ci_prebuilt_artifacts.py` — 19 passed
- `python3 scripts/tests/test_ci_preflight_dispatcher.py` — 186 passed
- `python3 scripts/tests/test_check_gate_reachability.py` — 74 passed
- `python3 scripts/tests/test_ci_workflow_contract.py` — 23 passed
- `python3 scripts/tests/test_release_workflow_contract.py` — 83 passed
- `python3 scripts/tests/test_rust_dependency_cache_contract.py` — 12 passed
- `python3 scripts/tests/test_target_dir_gate_wiring.py` — 2 passed
- `python3 scripts/tests/test_scheduled_failure_report.py` — 13 passed
- Focused counterfactual on the timing corpus (malformed row / missing
  field / zero-row) — passed
- `git diff --check` across the full range — clean
- Manual repro: mutated the extracted archive's `libhew.a` after
  materialization and confirmed the forced-cancel warm-up no longer writes
  into it (isolated `CARGO_TARGET_DIR` matches the proving gate's); fed the
  sccache telemetry step malformed JSON and a nonzero-exit fake `sccache`
  under `bash -eo pipefail` and confirmed the job summary degrades instead
  of reddening the job
- Preflight: ready-to-push (pre-push hook: `cargo fmt --check`, shell-script
  lint, orchestration-leak lint on source and commit messages — all pass)
- The 342-runner-minute / 59-minute projection is not independently
  recomputable from a local checkout; hosted CI is the validation step for
  those numbers, not this checklist

## Out of scope

- Making the nightly/scheduled freshness check a required gate — deferred
  until one scheduled run reports green under the new owner/reporter
  contract
- Migrating the release-gate ruleset's required-context list — deferred
  until the new aggregate platform/lint contexts have reported at least once
- Merge queue adoption — not attempted here

Squash merge planned. Subject/body at merge time will match this PR's
title and summary above — no per-commit fixup noise carried into main.


slepp:
Closing: the change this builds on is gone. I removed the change-aware routing dispatcher in favour of running the pipeline unconditionally, after measuring that the router never actually narrowed on the Linux path — the wasmtime install untarred into the working tree with no gitignore entry, so the classifier's untracked-file scan always found undeclared paths and fell back to comprehensive on every run, including pushes to main. Unconditional CI was already what we were getting; the router's remaining cost was about 1,450 machine-seconds per run of its own self-tests.

The build-once half of this PR is still worth having. I have noted it against the CI plan so it is not lost, and the branch stays for reference.

PR #3049, opened 2026-08-27
feat(sir): canonicalize constant CFGs

slepp:
## Summary

- Add slot-addressed semantic CFG edges and a deterministic CFG index with predecessors, successors, reachability, and RPO.
- Canonicalize direct constant-boolean branches to their selected edge, then compact unreachable blocks with canonical BlockId remapping.
- Run the same verifier-gated transformation at the shared HIR -> SIR boundary for dumps, shadow evidence, and strict lowering.

## Cutover boundary

This is a bounded semantic SIR pass, not a second lowering path or a generic pass manager. It preserves ValueId, OpId, types, provenance, edge operands, and UseMode; only compacted BlockIds change. Disabled compilation remains unchanged. SCCP, DCE, block-argument elimination, and Raw-MIR CFG values remain follow-up slices.

## Hardening

- Ignore unreachable predecessors when computing dominance for reachable blocks.
- Verify after branch folding and after compaction, transactionally.
- Cover duplicate edges, multi-block loop remapping, malformed input/module atomicity, and all three SIR driver modes.

## Validation

- `RUSTC_WRAPPER=sccache CARGO_TARGET_DIR=/Volumes/Extreme SSD/src/hew-target-sir-cfg-foundation-clean cargo test -p hew-sir --no-fail-fast`
- `RUSTC_WRAPPER=sccache CARGO_TARGET_DIR=/Volumes/Extreme SSD/src/hew-target-sir-cfg-foundation-clean cargo clippy -p hew-sir --all-targets -- -D warnings`
- `RUSTC_WRAPPER=sccache CARGO_TARGET_DIR=/Volumes/Extreme SSD/src/hew-target-sir-cfg-foundation-clean cargo test -p hew-cli --test sir_shadow_e2e --no-fail-fast`
- `cargo fmt --all --check` and `git diff --check`

Stacked on #3039 so its Raw-value semantics remain the sole parent change.

### DIFF
--- a/hew-cli/src/main.rs
+++ b/hew-cli/src/main.rs
@@ -411,18 +411,34 @@ fn report_strict_sir_missing_body(
 }
 
 fn lower_verified_hir_to_sir(module: &hew_hir::HirModule) -> Result<hew_sir::LoweredModule, ()> {
-    let sir = hew_sir::lower_module(module);
+    let mut sir = hew_sir::lower_module(module);
     let diagnostics = hew_sir::verify_module(&sir.module);
     if !diagnostics.is_empty() {
-        for diagnostic in diagnostics {
-            eprintln!(
-                "SIR verifier error in `{}`: {:?}",
-                diagnostic.function, diagnostic.kind
-            );
-        }
+        render_sir_diagnostics("verifier", diagnostics);
+        return Err(());
+    }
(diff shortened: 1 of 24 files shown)


slepp:
Superseded by #3065. Verified by range-diff: the three CFG commits are patch-identical there, and the Raw virtual-values commit is already in that branch's base via #3062.

PR #3052, opened 2026-08-27
fix(mir): publish ownership transfers for consuming calls

slepp:
## Summary

A consuming call on an owned `#[resource]` binding published no terminal ownership
transfer, so scope-exit lowering emitted a second, unguarded release. That is a double
free, and it reaches users through the most ordinary shapes in the language.

On current `main`:

```
$ hew run examples/v05/surfaces/typed_streams.hew
hew: fatal synchronous hardware fault          # exit 139
```

```
$ make mqtt-broker-e2e
mqtt-broker-e2e: FAIL: malformed Remaining Length was not rejected
hew: fatal synchronous hardware fault
```

The MQTT parser is never reached. Six input shapes were tried — empty connection,
truncated, fifth continuation, oversized, zero-body, and a **fully valid CONNECT** — and
all six faulted with no parser log emitted. Opening the connection is sufficient. This is
not a missing bounds check: `read_varint_len` checks `i < data.len()` before indexing and
the emitted LLVM retains that branch.

`Builder::lower_value`'s consumed `BindingRef` handling set
`affine_release_flags[binding] = 1` directly. `ownership.rs` documents
`set_owned_local_consumed` as the mandatory consume chokepoint, which publishes the flag,
the terminal `OwnershipEvent::Transfer`, and the disposition together;
`materialize_explicit_scope_exits` reads that event to decide whether the owner still
needs releasing. Setting the flag alone left the transfer invisible to it.

The fix routes that site through the chokepoint. Eleven lines in `expr.rs`; the rest of
the diff is regression coverage.

Every other site that touches `affine_release_flags` was audited and is correct:
`mod.rs:5276` and `ownership.rs:1228`/`:1286` pair the flag with exact transfer
operations, `:3633`/`:3670` are the chokepoints themselves, `:4727` is allocation, and
`:5322` is aggregate ingress and explicitly not a dataflow terminal. This was one bypass,
not a pattern.

## Impact

Any service using `Connection::into_stream_sink` could be killed by a remote peer on its
first accepted connection. `std.stream`'s `try_pipe`, `try_bytes_pipe` and
`StreamPair::close`, the `ChannelPair` construction in `std/channel/channel.hew`, and
compiler-generated receive-generator channels share the shape and fault with no
networking involved.

## Verification

After the fix:

```
$ hew run examples/v05/surfaces/typed_streams.hew
recv frame-0
recv frame-1
recv frame-2
drained 3 frames                               # exit 0
```

- `make mqtt-broker-e2e` — passes, including the malformed Remaining Length rejection it was written to check. The broker both survives the packet and rejects it.
- `cargo test -p hew-mir --lib` — 929/929
- Receive-generator channel example — exit 0, output `1 2 3`

Each new regression was confirmed to **fail without the fix**, which is what makes them
regression tests rather than assertions that happen to pass:

| Regression | Without the fix |
| --- | --- |
| Checked-MIR terminal transfer | terminal transfer count `0`, expected `1` |
| Stream pipe / bytes pipe | exit 1, sentinel absent |
| Channel pair | exit 1, sentinel absent |
| TCP split | exit 139, fatal synchronous hardware fault |

The Checked-MIR assertion sits at the layer the defect actually lives at; the runtime
cases assert on exit-code class and a sentinel line rather than golden stdout.

## Out of scope

The selected nextest gate stops at `mutable_direct_record_transfer_then_overwrite_uses_only_overwrite_flag`
with `DischargeAuthorityDrift`. That failure reproduces identically with this lowering fix
reversed, so it is pre-existing and unrelated; it is not addressed here.

Fixing this by removing the explicit free from `examples/mqtt_broker.hew` or the stdlib
wrapper was rejected — it would mask the compiler defect and leave every other consumer of
the shape broken.

### DIFF
--- a/hew-cli/tests/consuming_call_owner_transfer_e2e.rs
+++ b/hew-cli/tests/consuming_call_owner_transfer_e2e.rs
@@ -1,0 +1,138 @@
+#![cfg(unix)]
+
+mod support;
+
+use std::path::PathBuf;
+use std::process::{Command, Output};
+
+use support::{describe_output, hew_binary, repo_root, require_codegen, run_bounded_command};
+
+const STREAM_PAIR_SOURCE: &str = r#"
+import std.stream;
+
+fn main() {
+    let (text_sink, text_input) = stream.pipe(2);
+    text_sink.send("text-frame");
+    text_sink.close();
+    match text_input.try_recv() {
+        .Some(frame) => {
+            if frame != "text-frame" {
+                panic("stream text payload changed");
+            }
+        },
+        .None => panic("stream text frame missing"),
+    }
+
+    let (bytes_sink, bytes_input) = stream.bytes_pipe(2);
+    bytes_sink.send(b"ok");
+    bytes_sink.close();
+    match bytes_input.try_recv() {
+        .Some(frame) => {
+            if frame.len() != 2 {
+                panic("stream bytes payload changed");
+            }
+        },
+        .None => panic("stream bytes frame missing"),
+    }
+    println("stream-owner-transfer-ok");
+}
+"#;
+
+const CHANNEL_PAIR_SOURCE: &str = r#"
+import std.channel.channel;
+
+fn main() {
+    let (tx, rx): (channel.Sender<string>, channel.Receiver<string>) = channel.new(1);
+    tx.send("channel-frame");
+    tx.close();
+    match rx.try_recv() {
+        .Some(frame) => {
+            if frame != "channel-frame" {
+                panic("channel payload changed");
+            }
+        },
+        .None => panic("channel frame missing"),
+    }
+    println("channel-owner-transfer-ok");
+}
+"#;
+
+const TCP_SPLIT_SOURCE: &str = r#"
+import std.net;
+import std.stream.{Sink, Stream};
+
+fn main() {
+    let listener = net.listen("127.0.0.1:0");
+    let port = listener.local_port();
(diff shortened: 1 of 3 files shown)


slepp:
Blast radius confirmed by running each case against a build of this branch. This 11-line
lowering change clears fatal faults in four suites that looked unrelated:

| Suite | On `main` | With this fix |
| --- | --- | --- |
| `mqtt-broker-e2e` | faults on **any** connection, incl. a valid CONNECT; parser never reached | passes, incl. the malformed-length rejection it was written to check |
| `examples/v05/surfaces/typed_streams.hew` (shipped example) | exit 139 | exit 0, correct output |
| vertical slice `sink_half_in_actor_state` | exit 139, 5/5 | — blocks #3055's full-slice gate |
| `tests/hew/fs_stream_test.hew` | 4 cases, fatal synchronous hardware fault | **23 passed, 0 failed** |

The common shape is a consuming call on a `#[resource]` — `sink.close()`,
`Connection::into_stream_sink`, `stream.pipe`, `bytes_pipe`, `StreamPair::close`,
`ChannelPair`. Any file exercising resource handoff was exposed, which is why four
independent lanes each hit a different-looking fatal fault and each correctly classified it
as pre-existing and unrelated to their own work. They were all this.

Merging this first unblocks #3058's preflight (shard 1, `test-hew-ratchet`) and #3055's
full vertical slice.

PR #3053, opened 2026-08-27
fix(types): recheck defer moves on every exit

slepp:
## Summary

A read of a consumed value is correctly rejected inline but was silently permitted inside a
`defer` block, where it read moved-from, zeroed storage.

On `main`:

```hew
let c = Conn { fd: 3 };
defer { println(f"DEFER reads {c.id()}"); }
c.close();
```
```
CLOSE fd=3
after explicit close
DEFER reads 0        <- exit 0, no diagnostic
```

The identical read written inline is a hard error:

```
error: use of moved value `c`
note: value was consumed here
```

Defer bodies were type-checked once at their registration point
(`hew-types/src/check/statements.rs`) and inlined at each exit edge
(`hew-mir/src/lower/scope.rs`), so the move-state at the materialization site was never
re-checked. The comment at `hew-mir/src/lower/mod.rs:1581-1583` asserted the opposite —
that moved or consumed bindings are rejected by the move-checker at the materialization
site. They were not; the comment is corrected here.

Each defer body is now re-checked against the move-state at every site where it
materializes.

## Behaviour

After the fix, the reported case:

```
d_moved.hew:18:36: error: use of moved value `c`
 18 |     defer { println(f"DEFER reads {c.id()}"); }
    |                                    ^
d_moved.hew:19:5: note: value was consumed here
 19 |     c.close();
    |     ^
  = help: duplicate `c` with `clone c` before the consuming use to keep the original usable
```

Same diagnostic quality as the inline path — binding named, consuming use pointed at, same
help text.

A consume on only one branch rejects that materialization edge and names the consuming
call inside that branch; the non-consuming sibling edge stays valid.

Registration-point *scoping* was already correct and is unchanged: a forward reference
(`defer { println(f"{x}") } let x = 5;`) still fails as `undefined variable`. Only the
move-*state* was stale.

## Verification

- Reported reproduction: exit 0 with `DEFER reads 0` before; exit 1 with the diagnostic above after.
- Focused regressions: 3 passed, 0 failed.
- `cargo check -p hew-types -p hew-mir -p hew-cli`: pass
- Formatter: 654 Hew sources clean; Rust formatting clean
- Workspace clippy with warnings denied: pass on both commits
- `git diff origin/main --check`: pass

Each regression was confirmed to fail without the fix:

| Shape | Without the fix |
| --- | --- |
| Unconditional consume | diagnostic absent |
| Conditional consume | only the lower-quality `E_MIR_CHECK`, no clone help |
| Early-return consume | diagnostic absent |
| Valid on every edge | compiled (must keep compiling) |
| Forward reference | rejected (must stay rejected) |
| Unconsumed resource | compiled (must keep compiling) |

## Scope

This closes the hole on the exit edges `defer` already runs on. Whether `defer` should also
run on trap, cancel and unwind edges is a separate open design question. This change makes
that work easier rather than harder: future exit kinds can reuse the same
materialization-edge replay with their own ownership state.

## Out of scope

Two pre-existing environment-coupling failures were observed and are not addressed here:
`test-build-harness` assumes `../../target/release-lib/hew` and rejects an external
`CARGO_TARGET_DIR`, and two `hew-types` library tests fail under an external target because
they cannot locate the development stdlib (2120 passed, 2 failed). The known
`dogfood-compile-measure` baseline mismatch on `main` is also unrelated.

### DIFF
--- a/hew-cli/tests/defer_move_state_e2e.rs
+++ b/hew-cli/tests/defer_move_state_e2e.rs
@@ -1,0 +1,161 @@
+//! End-to-end coverage for move-state replay at defer materialization edges.
+
+mod support;
+
+use std::process::{Command, Output};
+
+use support::{hew_binary, strip_ansi};
+
+fn check_source(name: &str, source: &str) -> Output {
+    let fixture = support::tempdir();
+    let source_path = fixture.path().join(name);
+    std::fs::write(&source_path, source).expect("write Hew fixture");
+    Command::new(hew_binary())
+        .args(["check", source_path.to_str().expect("UTF-8 fixture path")])
+        .current_dir(fixture.path())
+        .output()
+        .expect("hew check must run")
+}
+
+const RESOURCE_DECL: &str = r"
+#[resource]
+type Conn {
+    fd: i64;
+}
+
+impl Conn {
+    fn id(self) -> i64 {
+        self.fd
+    }
+
+    fn close(self) {}
+}
+";
+
+#[test]
+fn defer_reads_recheck_unconditional_conditional_and_return_edges() {
+    let source = format!(
+        r#"{RESOURCE_DECL}
+fn unconditional() {{
+    let unconditional_conn = Conn {{ fd: 1 }};
+    defer println(f"{{unconditional_conn.id()}}");
+    unconditional_conn.close();
+}}
+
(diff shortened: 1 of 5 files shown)


slepp:

PR #3055, opened 2026-08-27
fix(actor): carry supervised roles in ChildRef

slepp:
## Summary

A supervised child handle held across a function boundary could dereference a freed actor
and terminate the process. `ChildRef<T>` replaces `LocalPid<T>` for supervised children so
that cannot be expressed.

Three conditions had to combine, which is why the suite missed it entirely: the child's
restart budget **exceeded** (permanently dead, not merely restarted), a **stale** handle
captured before the death, and a call crossing a **function boundary**. A single in-budget
crash through the identical helper is fine.

```hew
fn call_work(w: LocalPid<Fragile>) -> i64 { match await w.work() { ... } }
// ... crash past `intensity: 2 within 60s` ...
let _r = call_work(stale);
```
```
inline equivalent:  Err, exit 0
through the helper: hew: fatal synchronous hardware fault, exit 139
```

`sup.child` handles were fungible only because of per-function MIR lowering state
(`fungible_child_refs`) recorded at two sites. Nothing propagated it across a parameter,
field, return, or `Vec`, so the callee received a fresh `Place::ActorHandle` with no entry,
the re-resolve was skipped, and the send dereferenced a freed actor. The pool accessor
`sup.workers[i]` was a third path with a fourth behaviour.

Fixed by type split rather than by propagating the side table to more lowering paths.
Per-path patching had already failed once — the pool accessor was the evidence.

## What changed

`ChildRef<T>` carries the supervised role, so re-resolution is correct by construction in
every value-flow position. **The MIR-local side tables are removed.** Both former
`fungible_child_refs` registration paths and the static-pool accessor now construct the same
`ChildRef` value; pool indexing translates an index to a stable slot and no longer requires a
live actor. Static access, pool index and `.get()`, asks, tells, selects and `await_restart`
all use the one mechanism.

No new substrate was needed — no new place kind, no runtime generation table, no replacement
side table.

### Breaking surface

| Before | After |
| --- | --- |
| `sup.child` → `LocalPid<T>` | `ChildRef<T>` |
| `sup.pool[i]` → `LocalPid<T>` | `ChildRef<T>` |
| `sup.pool.get(i)` → `Option<LocalPid<T>>` | `Option<ChildRef<T>>` |
| `await_restart sup.child` → `LocalPid<T>` | `ChildRef<T>` |

`ChildRef<T>` implements `send` and `Pid`. Nested supervisors stay `LocalPid<Supervisor>`;
nested actor leaves are `ChildRef<Actor>`. Ordinary spawned `LocalPid` values are unaffected
and needed no migration. Builtins documentation, the language specification, the changelog,
and affected fixtures are updated.

## Verification

The old signature is now a compile error, which is the point — the defect is unrepresentable
rather than merely fixed:

```
error: type mismatch: expected `LocalPid<Fragile>`, found `ChildRef<Fragile>`
```

| | Before | After |
| --- | --- | --- |
| helper path | abort in 9/10 runs (debug `134`, release `139`) | exit 0, 5/5 |
| inline path | exit 0, 5/5 | exit 0, 5/5 |

- Cross-product fixture — all **24** cells: inline, parameter, return, record field, `Vec`, closure × live, restarting, budget-exceeded, permanently stopped. Each cell exercises both ask and tell, plus stale, indexed and `.get()` static-pool access after terminal failure.
- MIR supervisor tests 29/29; codegen stable-role tests 2/2; runtime supervisor tests 119/119
- Workspace clippy with tests and `-D warnings`: pass
- Hew formatter on all changed `.hew` files: pass
- 4/4 commits signed

The fixture fails on an isolated `origin/main`: the source lacks `ChildRef`, and a mechanical
`ChildRef`→`LocalPid` back-port also fails because record storage remains affine and closure
extraction is unsupported there.

## Out of scope

Three pre-existing failures, none caused by this change:

- `dogfood-compile-measure` — byte-exact `ll_bytes` baseline, expected `453006`, got `440812`; the IR got *smaller*.
- Two `hew-types` library tests (2121/2123) cannot discover the development stdlib under an external `CARGO_TARGET_DIR`.
- The aggregate vertical slice reaches `sink_half_in_actor_state` and hits the consuming-call double free (exit 139, 5/5). That is #3052; `hew-mir/src/lower/expr.rs` is untouched here, so the two land without overlap.


slepp:

PR #3056, opened 2026-08-28
fix(ci): restore checkout fetch-depth and de-duplicate step keys

slepp:
## Summary

#3054 inserted the Python 3.12 setup steps between each `checkout` and its `with:` block.
`fetch-depth: 0` was orphaned onto `setup-python`, each `checkout` lost it, and three steps
ended up with a duplicated `with:` key. `actionlint` rejects it on `main` today:

```
ci.yml:235:9: key "with" is duplicated in element of "steps" section. previously defined at line:232
ci.yml:292:9: key "with" is duplicated in element of "steps" section. previously defined at line:290
ci.yml:833:9: key "with" is duplicated in element of "steps" section. previously defined at line:830
```

One of the three carries a comment saying exactly why the depth is needed: the
orchestration-token scan reads the `origin/main..HEAD` commit range, which a shallow clone
cannot resolve.

Each Python step moves below its checkout so the depth stays on the step that needs it.

## Verification

- `actionlint .github/workflows/ci.yml` — exit 0 (three errors on `main`)
- Diff is a pure move: 9 insertions, 9 deletions, no step added or removed, no `if:` condition changed

## Scope

Deliberately limited to the YAML repair so it can land immediately — `main` is affected now.

The fail-fast-on-pull-requests change and the sccache statistics fix are held back on a
separate branch: extending fail-fast to PRs breaks an existing dispatcher self-test that
asserts exactly two push-only fail-fast checks, and that assertion encodes the policy being
changed. It needs its own decision rather than riding along here.

## How this got in

#3054's brief required `make lint` and preflight but never required `actionlint` on a
workflow change, and neither gate parses workflow YAML. A lane that edits
`.github/workflows/**` should run `actionlint` before push; that is a briefing gap, not a
missing repository gate — `actionlint` already exists and is already wired into CI, it just
was not run locally before the merge.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -733,9 +733,23 @@ jobs:
           BUILD_RESULT: ${{ needs.compiled-hew-linux.result }}
           SHARD_RESULT: ${{ needs.compiled-hew-shards.result }}
         run: |
-          test "$CHANGE_RESULT" = success
-          test "$BUILD_RESULT" = success
-          test "$SHARD_RESULT" = success
+          set -euo pipefail
+          # When the code path is selected every leg must have SUCCEEDED. When it
+          # is not, the legs are legitimately skipped and that satisfies the gate --
+          # but a leg that actually ran and failed is still a failure, so `skipped`
+          # is the only non-success result accepted here.
+          for leg in CHANGE_RESULT BUILD_RESULT SHARD_RESULT; do
+            result="${!leg}"
+            if [ "$result" = success ]; then
+              continue
+            fi
+            if [ "${RUN_CODE_PATH:-}" != 'true' ] && [ "$result" = skipped ]; then
+              echo "$leg skipped; code path not selected"
+              continue
+            fi
+            echo "::error::$leg is '$result' (expected success)"
+            exit 1
+          done
 
       - name: List the independent full suite inventory
         if: env.RUN_CODE_PATH == 'true'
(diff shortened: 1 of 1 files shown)


slepp:
Added a second commit, `30acb8af4`, after this PR's own CI exposed a bigger defect than the
one it was opened for.

`Build & test (Linux)` failed on this branch with:

```
CHANGE_RESULT: skipped
RUST_GATES_RESULT: skipped
COMPILED_HEW_RESULT: failure
```

Both required aggregates — `linux-required` and `compiled-hew-aggregate` — assert
`test "$RESULT" = success` on every leg, with **no** `RUN_CODE_PATH` guard, while the
sibling steps in the same jobs *are* guarded. So when change detection correctly skips the
code jobs, the legs report `skipped`, the assertion fails, and the required check goes red
for a change that touched no code.

That means a docs-only or workflow-only pull request has been structurally unable to pass
the required Linux check since `b7144902b` (2026-08-09, "perf(ci): shard the compiled Hew
suites"). This PR is itself an instance: a nine-line YAML move cannot go green.

The fix accepts `skipped` **only** when the code path was not selected. When it was, every
leg must still be `success`, and any other result — `failure`, `cancelled`, `timed_out` —
fails the check either way. It also reports which leg and what result through `::error::`
instead of a bare non-zero exit, so the next failure names itself.

Verified with `actionlint` (clean) and by reading the `env:` scope of both jobs —
`compiled-hew-aggregate` binds `RUN_CODE_PATH` to `selected_compile`, `linux-required` to
`platform_smoke`; both are in scope at the step.

This is deliberately fail-closed in both directions. It does not make a red leg pass; it
stops a correctly-skipped leg from being read as a failure.

PR #3057, opened 2026-08-28
fix(actors): make message loss visible at the send site

slepp:
## Summary

Three defects in one contract, all found by first-time users trying to build a service.

**A send could discard every message and the program exited 0.** `overflow drop_new`
dropped 100% of valid work, and `HEW-SPEC-2026.md:163` *defined* that loss as a successful
send. **`block`, the default, deadlocked at `HEW_WORKERS=1`** — it parked the OS worker
thread rather than yielding. **`fail` trapped the sender with empty stderr** while the
normative table promised an error return that a unit-typed `tell` can never deliver.

Reliability is the language's first tenet; silently losing work and reporting success is
the opposite of it.

## What changed

Lossy sends are now visible at the call site. The common case is untouched:

```hew
worker.work(42);          // unchanged — reliable send, no ceremony
```

Only policy-sensitive sends require `match`, `?`, or an explicit `let _ =`. Ignoring a
lossy result is rejected rather than warned about.

`fail` returns an error the caller can actually observe. `block` suspends the sender
instead of parking the worker thread, so a single-worker runtime makes progress.

The specification is updated to match — §9.3 is recast around actor mailboxes and FIFO
producer suspension, and the loss-as-success definition is gone. These are deliberate
language decisions, not documentation of existing behaviour.

## Breaking

The old unit-typed `drop_new` no longer compiles against the new result — which is the
change working as intended. Stdlib, examples, docs, the language guide, the backpressure
example, and the structural authority inventory are updated.

## Verification

Every case proven to fail without the fix:

| Case | Before | Now |
| --- | --- | --- |
| Lossy overflow | discards all, exit 0 | exit 42, `LOSS_VISIBLE`, empty stderr |
| Single-worker `block` | deadlocks on the condvar wait | exit 42, `BLOCK_WORK_DELIVERED` at `HEW_WORKERS=1` |
| `fail` policy | traps, no observable return | exit 43, `FAIL_VISIBLE`, empty stderr |
| Normal send (control) | exit 42 | exit 42, `NORMAL_SEND_DELIVERED`, unchanged source |
| Ignored lossy result | silently accepted | rejected, exit class 1 |

- `cargo test -p hew-runtime --lib` — 2,450/2,450
- Native mailbox suite 76/76; WASM mailbox suite 55/55
- `make cabi-surface-check` — no C ABI delta
- Structural authority audit — 162 rows, 2,191 parsed test ranges
- Codegen 319/319, HIR 161/161, MIR 929/929
- Hew formatter — 654/654 sources
- Documentation examples — 207 pass

## Rebased over #3040

#3040 merged mid-flight and extracted the `repr(C)` envelope and its five COW lifecycle
operations into a shared core. Rebasing conflicted on the sender-suspension commit. Both
properties are preserved and verified: the envelope single-definition structural guard
passes with exactly one native and one WASM delegation per operation, and suspension does
not regress to the condition-variable wait. Suspension uses mailbox-node/read-slot
ownership and needs no COW lifecycle hook, so no addition to the shared core was required.

Commits 1, 2 and 4 are unchanged by the rebase; commit 3 differs only by removing an
obsolete `header_validate` import.

## Out of scope

Pre-existing and unrelated: the `dogfood-compile-measure` byte-exact `ll_bytes` baseline
(the emitted IR got *smaller*), and two `hew-types` tests that cannot locate the
development stdlib under an external `CARGO_TARGET_DIR`.


slepp:

PR #3058, opened 2026-08-28
ci: delete change-aware routing and run the pipeline unconditionally

slepp:
## Summary

Deletes the change-aware router and makes the pipeline unconditional.

The router never narrowed anything on the Linux path. The `Install wasmtime` step untarred
into the working directory with no `.gitignore` entry, so the dispatcher's untracked-file
scan always found undeclared paths and failed closed to `comprehensive` — confirmed in 3 of
3 sampled runs, including a push to `main`. Unconditional CI was already the status quo;
what the router actually cost was **~1,450 machine-seconds per run running its own
self-tests**.

## Changes

**Router removed.** The dispatcher shell entry point, the Python selection and
classification core, the diff/path-filter job, and the routing tests are gone — 3,797
dispatcher self-test lines, 4,100 including the 303-line playground path-filter oracle.
Replaced by a checked-in four-shard assignment covering all 86 participating commands
exactly once. The independent `main-health` gate and its behavioural tests are preserved.

**sccache observability.** `SCCACHE_IDLE_TIMEOUT=0` keeps the daemon alive until the
post-step, so per-job hits, misses, writes and errors are real numbers instead of a reset
counter. No hit-rate improvement is claimed here — this is what makes one measurable.

An earlier audit reported a 64–98% sccache write-failure rate. That figure is **not
supported**: the post-action ran 637s after the last compile, sccache idles out at 600s, so
a fresh daemon reported zero requests. The denominator was zero. This commit fixes the
observability so the real rate can be read next run.

**PR fail-fast.** Both dispatcher invocations enable `--fail-fast` for pull requests as well
as pushes; in the sampled run about 25 minutes of shard 2 executed after `make test` had
already gone red. The self-test asserting exactly two *push-only* fail-fast checks is
updated to assert two *push-or-PR* checks — that assertion encoded the policy being changed.

**Wasmtime spill.** Linux and macOS extract under `${RUNNER_TEMP}/wasmtime`, Windows under
`$env:RUNNER_TEMP\wasmtime`. Nothing unpacks into the checkout.

## The zero-gate trap, checked explicitly

Shards previously passed `--base origin/main`, which on a push to `main` **is** HEAD, and
the dispatcher exited 0 on an empty diff. A partial change here could have made every main
push run zero gates and report green.

After this change, a push to `main` runs: `main-health` (its PR-only step skipped), lint,
licenses, playground, compiled-Hew build/shards/aggregate, all four Linux static shards,
full Windows tests, and full macOS tests. **No command compares against `origin/main`**, so
`HEAD == origin/main` cannot produce an empty selection.

## Verification

- actionlint — 1/1 changed workflow
- Dispatcher suite before deletion — 162/162 Python, 13/13 timeout tests
- Static shard assignment — 86/86 commands exactly once
- Reachability — 74/74 plus lint-coverage counterfactuals
- Workflow/release contracts — 131/131; compiled-Hew shard contracts 11/11; FreeBSD 28/28
- Build harness — 2 counterfactual suites, 9/9 interface tests

**Projected critical path:** measured before `61m46s`; estimated after
`max(~46m Linux, 51m19s measured Windows) ≈ 51m19s`. The ~1,450 machine-seconds removed
cannot be subtracted from wall time directly, and Windows becomes the limiter — that is the
next target, not this PR.

## Out of scope

`make preflight` stops in shard 1 at the unchanged `test-hew-ratchet`: `fs_stream_test.hew`
fails 4 existing filesystem cases with fatal synchronous hardware faults. That file uses
`sink.close()` on a `#[resource]`, the consuming-call pattern fixed by #3052, so it is
very likely the same defect and not a CI change. Being confirmed separately.


slepp:
The `fs_stream_test.hew` blocker is confirmed as #3052's defect, not a CI change.

Built #3052's compiler and ran the file directly: **23 passed, 0 failed**, against 4 cases
failing with fatal synchronous hardware faults on `main`. Same for `mqtt-broker-e2e`, the
`typed_streams` shipped example, and the vertical slice's `sink_half_in_actor_state` — all
one consuming-call double free in MIR lowering.

This PR is unblocked once #3052 lands.

slepp:
Added `f4cfb41d6` — the reason every gate was skipping is bigger than leftover references.

**Every gate job declared `needs: main-health`.** So whenever `main` was red, every gate job
on every pull request skipped — including the pull requests that fix `main`. That single
dependency explains the universal `SKIPPED` state, why required checks could not pass, why
bypasses were unavoidable, and why `main` stayed red once it went red.

Trunk health now reports without suppressing gates. The self-test that encoded the old
coupling, `test_every_job_waits_for_a_green_main`, is rewritten as
`test_trunk_health_does_not_suppress_gate_jobs` — it asserts the inverse, so the coupling
cannot come back silently.

Now unconditional: `main-health`, `lint`, `license-check`, `playground-wasm-build`,
`compiled-hew-linux`, all four `build-and-test` matrix legs, Windows, macOS, and both
aggregates.

One job stays conditional: `compiled-hew-shards` requires a successful `compiled-hew-linux`
because it consumes that artifact. That is build ordering, not path or event selection — a
skip there is still illegitimate and the aggregate still fails it.

The PR 3056 aggregate guard is untouched. Both aggregates still accept only `success`; no
`skipped` relaxation was added.

**Push-to-main trap, rechecked:** a push to `main` schedules 17 job executions. The PR-only
trunk-health step skips but its job succeeds. All root gates schedule independently,
including four Linux shards running the static 4/26/26/30 assignment — 86 commands. Shards
follow the artifact build, then both aggregates run. No diff base and no dispatcher remain,
so a zero-gate green run is impossible.

Verification: stale-reference sweep 0 matches across 16 files; actionlint 1/1; workflow
contract 86/86; static shard assignment 86/86 exactly once; fmt and diff clean.

`make preflight` did not complete — the workspace volume hit `errno=28`, no space left, at
the linker after 8/8 fast baselines. Local capacity, not a code failure; reclaimed since.

slepp:

PR #3059, opened 2026-08-28
fix(mir): canonicalize edge carry emission order

slepp:
## Summary

The compiler emitted ownership `EdgeCarry` facts in a different order on different runs of
the same input. Forty pre-fix compiles produced **37 distinct `EdgeCarry` sequences** for
`rc_weak_lifecycle` and 6 for `r4_runtime_abi`.

`ExactOwnerState` (`hew-mir/src/lower/drop_plan.rs`) is a randomized `HashMap`, and
`materialize_explicit_goto_edge_carries` (`hew-mir/src/lower/mod.rs`) iterated it directly,
appending facts in bucket order.

This surfaced through `sir-shadow-verify`, but it is **not a SIR defect**. The parity
failure was between two runs of the ordinary *legacy* path; the shadow oracle merely
noticed. Hew is deterministic by design, so an input producing different fact sequences
across runs is a defect on its own terms.

## Fix

Facts are ordered by `OwnerId` — source binding declaration identity, then ownership
generation — sorted at the emission boundary. That is a meaningful key rather than a
convenient one: it derives from declaration order, not from iteration accident.

Both paths emit it. Ordinary lowering canonicalizes before the Raw/Checked MIR freeze;
shadow mode builds that same pipeline and clones it before SIR replacements. Independent
ordinary and `--sir-shadow` invocations are now byte-identical.

Deliberately **not** fixed by sorting at the comparison site. That would make the oracle
lenient and leave the nondeterminism in the compiler.

## Verification

- Repeated-compile determinism regression: fails on compile 2 for both fixtures without the fix; passes 2 fixtures × 4 compiles with it
- `sir-shadow-verify`: full corpus 15/15; `rc_weak_lifecycle` and `r4_runtime_abi` pass individually
- `cargo test -p hew-mir --lib` — 929 passed
- `cargo test -p hew-sir` — 31 passed
- Shell script lint — 54 scripts

**No other fact kind carries the same exposure.** Removing `EdgeCarry` lines left one
identical non-`EdgeCarry` raw-MIR sequence across all 40 pre-fix runs for each fixture.

## Why it matters beyond SIR

This is slice S0b of the compiler-core modernization, and a precondition for judging
#3039's `r4_layout` and replaying the SIR CFG work. But the determinism defect stands
independently: any consumer comparing two compilations — a golden fixture, a differential
oracle, a reproducible-build check — could have been derailed by it.

### DIFF
--- a/hew-mir/src/lower/mod.rs
+++ b/hew-mir/src/lower/mod.rs
@@ -9006,11 +9006,17 @@ fn materialize_explicit_goto_edge_carries(blocks: &mut [BasicBlock], builder: &m
                 _ => None,
             })
             .collect::<HashSet<_>>();
-        let carries = exit
+        let mut carries = exit
             .iter()
             .filter(|(owner, place)| !existing.contains(&(**owner, **place)))
             .map(|(owner, place)| (*owner, *place))
             .collect::<Vec<_>>();
+        // `ExactOwnerState` is a HashMap because transfer replay needs keyed
+        // mutation, but its randomized iteration order must not become MIR
+        // instruction order. OwnerId follows source binding declaration order
+        // and then the binding's ownership generation, so it is the semantic
+        // order for the exact generations carried across this source edge.
+        carries.sort_by_key(|(owner, _)| *owner);
         for (owner, place) in carries {
             let insert_at = block.instructions.len();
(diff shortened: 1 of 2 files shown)


slepp:

PR #3060, opened 2026-08-28
feat(mir): add verified virtual Raw values

slepp:
## Summary

Slice S1 of the compiler-core modernization: the first durable value-only lane inside Raw
MIR, replayed from #3039 onto current `main` and completed to the slice's definition.

This is not another general-purpose IR. It separates virtual `RawValueId` values from
addressable `Place`s for a deliberately small, no-drop proof.

- `a94bfd63d` — replays #3039's verified virtual Raw values: `RawValueId`, typed value definitions, `RawValueOp::{Param, ConstI64, ConstBool, TupleMake, TupleGet}`, and the sole materialization boundary `ReturnAbi -> Place::ReturnSlot`.
- `f1d161ca4` — the slice's remaining requirements: a closed `RawVirtualClass` admitted-class enum, the parameter-index fix, the `temp_drop` carve, and shadow-floor ratchets.

#3039 and #3049 are superseded by this branch. Most of their red was stale-base drift —
both forked before the ownership/unwind cutover and the release fixes.

## The admitted domain, and how it fails

Admits default-convention, single-block, storage-free virtual Raw bodies with integer and
bool scalar parameters and returns, recursively bit-copy non-empty tuples, and `Param`,
integer/bool constants, tuple construction and projection, plus the sole `ReturnAbi`
materialization boundary.

Rejects — **explicitly, with a lowering error** — ownership, layout-dependent values, tuple
ABI, coroutine or synthesized provenance, context and scheduler facts, extra blocks, storage
mixing, and missing Checked or zero-drop Elaborated stages.

That distinction is the point of the whole program. A body outside the admitted domain
fails closed; it never silently reaches a legacy branch. A partial SIR that can quietly
diverge would be worse than no SIR, so this is the property to check in review.

## Verification

- Strict source → SIR → Raw → Checked → Elaborated → LLVM E2E: 1/1, confirming **no legacy MIR bodies lowered** and no parameter or tuple allocas
- `cargo test -p hew-sir` — 35/35
- SIR MIR suite — 17/17
- Raw-value LLVM tests — 8/8
- `temp_drop` size tests — 3/3; file reduced to 10,839 lines from 10,851

## Scope

No legacy path is deleted here — correct for S1 and S2 per the plan; deletions begin at
later slices. Nothing widens the admitted domain to make a test pass.

## Follow-on

S2 (constant-CFG canonicalization from #3049) is **not** in this branch. It was held: its
`sir-shadow-verify` parity failure turned out to be legacy-path nondeterminism, fixed
separately in #3059. S2 replays once that lands.

### DIFF
--- a/hew-cli/tests/sir_shadow_e2e.rs
+++ b/hew-cli/tests/sir_shadow_e2e.rs
@@ -45,6 +45,30 @@ fn increment(value: i64) -> i64 {
 }
 ";
 
+/// The smallest source-level vertical slice for Raw MIR virtual values. The
+/// tuple must remain semantic/value-only through strict SIR lowering; only
+/// its selected scalar field reaches the existing return ABI slot.
+const VIRTUAL_TUPLE_PROJECTION: &str = r"
+fn main() -> i64 {
+    let pair = (0, 42);
+    pair.0
+}
+";
+
+/// Extends the virtual tuple proof to the only ABI shape admitted by the
+/// first Raw-value slice: `BitCopy` scalar parameters. The tuple itself remains
+/// internal and semantic; `pair_second` returns only a selected scalar.
+const VIRTUAL_TUPLE_SCALAR_PARAMS: &str = r"
+fn pair_second(x: i64, y: i64) -> i64 {
+    let pair = (x, y);
+    pair.1
+}
+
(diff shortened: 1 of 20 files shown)


slepp:

PR #3061, opened 2026-08-28
test(compiler): cap dogfood IR size instead of pinning it

slepp:
## Summary

`make dogfood-compile-measure` fails on `main` because the emitted IR got **smaller**:

```
expected 453006 bytes, observed 440812
```

A byte-exact equality on IR size taxes every improvement and teaches contributors to bump
the number rather than reason about it. This is the last standing red on `main` and it has
been blocking preflight for several lanes.

## Changes

**Ceiling instead of equality.** `ll_bytes_ceiling=445221` — 4,409 bytes (1%) above the
current measured 440,812. The gate now fails when the IR grows past the ceiling and passes
when it shrinks. The 12,194-byte improvement is banked: 7,785 bytes of it stay enforced
budget rather than being handed back.

The ceiling is one value in one place with a comment on when and how to lower it. It does
**not** auto-update from the observed value — a self-updating ratchet defends nothing.

**`defines` and `basic_blocks` dropped.** Checked-MIR and the LLVM oracle already cover
structural and codegen shape. This gate now has one responsibility: IR-size regression
detection. Fewer numbers that each mean something.

**Isolated-target fix.** The gate forced `build/bin/hew` and ignored an external
`CARGO_TARGET_DIR`, so it could not run in any isolated lane — two separate lanes hit
`Error: cannot find libhew.a. Build with: make stdlib`. It now defaults to Cargo's resolved
`release-lib` binary and honours an explicit `HEW_BIN`.

## Verification

- `make dogfood-compile-measure` — PASS, `440812 / 445221`
- Fresh worktree with an external `CARGO_TARGET_DIR` — PASS, including freshness verification against the external `debug/libhew.a`
- Explicit `HEW_BIN` recipe — PASS, uses the supplied path
- **Counterfactual:** a synthetic 445,222-byte IR sample is rejected as above the ceiling — the gate still catches a regression
- `scripts/tests/test_baselines.py` — PASS

## What the gate still defends

#3014 cut the dogfood build from 113 s to 12 s. This gate is what keeps that from silently
regressing, so it is coarsened rather than deleted.

## Out of scope

`make lint` is red on `test-check-gate-reachability`, which reproduces identically on clean
`main`: A11 reports only **14 of 39 script self-tests are invoked by CI**, with 25 flagged
`no CI-reached command invokes this self-test`. That gate states every entry is a
wire-or-cut decision rather than a waiver. Real, pre-existing, and worth its own lane —
unrelated to this change, whose diff touches four gate files.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1346,12 +1346,12 @@ ll-golden: hew-native
 # inputs: hew-codegen-rs/** hew-mir/** hew-cli/**
 #         tests/compile-measure/** scripts/dogfood-compile-measure.sh
 # The gate measures define blocks, excluding host-specific module headers.
-#         build/bin/hew (release-lib profile)
+# It uses Cargo's resolved release-lib binary by default, and honours HEW_BIN
+# when a caller supplies a staged compiler explicitly.
+HEW_BIN ?= $(RELEASE_LIB_HEW)
 LINT_GATES += dogfood-compile-measure
 dogfood-compile-measure: hew $(LIBHEW_READY)
-	@args=""; \
-	if [ "$(DOGFOOD_MEASURE_UPDATE)" = "1" ]; then args="--update"; fi; \
-	HEW_BIN="$(BUILD_DIR)/bin/hew" bash scripts/dogfood-compile-measure.sh $$args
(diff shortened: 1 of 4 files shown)


slepp:

PR #3062, opened 2026-08-28
ownership, supervision, send contract, and SIR foundations

slepp:
## Summary

Six branches combined so one CI run covers them together. Each was verified on its own, but
none got real CI coverage: every gate job declared `needs: main-health`, and with `main` red
that skipped every gate on every pull request. Fixed separately in #3058.

Supersedes #3053, #3055, #3057, #3059, #3060 and #3061, and #3039/#3049 which #3060 replaced.

| Branch | What |
| --- | --- |
| `fix/dogfood-ir-size-ceiling` | IR-size ceiling replacing a byte-exact baseline |
| `fix/canonical-edgecarry-order` | deterministic `EdgeCarry` emission |
| `sir/land-raw-virtual-and-cfg` | verified virtual Raw values |
| `fix/defer-move-state-recheck` | `defer` bodies move-checked at every exit |
| `fix/childref-supervised-handle` | `ChildRef` split for supervised handles |
| `fix/send-contract-visible-loss` | message loss visible at the send site |

## Interactions between the changes

These only appeared once the branches were combined, which is why they are here together
rather than landing one at a time:

- **Blocking sends through supervised `ChildRef`s** needed shared role propagation and source-place liveness changes. The send contract and the `ChildRef` split each worked alone; together they required a shared role-aware cooperative-send ABI so a suspended send resolves the exact incarnation.
- **The SIR shadow floor exposed a one-fixture corpus deficit.** Resolved by adding `sir_virtual_tuple.hew` as the 16th fixture — the floor was *not* lowered.
- **`ChildRef` exposed a stale `LocalPid`-only sandbox profile**, updated along with its baselines.

No incompatibility remains. `ChildRef` roles stay value-carried with no side tables; lossy
outcomes stay visible and must be consumed.

## Verification on the combined tree

- **Supervision:** 24/24 flow × state cells — inline, parameter, return, record field, `Vec`, closure × live, restarting, budget-exceeded, permanently stopped. The old signature is still a compile error: `expected LocalPid<Worker>, found ChildRef<Worker>`.
- **Send contract:** lossy overflow exit 42 `LOSS_VISIBLE`; single-worker block exit 42 `BLOCK_WORK_DELIVERED`; fail policy exit 43 `FAIL_VISIBLE`; normal send exit 42 `NORMAL_SEND_DELIVERED` with source SHA unchanged; ignored lossy result rejected, exit class 1.
- **Defer soundness:** 3/3 — consumed capture gives inline-quality diagnostics; forward reference still `undefined variable`.
- **SIR:** 16 fixtures, 16 baseline compiles, 3/7 concrete bodies realized; `EdgeCarry` order identical across four compiles per fixture; strict source→LLVM end-to-end 13/13 with no legacy fallback.
- **IR size:** 440,812 against the 445,221 ceiling; the 445,222 counterfactual is rejected.
- `cargo test -p hew-mir --lib` 932/932 · `-p hew-sir` 35/35 · `-p hew-runtime --lib` 2450/2450
- Consuming-call owner-transfer regressions 2/2 · `cargo fmt --all -- --check` · `git diff --check` — clean

## Known failures, none introduced here

- **`arena_nested_vec_reuse.hew` exits 139** reproducibly on this head **and on the `origin/main` base**. Pre-existing; being investigated separately.
- **`fuzz-oracle` exceeded the 600-second dispatcher limit** (exit 143) after 15 stages passed. That limit belongs to the change-aware dispatcher #3058 deletes.
- **`make lint`** — reachability self-tests 72/74; two report 25 CI harness tests that nothing in CI invokes. Workflows and checker inputs are unchanged from the base; fixing it overlaps #3058, deliberately excluded here.


slepp:
Merged `ci/unconditional-pipeline` (#3058) into this branch — the two were circularly
blocked and had to land together.

#3058 makes CI real: every gate job declared `needs: main-health`, so a red `main` skipped
every gate on every pull request, including the ones that fix `main`. With that inverted,
#3058's own run finally executed the gates — 18 passed, and it failed at
`dogfood-compile-measure`, which is the byte-exact IR baseline this branch already replaces
with a ceiling. So #3058 could not go green without this branch, and this branch could not
get real CI without #3058.

Now one head carries both: the fixes and the pipeline that actually exercises them.

`actionlint` clean on the merged workflow. #3058 remains open for review of the CI change
in isolation; it is subsumed here.

PR #3063, opened 2026-08-28
fix(mir): preserve moved field cleanup ownership

slepp:
## Summary

Moving an owned field out of a local record double-frees. Seven lines:

```hew
type W { v: Vec<i64>; }
fn main() {
    let w = W { v: [1] };
    let v = w.v;
}
```

Exits 133 or 134 under libmalloc's breakpoint trap, 139 elsewhere, and bare 1 on some
builds — for a program that is correct. No arena, no `Option`, no reuse. It was found
through `arena_nested_vec_reuse.hew`, but that fixture is incidental to the defect.

The re-admit-a-transferred-source loop in `hew-mir/src/lower/drop_plan.rs` re-admitted from
**every** `owned_locals_snapshot` entry, including an ordinary record that
`derive_owned_record_drop_allowed` had already excluded. The parent record's drop was
reinstated after its field moved out, so both the extracted owner and the parent's field
were released.

This is not the consuming-call defect from #3052 — that fix is already on `main` and this
survives it. No Hew-level bounds check is lost in lowering: `std/arena.hew` calls
`self.slots.get(key.index)`, `hew_vec_get_clone` checks the index, and the emitted LLVM
branches on that result.

## Fix

The re-admit loop is gated to actual explicit transfer-generation sources rather than the
full snapshot, and no longer reinserts a record the allow-set derivation excluded.
Generation-specific pre-transfer unwind cleanup, the terminal sole-owner exclusion, and
escaped-sibling discharge are all preserved.

## Verification

| Shape | Before | After |
| --- | --- | --- |
| `Vec<i64>` / `Vec<bool>` / `Vec<string>` field | 133 | 0 |
| `Vec<Vec<i64>>` field | 133 | 0 |
| Nesting depth 1 / depth 2 | 133 | 0 |
| `Arena<Vec<_>>` — zero / one reuse | 133 | 0 |
| `Arena<Vec<_>>` — two reuses | 134 | 0 |
| Arena population 1 / 2 | 133 / 134 | 0 |
| Scalar `Arena<i64>` *(control)* | 0 | 0 |
| Retain-backed `Arena<string>` *(control)* | 0 | 0 |
| `Option<string>` record field *(control)* | 0 | 0 |
| Tuple `Option<Vec<i64>>` *(control)* | 0 | 0 |

- Generality/control matrix — 15/15
- `arena_nested_vec_reuse` — passes, expected `3, true, true, 3`
- `cargo test -p hew-mir --lib` — 930/930
- `fuzz-oracle` — 754 total: 612 clean, 103 frontend rejects, 39 expected abort/crash verdicts
- `cargo fmt --all -- --check`, Hew formatter 2/2, `git diff --check` — clean

The new test **fails without the fix**: the counterfactual run reports `0 passed, 1 failed`,
with the sibling-release assertion observing 0 releases instead of 1.

## The coverage gap, which is the real lesson

No test covered the **final allow-set composition**. The focused sole-owner unit test passes
because it only verifies that `derive_owned_record_drop_allowed` excludes the parent — it
never verifies that final elaboration *preserves* that exclusion. A derivation was tested in
isolation while the plan consuming it went untested.

Added: a final-elaboration and execution test for scope-live record → move out a
non-retain-owned field → independently drop the extracted owner, asserting exactly one
destructor call and correct owned-sibling cleanup. Plus the seven-line program as a corpus
fixture — it is the smallest statement of the bug.

## Pre-existing failures, reproduced unchanged on `origin/main`

Not widened into this fix: `mutable_direct_record_transfer_then_overwrite_uses_only_overwrite_flag`,
three recursive JSON/TOML/YAML WASM lifecycle cases, and `consume_param_transfer_builder`.


slepp:

PR #3067, opened 2026-08-29
fix(mir): make consuming call handoffs unwind-safe

slepp:
## Summary

- distinguish direct Hew, declared-extern, and catalogued runtime consuming-call ownership contracts
- transfer direct Hew consume arguments before invoke so the callee owns normal and unwind cleanup from entry, including fresh rvalues and projected resource fields
- retain declared extern and runtime collection consume owners in the caller through invoke, then commit their guard, physical neutralization, and terminal transfer only on the normal successor
- keep fluent receivers on their existing discharge lineage and consuming parameters out of the separate owned-carrier protocol
- prevent stale/double cleanup across JSON, TOML, YAML, channel, and runtime collection resource handoffs

## Verification

- `git diff --check`
- all-target Clippy with `-D warnings` for `hew-mir`, `hew-codegen-rs`, and `hew-cli`
- direct/extern Checked-MIR ownership suite (2/2), including fresh-owner transfer and unwind exclusion
- resource projection transfer, reusable-closure rejection, associated/static parameter-zero, runtime collection exactly-once, and post-transfer reuse oracles
- affine resource carrier boundary suite (9 passed, 1 macOS-only ignored)
- native TOML fresh/fluent resource fixture (exit 0, output `8`)
- TOML and JSON WASI execution regressions
- focused resource-close vertical suite (8/8) and direct-consume negative regression (1/1)

The exact `generator_in_local_tuple_not_extracted_keeps_source_drop` double-free remains red, but it is already identified by test ID and signature in the measured `main` baseline under generator extraction/reaggregation authority. It is not introduced by this branch.

### DIFF
--- a/hew-cli/tests/affine_resource_carrier_boundaries.rs
+++ b/hew-cli/tests/affine_resource_carrier_boundaries.rs
@@ -551,6 +551,19 @@ fn consuming_a_resource_from_a_reusable_closure_fails_closed() {
 fn channel_handle_clone_terminals_match_runtime_semantics() {
     assert_exact_runtime("channel_sender_clone", CHANNEL_SENDER_CLONE_BODY, "1\n");
     let sender_mir = raw_mir("channel_sender_clone", CHANNEL_SENDER_CLONE_BODY);
+    let channel_ctor = function(
+        &sender_mir,
+        "fn std$channel$try_new(i64) -> Result<(Sender, Receiver), string>",
+    );
+    let (_, after_pair_free) = channel_ctor
+        .split_once("call hew_channel_pair_free")
+        .expect("channel construction must discharge its transient pair through the extern ABI");
+    assert!(
+        after_pair_free.contains("neutralize_payload _13 [CallDischargeConsume]")
+            && !after_pair_free.contains("drop _13 ty=std.channel.ChannelPair"),
+        "a consuming extern must neutralize the pair on its normal successor; \
+         its source slot cannot remain armed for ChannelPair::close:\n{channel_ctor}"
+    );
     assert!(
         sender_mir.contains("call hew_vec_clone_owned(")
             && sender_mir.contains("call hew_vec_push_owned_move(")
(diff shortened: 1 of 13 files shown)


slepp:
Follow-up head `a047d3dd2`: the macOS exact-base comparison exposed one branch-only structural failure (`hew-mir::lower_module_size::lower_modules_stay_under_line_ceiling`). I moved the new affine call-consume candidate/site-context helpers from `lower/expr.rs` into the existing ownership lowering module; behavior is unchanged and `expr.rs` is now 9,956 lines (cap 10,000).

Validated locally with `RUSTC_WRAPPER=`:
- exact `lower_modules_stay_under_line_ceiling`
- `consuming_call_owner_transfer` (direct Hew unwind ownership + declared-extern normal-edge commit)
- exact affine carrier ABI regression
- exact projection-to-direct-consume transfer regression
- exact reusable-closure fail-closed regression
- `cargo clippy -p hew-mir --all-targets -- -D warnings`
- `cargo fmt --all -- --check` and `git diff --check`

The prior exact-head hosted comparison also showed the patch fixed the base TOML WASI round-trip and both affine carrier tests; all other completed reds matched exact base commit `c5a5cf459` by test ID/signature.

slepp:
Merge evidence (red-main protocol): two signed commits rebased onto `0b62a89d3`; 13 files as stated. `make hew` and `make lint` clean. `consuming_call_owner_transfer` 2/2 with two tests asserting opposite unwind-edge behaviour (declared extern consume commits only on normal return; direct Hew consume transfers before the invoke) — reverting the fix flips the second; the codegen helper rename gains a `transferee: None` counterfactual. `lower_modules_stay_under_line_ceiling` passes: `expr.rs` 10 024 → 9 956 lines via the helper move. Every failing ID in the targeted hew-mir/hew-cli/hew-codegen-rs runs is a pre-existing `record`-fixture or already-baselined MIR/release-count failure in files this diff does not touch. Not a new cause of failure.

PR #3091, opened 2026-08-29
fix(types): narrow needless range loop lint span

slepp:
## Why

`needless_range_loop` diagnostics had two independent accuracy problems. The warning reused a whole-statement span even though it describes only the loop header, and lint diagnostics from a peer file in a directory-assembled module lost that file identity and rendered against the module's first source file.

## What

Build the lint span from the loop binding through the iterable expression so unrelated body and following statements are excluded. During the post-typecheck lint sweep, resolve every item and raw-source scan through the module graph's per-file source table and span-index namespace, then carry that source routing token on the diagnostic. Add end-to-end negative controls for both a following statement and a lint-free primary module file.

## Test

The CLI lint end-to-end suite covers the narrowed header and verifies that a loop in `recovery.hew` renders with that file's path, line, and source while excluding `journal.hew`. The `hew-types` lint-focused suite passes, formatting is clean, and workspace clippy passes with warnings denied.

### DIFF
--- a/hew-cli/tests/lint_pass_e2e.rs
+++ b/hew-cli/tests/lint_pass_e2e.rs
@@ -54,6 +54,36 @@ fn run_check(source: &str, extra_args: &[&str]) -> Output {
         .expect("failed to spawn hew check")
 }
 
+/// Run `hew check` on a program importing a directory module assembled from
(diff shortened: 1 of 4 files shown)


slepp:

PR #3092, opened 2026-08-29
docs(guide): document silent reference-cycle leaks

slepp:
## Why

Reference-count cycles can silently retain actor state and stored structures,
but the limitation is easy to miss where users learn those shapes.

## What

The language guide now calls out the actor-to-actor and self-referential
structure shapes that leak, states that no diagnostic or runtime warning is
emitted, and explains how to break cycles with `Weak`, `LocalPid`, stable-id
lookup, or acyclic ownership. The ownership contract links to that guidance
instead of duplicating it.

## Test

Rendered both changed Markdown files with Pandoc and checked the new callout,
heading, cross-reference, and the existing multi-line blockquotes around
`default { state }` and generic function values. Checked that the guide diff
contains only the two intended additive hunks and no removed `> ` continuation
lines. `make lint` reaches the documentation-aware gates but is blocked by the
unchanged `hew-fmt-check` baseline, which reports existing formatting drift in
eleven example programs.


slepp:

PR #3093, opened 2026-08-29
fix(lsp): round-trip paths through file URIs

slepp:
## Why

When a workspace path contained a space or other URI-sensitive characters, hew-lsp interpolated the raw path into a `file://` string. The resulting invalid LSP URI could panic test helpers and caused production workspace scans, import navigation, and rename planning to omit unopened files.

## What

LSP file URI conversion now goes through a platform-aware path-to-URL implementation before constructing `lsp_types::Uri`, so path components are percent-encoded and conversion failures remain fallible. All server path conversion sites use the same implementation, including Windows drive and UNC path handling.

## Test

`absolute_paths_round_trip_through_file_uris`, `absolute_paths_round_trip_with_uri_encoding`, `windows_drive_and_unc_paths_round_trip`, `coverage_scanner_surface`, `plan_workspace_rename_disk_scan_skips_symlinked_directory`, and `importer_originated_rename_includes_unopened_definition_file_edits`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1870,6 +1870,7 @@ dependencies = [
  "tower-lsp-server",
  "tracing",
  "tracing-subscriber",
+ "url",
 ]
 
 [[package]]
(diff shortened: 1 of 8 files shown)


slepp:

PR #3107, opened 2026-08-29
test(types): accept nominal identity comparisons

slepp:
## Why

Three tests still described diagnostics from before HashMap key validation, ownership markers, and nominal type declarations changed. Their expectations no longer matched the compiler paths users reach today.

## What

- Pin the bytes HashMap case to the earlier key-ownership diagnostic while preserving its exactly-one-error contract.
- Exercise ownership-marker applicability with a type alias, while retaining nominal type declarations as the legal control.
- Treat nominal type values as identity-bearing under `is`; scalar values remain rejected with `E_IS_VALUE_TYPE`.

## Test

Updated `hashmap_keys_bytes_rejected_by_key_branch_with_one_exact_diagnostic` and `ownership_marker_on_type_alias_fails_closed_before_hir`, and removed the stale `record_is_record_rejected`. `ownership_markers_on_nominal_types_still_check_clean`, `user_type_decl_is_user_type_decl_accepted`, and `int_is_int_rejected` provide the behavioural controls. Verified with the `hew-types` coverage suite, `cargo fmt`, and the exact JSON Clippy gate.

Issue #3082, opened 2026-08-29
Make the FFI ownership contract a total, build-enforced table


slepp:
## Problem

The ownership contract for every runtime and FFI callee lives in a
hand-maintained table. `callee_ownership_contract`
(`hew-mir/src/runtime_symbols.rs:411`) returns receiver-borrow,
string-argument-borrow and result-ownership facts by matching the callee's
symbol string; MIR's drop-admission provers read those projections, and an
unknown callee falls back to the fail-closed `Escapes`/`Escaping`/`Untracked`
contract. A second hand-maintained list, `runtime_symbol_consumes_receiver`
(`hew-types/src/builtin_names.rs:564`), answers an overlapping question from
another crate.

Nothing fails the build when a new exported `hew_*` symbol arrives without a
row, and hew-std symbols are skipped by the validator entirely. The compiler's
belief about who owns an FFI result can be silently absent, and the fail-closed
default turns that into a leak or a rejected program rather than an error
naming the missing row.

## Expected

One typed table, machine-checked:

- result ownership: `fresh` / `retained` / `borrowed`
- parameter modes: `borrow` / `consume` / `retain`
- discharge depth for aggregates

Every exported `hew_*` symbol has a row, hew-std included, and the build fails
if one does not. `callee_ownership_contract` becomes a checked projection of
that table, and `runtime_symbol_consumes_receiver`'s allowlist is derived from
it.

## Notes

- `hew-mir/src/runtime_symbols.rs:411`, `hew-mir/src/ownership.rs:69` (the
  contract-consolidation note)
- `hew-types/src/builtin_names.rs:564`, `hew-types/src/runtime_call.rs:538,695`
- `hew-mir/src/lower/facts.rs:1345`, `hew-mir/src/lower/vec_index.rs:183,495`
- `docs/hew-language-guide.md` FFI boundary appendix
- #2836 declares the contract in the source; this issue makes the compiler's
  internal table total and enforced.

Scheduled on `refactor/ownership-seams` as Track A7; this issue is the tracking
record.

slepp:
Once the `record` fixtures are migrated (#3109), seven `extern_ownership_double_release_e2e` cases reach their release-counter assertions and observe `releases=0` / `after_host_release=0` where 1 is expected — the extern release-counter blindness mechanism from the baseline audit, now visible again. That is this issue's oracle; the counts should move when the contract table lands.

slepp:
[...]

Issue #3068, opened 2026-08-29
WASI encoding Value teardown recursively calls close until it traps


slepp:
## Problem

`hew run --target wasm32-wasi` traps while tearing down values produced by the JSON, TOML, and YAML standard-library builders. The same programs exit normally on the native target.

## Reproduction

```hew
import std.encoding.json;

fn main() {
    let value = json.null();
    println("created");
}
```

```console
make hew-native wasm-runtime
target/debug/hew run --target wasm32-wasi repro.hew
```

The equivalent `toml.table()` and `yaml.object()` programs reproduce the same failure. Both implicit scope-exit teardown and an explicit `value.close()` reach the recursive close path and trap.

## Actual behavior

The program prints `created`, repeatedly enters `std.encoding.json.Value::close`, and exits nonzero. The terminal Wasm trap reports either an out-of-bounds access near `0xfffffff8` or a misaligned-pointer panic before `unreachable`, depending on the runtime build.

The JSON backtrace contains repeated frames like:

```text
std.encoding.json.Value::close
std.encoding.json.Value::close
std.encoding.json.Value::close
```

## Expected behavior

The value should release its native handle exactly once and the WASI process should exit successfully, matching the native target.

slepp:
The WASI TOML round-trip now compiles, runs, preserves the encoded values, and exits successfully. The exact toml_encoding_round_trips_under_wasi regression passes serially after #3067.

PR #3109, opened 2026-08-29
test: migrate record fixtures to type declarations

slepp:
## Why

Embedded Hew programs in Rust tests still declared records with the removed `record` keyword. They stopped at the migration diagnostic instead of reaching the behaviour each test was intended to exercise.

## What

Migrated the embedded declarations to the current `type` syntax with literal token substitutions. Updated the formatter round-trip fixture to use its normalized, semicolon-terminated form.

## Test

Ran `make hew`, the scoped parser, analysis, LSP, types, MIR, codegen, and CLI nextest suites, `make test`, `cargo fmt`, `make lint`, and the JSON-formatted workspace clippy check. The parser migration-hint negative control remains in place, while the migrated extern-ownership and formatter fixtures now proceed past the removed-keyword diagnostic.


slepp:
Merge evidence: one signed commit (`8dbab1123`), pure `record `→`type ` swap (diff filtered for non-swap lines is empty), rebased onto `0b62a89d3`. Local spot runs across hew-parser/analysis/lsp (all green), hew-mir lowering suites, hew-codegen-rs emission/exec and the seven hew-cli e2e binaries: no failure mentions the removed keyword; five baseline IDs now pass (`a_hashmap_insert_of_a_domestic_record_still_compiles_and_runs`, `a_hashmap_insert_of_a_wrapped_extern_record_is_refused`, `functional_update_carry_rule_depth_matrix`, `hashmap_keys_get_layout_path_copy_record_returns_correct_fields`, `hashmap_values_get_layout_path_copy_record_returns_correct_fields`); ten baseline IDs still fail for their known mechanisms (eight extern release-counter cases now visible → #3082; hashmap bytes diagnostic → #3107; tuple carry → ownership lane); two failures outside the baseline are untouched by this diff (`vertical::value_move_into_free_fn_consumes_caller_binding`, fixed by #3067; `toml_encoding_round_trips_under_wasi`, #3068). Not a new cause of failure.

PR #3125, opened 2026-08-29
docs: teach the owning-actor pattern as the road for resource-shaped values

### LINKED ISSUE #3076: Give resource-shaped values a road for 0.6.0: ship the affine kind or teach the owning-actor pattern
## What happens now

`resource` is a reserved word naming an affine kind that has not shipped. The
grammar reserves it, the type registry answers `is_resource`, and the checker
has walls keyed off it — but there is no `resource type` a user can write. What
a user meets instead is the fail-closed wall: sending any resource-shaped value
is a hard compile error, with no spelled-out way forward at the error site.

`docs/v05/ownership.md` §"`resource` — reserved for a future affine kind" says
this, and the diagnostic table says "use the owning-actor pattern". The
language guide does not teach that pattern as the answer, so the error is a dead
end for anyone who hits it before reading the ownership doc.

## What should happen

Pick one for v0.6.0:

1. Ship the affine kind — `resource type` is writable, move-only, transferable
   across a send, and the wall lifts; or
2. Keep the wall and make the owning-actor pattern the documented, taught way,
   loudly: a worked example in `docs/hew-language-guide.md`, the same pattern
   named in `docs/v05/ownership.md`, and the send-of-non-sendable diagnostic
   pointing at it by name.

Either way a user who writes a socket-holding type finds the road, not the wall.

## Why we're committing to this

Option 2 is the honest v0.6.0 answer unless the affine kind is genuinely close:
the kind is the one place a move-only value is justified, and shipping it
half-formed costs more than deferring it. What we cannot ship is the current
state, where the model says "you never move" and the compiler says "you cannot
send this" with no third sentence.

## Where to look

- `docs/v05/ownership.md` §`resource`, and its diagnostic table row
  "send of a non-sendable"
- `hew-types/src/traits.rs:592` (`is_resource`),
  `hew-types/src/check/admissibility.rs:1071,1267,1512`
- `docs/specs/handle-safety-and-resource-lifetime.md`
- `docs/hew-language-guide.md` — where the owning-actor pattern would be taught

slepp:
## Why

Sending a non-sendable, resource-shaped value to an actor stopped at the `Send`
wall without telling users how to keep the resource usable. The ownership
reference named the owning-actor pattern, but the language guide did not teach
it.

## What

Teach an executable owning-actor example that keeps a descriptor private while
ordinary operation messages cross the actor boundary. Link the ownership guide
to that example and add the same named road to the non-sendable actor diagnostic.
Pin the diagnostic help alongside a sendable-value negative control.

## Test

`make lint` and `cargo clippy --workspace --tests --message-format=json -- -D warnings`
completed cleanly. The focused `hew-types` tests
`non_send_actor_state_points_to_the_owning_actor_pattern` and
`sendable_actor_state_has_no_owning_actor_help` passed, and the
`vec_iter_actor_field_non_send` vertical-slice fixture emitted the new help.

`make test-doc-examples` compiled the new guide fence; the aggregate ratchet
remains blocked by an unrelated existing `Stack<i64>` unwind-cleanup failure and
stale spec metadata. `make test-compiler-pipeline` remains blocked by unrelated
stale WASM resource-lifecycle ratchet entries whose formerly failing cases now
pass.

Fixes #3076

### DIFF
--- a/docs/hew-language-guide.md
+++ b/docs/hew-language-guide.md
@@ -137,12 +137,12 @@ fn main() {
 
 Use `as` for all defined numeric conversions: widening, narrowing, int↔float, and signed↔unsigned. Integer narrowing keeps the target width's low bits. Integer→float and float→float casts round to the target float type. Float→int casts truncate toward zero for in-range finite values and use defined saturation for every edge case:
 
-| Source value | Signed result | Unsigned result |
-| --- | --- | --- |
-| In-range finite | Truncated toward 0 | Truncated toward 0 |
-| `+Inf` or greater than the target maximum | Target maximum | Target maximum |
-| `-Inf` or less than the target minimum | Target minimum | `0` |
-| `NaN` | `0` | `0` |
+| Source value                              | Signed result      | Unsigned result    |
+| ----------------------------------------- | ------------------ | ------------------ |
+| In-range finite                           | Truncated toward 0 | Truncated toward 0 |
+| `+Inf` or greater than the target maximum | Target maximum     | Target maximum     |
+| `-Inf` or less than the target minimum    | Target minimum     | `0`                |
+| `NaN`                                     | `0`                | `0`                |
 
 Use `.try_to_X()` when a conversion must fail instead of losing information. It returns `Option<X>` and produces `Some` only when the value round-trips through the target type exactly.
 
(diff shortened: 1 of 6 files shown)


slepp:
## CI evidence

- base `90224616a` (== current `origin/main` head), head `417bc3afc`.
- Jobs compared: PR run [33240601883](https://github.com/hew-lang/hew/actions/runs/33240601883) vs main's latest completed push run [33239070909](https://github.com/hew-lang/hew/actions/runs/33239070909) (main is red — `main is green` fails on this head).
- Failing job set is identical on both: `Build & test (Linux/Windows/macOS arm64)`, `Linux gates (shard 1/3, 2/3, 3/3)`, `Compiled Hew Test Results`.
- `Linux gates (shard 2/3)` (`make test`, nextest): extracted `FAIL [` IDs from `gh run view --job <id> --log-failed` for both runs — **69/69 identical, 0 unique to this PR**.
- `Linux gates (shard 1/3)`: `make test-opaque-resource-lifecycle-matrix` fails on both — stale ratchet (`std.encoding.{json,toml,yaml}.Value::public-{explicit,implicit}` now pass in wasm and need their ratchet entries removed; not this PR's doing).
- `Linux gates (shard 3/3)`: `make checked-mir-verify` fails on both — `DischargeAuthorityDrift` on `examples/v05/checked-mir/vec_index_assign_consumed_move.hew`, introduced by main's own `90224616a` (`fix(mir): make consuming call handoffs unwind-safe`), whose own commit message notes one baseline-tracked double-free remains red.
- Local spot-run in this PR's worktree: `cargo nextest run -p hew-types` (touched crate) reproduces the same single failure as main: `hew-types::coverage is_allowance::record_is_record_rejected`.

**Verdict: not a new cause of failure.** All red is inherited from `origin/main@90224616a`. `fix-main` label applied.

PR #3131, opened 2026-08-29
refactor: delete the SIR shadow lane and the THIR layer

slepp:
## Why

The compiler carried two lowering ladders at once: the established HIR→MIR path and a `--sir-shadow` lane that re-lowered every function through SIR and diffed the result, plus a THIR layer that no consumer read. Both were second authorities over the same facts: every ownership or lowering change had to be made twice and gated twice, and the shadow corpus gate was the slowest of the CI shards. The IR plan commits to one ladder (HIR → SIR → MIR) with no shadowing.

## What

- **cli**: delete the `--sir-shadow` lane, its corpus script and Makefile gate; `--sir-shadow` is now rejected with a diagnostic (negative control kept in `args.rs`). `--dump-sir` reports exactly the subset it emits.
- **mir**: delete the THIR layer and its re-flatten bridge; mark the surviving checked-statement re-flatten seam.
- **sir**: select the entry from HIR's resolved declaration id, build the callable-to-body association once, lower bodies on demand from the entry and report every refusal; pin the rule that refuses a generic instance.
- **ci**: the repeated-compile determinism assertions the shadow corpus used to carry become their own gate (`compile-determinism-verify`, `compile-determinism-selftest`) in shard 2.
- **docs**: the ladder docs describe the compiler that exists; doc comments say which lowering path they describe.

Rebased over the build-authority restructure: main's `sir-shadow-verify` target and `scripts/sir-shadow-corpus.sh` are deleted here as well.

## Test

`make test` failure set compared by ID against the main baseline: no failure is new to this branch; the strict-lane `--dump-sir` outputs are byte-identical to the pre-change commit across the corpus. New: `compile-determinism-selftest`, `sir` entry-refusal pin (`test(sir): pin the entry rule that refuses a generic instance`), `test(cli): anchor the strict-lane assertions on what it realizes`.


slepp:
## CI evidence — #3131 vs main

**Runs compared**
- PR: `refactor/sir-hygiene` @ 88b6bd64c, run [33250985698](https://github.com/hew-lang/hew/actions/runs/33250985698)
- main: 90224616a, run [33239070909](https://github.com/hew-lang/hew/actions/runs/33239070909)

**Failing job names — identical set on both**
`Build & test (Windows)`, `Build & test (macOS arm64)`, `Build & test (Linux)` (roll-up), `Linux gates (shard 1/3)`, `Linux gates (shard 2/3)`, `Linux gates (shard 3/3)`.

**Per-job ID diff** (`gh run view <id> -R hew-lang/hew --log-failed`, ANSI/timestamp-stripped, `comm -3` against the same job on main):

| Job | PR fail IDs | main fail IDs | unique to PR |
| --- | --- | --- | --- |
| Build & test (Windows) | 46 | 46 | none |
| Build & test (macOS arm64) | 47 | 47 | none |
| Linux gates (shard 1/3) | ratchet script failure, no nextest IDs | same | none — `test-opaque-resource-lifecycle-matrix` fails with `Wasm runtime failures now pass; remove their ratchets: ['std.encoding.json.Value::public-explicit', 'std.encoding.json.Value::public-implicit', 'std.encoding.toml.Value::public-explicit', 'std.encoding.toml.Value::public-implicit', 'std.encoding.yaml.Value::public-explicit', 'std.encoding.yaml.Value::public-implicit']` on both runs, byte-identical message |
| Linux gates (shard 2/3) | 69 nextest FAIL | 69 nextest FAIL | none |
| Linux gates (shard 3/3) | 51 oracle FAIL entries (`scripts/fuzz` corpus + `CF-*` crash-fixture cases) + `checked-mir-verify` → `MIR kind: DischargeAuthorityDrift` | same 51 + same `DischargeAuthorityDrift` | none |

No job on the PR fails for a reason naming sir/shadow/thir/determinism, and no job fails on the PR that passes on main.

**Local spot-runs** (worktree `.claude/worktrees/refactor+sir-hygiene`, HEAD 88b6bd64c, via `in-wt.sh`):
- `cargo nextest run -p hew-cli --test sir_lower_e2e --profile ci --no-fail-fast` → 14/14 passed
- `cargo nextest run -p hew-sir -p hew-mir --profile ci --no-fail-fast` → 2028/2028 passed
- `make checked-mir-verify` → fails locally with the same `E_MIR_CHECK … MIR kind: DischargeAuthorityDrift` seen on both CI runs (reproduces the pre-existing main failure, not a PR regression)
- `make compile-determinism-selftest` → 8/8 passed standalone

**Gate-wiring finding (not a red check, but worth fixing before merge):**
Commit 4efe83d91 ("ci: restore repeated-compile determinism as its own gate") adds `compile-determinism-verify` / `compile-determinism-selftest` to the Makefile but does not add either to `ci-shard-1`, `ci-shard-2`, or `ci-shard-3`, and no `.github/workflows/*.yml` invokes them directly (`grep -rn compile-determinism .github/workflows/` → no match). The commit message says this re-lands the assertions the deleted SIR shadow-lane corpus harness used to carry and that "nothing else covers that" — that's true, and it still isn't covered: the gate is defined and self-tests clean (8/8 above) but is unreachable from CI, so it never runs on this or any future PR. Recommend adding both targets to `ci-shard-2`'s dependency list (matching the commit's own stated intent) in a follow-up before relying on it.

**Verdict:** not a new cause of failure. Every job that is red on #3131 is red on main for the identical reason (same nextest/oracle IDs, same ratchet message, same MIR verifier drift). The PR does not introduce new failures. Separately, the determinism gate the PR adds is not wired into any CI job and should be connected before it can be trusted to catch regressions — this is a gap, not a regression, and does not block on its own but should be fixed.

Not merging — evidence only, per dispatch instructions.

slepp:
Rerun 33254375764 (after wiring the determinism gates into ci-shard-2 in 30cb5c39d): shard 2 has 59 `FAILED` test names, every one present in the main baseline; the gates after `test` in `ci-shard-2` (determinism, stdlib-lint, hew-fmt-property) do not execute because make stops at the red `test` target — true for main as well. Not a new cause of failure; merging.

PR #3133, opened 2026-08-29
fix(types): reject identity comparison on record values

slepp:
## Why

`p is q` on a `type Point { x: i64; y: i64; }` declaration passed the type
checker and then died in the codegen front with

```
E_CODEGEN_FRONT_FAIL_CLOSED: codegen-front validation failed: fail-closed: IdentityCompare lhs must be a pointer or integer value
```

with no source location and no remedy. That is a compiler-invariant message,
not a user diagnostic. The checker's identity-allowance predicate claimed
`type Foo { ... }` declarations were Rc-backed and therefore identity-bearing;
under the v0.5 value model (`docs/v05/ownership.md`) a record is a
copy-on-write value with structural `==` and no pointer identity, which is
exactly why the codegen front has no `IdentityCompare` lowering for its
representation.

## What

- **types**: `is_identity_capable` no longer admits `TypeDefKind::Struct`, so
  `is` on a record is rejected at the use site with `E_IS_VALUE_TYPE`. The
  checker is the last word; the codegen-front legality check becomes an
  unreachable backstop. Machines, actors and enums stay in the allowance set —
  only the record answer is settled here.
- **types**: the rejection reports once per `is` expression when both operands
  are the same value type. Two carets carrying a byte-identical message about
  one type read as two separate bugs; operands of different value types still
  get one diagnostic each.
- **types**: the message names the offending type and points at `==`, the
  operator that does compare two records, instead of talking about
  "heap-backed types".
- **spec**: the `is` entry in the operator-precedence table now names record
  declarations among the rejected operands and says why.
- **hir / lsp / sandbox**: four fixtures used a record receiver the checker now
  rejects, and move to a receiver that still reaches the behaviour they pin —
  the HIR type-pattern vertical and the LSP `is` fixture to an enum, and the
  sandbox profile's `is` rejection and its parity-ratchet row to `Vec` handles,
  so the sandbox's own reserved-feature rejection is still what fires.

## Test

- `hew-cli::is_value_type_reject_e2e` — the issue's repro through `hew check`:
  the rejection appears, it is attributed to the user's source line, and
  `E_CODEGEN_FRONT_FAIL_CLOSED` never surfaces. Negative controls in the same
  file: `==` on the same two records checks clean, and `is` on two `Vec`
  handles still checks clean.
- `hew-types::coverage is_allowance` — `record_type_is_record_type_rejected`,
  `record_type_rejection_names_the_type_and_suggests_equality`,
  `record_type_rejection_is_reported_once_per_expression`,
  `distinct_value_type_operands_are_each_reported` (counter-case for the
  de-duplication), and `record_type_structural_equality_still_accepted`. The
  existing actor-ref / `Vec` / `HashMap` / `HashSet` / `bytes` acceptances are
  unchanged.

### DIFF
--- a/docs/specs/HEW-SPEC-2026.md
+++ b/docs/specs/HEW-SPEC-2026.md
@@ -5072,7 +5072,7 @@ The methods `.try_to_i8()`, `.try_to_i16()`, `.try_to_i32()`, `.try_to_i64()`, `
 7. Bitwise XOR: `^`
 8. Bitwise OR: `|`
 9. Relational: `<`, `<=`, `>`, `>=`
-10. Equality: `==`, `!=`, `is` (`is` = reference identity on heap handles; `expr is TypeName` = static-only type check; rejected on string/scalars/records/tuples; regex matching is via `Pattern.is_match`)
+10. Equality: `==`, `!=`, `is` (`is` = reference identity on heap handles; `expr is TypeName` = static-only type check; rejected with `E_IS_VALUE_TYPE` on scalars, `string`, tuples, and `type Name { ... }` record declarations — records are copy-on-write values with no pointer identity, so `==` is the comparison for them; regex matching is via `Pattern.is_match`)
 11. Logical AND: `&&`
 12. Logical OR: `||`
 13. Range: `..`, `..=` (only lowered inside `for` loop iterables; standalone range value expressions are not lowered)
--- a/hew-cli/tests/is_value_type_reject_e2e.rs
+++ b/hew-cli/tests/is_value_type_reject_e2e.rs
@@ -1,0 +1,130 @@
+//! End-to-end coverage for `is` on a value type (#3108).
+//!
+//! `is` is reference identity on heap handles. A `type Point { ... }`
+//! declaration is a copy-on-write value under the v0.5 value model
+//! (`docs/v05/ownership.md`), so it has no identity to compare. The checker
+//! owns that answer: `hew check` must report `E_IS_VALUE_TYPE` at the `is`
+//! expression.
+//!
+//! Before the fix the checker admitted the program and it died later in the
+//! codegen front with a span-less
+//! `E_CODEGEN_FRONT_FAIL_CLOSED: … IdentityCompare lhs must be a pointer or
+//! integer value` — a compiler-invariant message, not a user diagnostic. This
+//! file pins both halves: the user-facing rejection appears, and the
+//! fail-closed backstop never surfaces.
+
+mod support;
+
+use std::process::{Command, Output};
+
+use support::{hew_binary, repo_root, strip_ansi, tempdir};
+
+/// Two records compared with `is` — the repro from #3108.
+const RECORD_IS: &str = "type Point {\n\
+     x: i64;\n\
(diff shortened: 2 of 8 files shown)


slepp:
## CI evidence pass — not a new cause of failure

**Jobs compared**
- PR run: [33266434382](https://github.com/hew-lang/hew/actions/runs/33266434382) (head `bb0e0d1d0`)
- main baseline: [33256989579](https://github.com/hew-lang/hew/actions/runs/33256989579) — commit `54238cc62`, which is this PR's own base commit (`git merge-base --is-ancestor` confirms), so it's the most recent completed main CI at or after base.

Failing job names are identical on both runs: `Build & test (Linux)`, `Build & test (Windows)`, `Build & test (macOS arm64)`, `Linux gates (shard 1/3)`, `Linux gates (shard 2/3)`, `Linux gates (shard 3/3)`, `Compiled Hew Test Results`.

**ID-level diff**

- `Build & test (Windows)`: 43 nextest `FAIL` IDs, identical set PR vs. main (empty symmetric difference).
- `Build & test (macOS arm64)`: 38 nextest `FAIL` IDs, identical set PR vs. main.
- `Linux gates (shard 2/3)`: 59 nextest `FAIL` IDs, identical set PR vs. main.
- `Build & test (Linux)`: roll-up job; its own log shows it fails solely because `RUST_GATES_RESULT=failure` (the shard 1–3 gate failures below), not an independent failure.
- `Linux gates (shard 1/3)`: same `AssertionError: Wasm lifecycle evidence failures` at the same script step, byte-identical signature PR vs. main.
- `Linux gates (shard 3/3)`: same `assert_eq_fail.hew killed by signal 6 (SIGABRT)` + linker failures, byte-identical signature PR vs. main.
- `Compiled Hew Test Results` (check-run, not a workflow job): PR shows "2808 tests run, 2800 passed, 0 skipped, 8 failed"; main shows the identical "2808 tests run, 2800 passed, 0 skipped, 8 failed." Annotation titles are the same 4 tests (each reported twice): `hew.imported_machine_transitions_and_supports_dotted_unit_construction`, `hew.imported_payload_machine_covers_construction_calls_returns_and_record_storage`, `hew.imported_private_machine_crosses_an_owned_call_boundary`, `hew.module_qualified_variant_pattern_selects_the_matching_arm_for_both_states`.

No failing ID on the PR is absent from the main baseline, and no job fails on the PR that passes on main.

**Local spot-run** (worktree `.claude/worktrees/oneshot-3108`, head `bb0e0d1d0`, via `in-wt.sh`)
- `cargo nextest run -p hew-types -p hew-hir -p hew-sandbox-wasm --profile ci`: 3896/3896 passed.
- `cargo nextest run -p hew-cli --profile ci -E "binary(is_value_type_reject_e2e)"`: 5/5 passed.
- `make playground-check` (impact-table target for the touched `hew-types` layer, per hew-ecosystem-map): succeeded, wasm pkg built cleanly.

**Verdict: not a new cause of failure.** Every failing ID/annotation on this PR's run also fails on the base-commit main run with the same signature; the touched crates and their impact-table targets pass locally at the PR head.

No merge action taken.

PR #3139, opened 2026-08-29
docs(spec): retire stale generator, harness, and THIR claims

slepp:
## Why

`docs/specs/HEW-FUTURE.md` §1.6 still listed `async gen fn` / `AsyncGenerator<Y>`
and `receive gen fn` as deferred, but both ship and run today. HEW-SPEC-2026 §11
carried a fixed 13/13 test-count claim for the two-process distributed harness
that no longer matches the harness's actual case count. HEW-SPEC-2026 also
still described THIR in the IR ladder (§8.1 diagram, §8.1 bullet, and a
changelog note) after THIR was deleted from the compiler (#3131).

## What

- `HEW-FUTURE.md` §1.6: keeps only what is actually deferred (`Lazy<T>`,
  `#[prefetch(N)]`); states that `gen fn`, `async gen fn`, and
  `receive gen fn` ship.
- `HEW-SPEC-2026.md` §11: the harness bullet now describes what it covers
  (send/ask, monitor/link, partition handling, SWIM membership) instead of
  a case count.
- `HEW-SPEC-2026.md` §8.1: the IR ladder diagram and its stage bullet now
  name SIR instead of THIR, matching the current HIR → SIR → Raw/Checked/
  Elaborated MIR ladder (`docs/internal/v05-ir-ladder.md`). The changelog
  note describing the §8 rewrite is updated to match.

Both spec files are in `.prettierignore`; only the changed lines were
touched (`git diff --stat`: 20/19 across both files, no reflow).

## Test

- Ran a three-yield `async gen fn` consumed with `for await` against the
  release `hew` binary — compiles and prints the expected sequence.
- Ran `tests/vertical-slice/accept/receive_gen_fn_string_yield.hew` and
  diffed its output against `receive_gen_fn_string_yield.expected` — exact
  match.
- Counted test functions in `hew-cli/tests/distributed_two_process_e2e.rs`:
  21, not 13 — confirms the stale count.
- `grep -n -i thir docs/specs/HEW-SPEC-2026.md docs/specs/HEW-FUTURE.md` —
  no real THIR mentions remain (one incidental substring match in
  "a *thir*d path is").

### DIFF
--- a/docs/specs/HEW-FUTURE.md
+++ b/docs/specs/HEW-FUTURE.md
@@ -99,21 +99,20 @@ Edition 2026 ships supervisor strategies (`one_for_one`, `one_for_all`,
 hot-swap upgrades, dynamic strategy changes, supervisor introspection
 APIs — are deferred.
 
-### 1.6 Generators (`gen fn`, `async gen fn`, `receive gen fn`, `Lazy<T>`, `#[prefetch(N)]`)
+### 1.6 Generators (`Lazy<T>`, `#[prefetch(N)]`)
 
-**[Scalar-parameter + fn-typed-parameter forms live in v0.5 / remaining forms deferred]**
+**[`gen fn`, `async gen fn`, and `receive gen fn` live in v0.5 / remaining forms deferred]**
 
-Zero-parameter `gen fn` functions compile and run. The LLVM coroutine machinery,
-`yield`, `.next()`, and `for x in generator()` are all live. Parameterized `gen fn`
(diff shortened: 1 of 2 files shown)


slepp:
Merge evidence: the diff touches only docs/specs/HEW-FUTURE.md and docs/specs/HEW-SPEC-2026.md (prose, no fences the doc-example gate runs); the failing jobs are main's own set at the base commit. A docs-only change cannot cause test failures.

PR #3143, opened 2026-08-29
fix(runtime): wake actors by incarnation instead of by address

### LINKED ISSUE #3069: Key runtime wakes by actor incarnation id, not by pointer
## Problem

Wake sites hand `scheduler::enqueue_resume` a bare `*mut HewActor` and nothing
else:

- `hew-runtime/src/supervisor.rs:1838,2454`
- `hew-runtime/src/reactor.rs:1053,1147,1288,1315`
- `hew-runtime/src/await_cancel.rs:226`
- `hew-runtime/src/channel_core.rs:248`
- `hew-runtime/src/task_scope.rs:933,1118`
- `hew-runtime/src/reply_channel.rs:419`
- `hew-runtime/src/hew_node.rs:824`

The pointer may name an actor that died and whose allocation was reused by a
replacement. `with_live_actor` never dereferences an untracked pointer, so the
failure is a wrong-incarnation wake: the replacement is CAS'd
`Suspended -> Runnable` with no readiness behind it and lands on the
fabricated-timeout path at `scheduler.rs:1310`. Three of the sites carry SAFETY
comments asserting that `enqueue_resume` re-validates `waiter.actor`, which it
does not do to that strength.

`mailbox_detach_await_send` (`hew-runtime/src/mailbox.rs:2926`) matches its
waiter by read-slot address, which holds only by convention.

## Expected

A wake names an incarnation. An id-keyed registry with never-reissued ids, and
a pin held across the whole `enqueue_resume` call, so a wake for a dead
incarnation is dropped rather than delivered to its successor. Waiter matching
is by id rather than slot address.

## Notes

Per DM4 the repro must be deterministic: a test hook that forces actor-slot
address reuse on demand, so the wrong-incarnation wake reproduces on every run
before the fix and cannot occur after it.

The unsafe-count waiver
(`docs/internal/waivers/runtime-unsafe-count-43907.md`) should be re-baselined
against the id registry and the pin; `spawn_serial` does not close this.

- `hew-runtime/src/scheduler.rs` — `enqueue_resume`, the fabricated-timeout
  path at `:1310`
- the call sites above, plus `hew-runtime/src/mailbox.rs:2926`
- `.tmp/orchestration/HANDOFF-2026-08-28.md` §3 item 4, §4

slepp:
## Why

Eleven readiness sources in the runtime recorded their wake target as a raw
`*mut HewActor` and woke it through `enqueue_resume`, whose liveness probe
matched the ADDRESS against the live-actor registry. That proved the address was
tracked, never that it still held the incarnation that parked. An actor that
dies while a wake is registered against it hands its box back to the allocator,
so the next spawn - a supervisor restart, most often - can occupy the same
address, and the late wake moved that stranger `Suspended -> Runnable` with no
readiness behind it. A suspending `select` resumed that way reports a timeout
that never happened.

Now a wake names an incarnation. The wake resolves the live incarnation or is
dropped, and the resolution pin is held across the enqueue, so the target cannot
die or have its address recycled mid-call.

## What

- **runtime (`lifetime/live_actors`)**: `ActorIncarnation` - the actor id paired
  with its never-reissued `spawn_serial` - plus `with_live_incarnation`, which
  pins the exact incarnation or returns `None` on an untracked id or a serial
  mismatch.
- **runtime (`scheduler`)**: `enqueue_resume_by_incarnation` is the single wake
  edge. The refusal happens before any mutation of the resolved actor, so a
  stale wake cannot leave a `pending_wake` marker on the wrong incarnation
  either. The old `pub unsafe fn enqueue_resume` is gone; the pointer entry
  survives only as `pub(crate) enqueue_resume_pinned`, reachable in production
  solely through the resolver. There is no second wake path.
- **runtime (wake sources)**: `supervisor`, `task_scope`, `reactor`,
  `await_cancel`, `channel_core`, `reply_channel`, `hew_node` and `mailbox` all
  record an `ActorIncarnation` at park time, when the actor is provably live.
- **docs**: three SAFETY comments that asserted the old edge re-validated the
  waiter now state the incarnation rule, and the runtime-unsafe-count waiver
  records that its tracked follow-up landed.

WASM is untouched: `scheduler_wasm::enqueue_resume` is a separate cooperative
twin with no address-reuse hazard (single-threaded, no supervision), and
`wasm_parity_tests` still passes.

## Test

`hew-runtime/src/wake_incarnation_tests.rs` is new: one negative and one
positive control per readiness family - timer, channel, reply, scope - plus the
node pair in `hew_node`'s test module. `TrackedTestActor::reincarnate_parked`
(`hew-runtime/src/test_actor.rs`) forces the interleaving with no allocator hook
and no threads, by destroying the registered incarnation in place and
constructing a fresh parked one at the same address under a new identity. The
five negatives were committed failing against the pointer path and pass on the
registry; the positive controls prove the same edges still wake the incarnation
that registered, so the negatives cannot be satisfied by a wake path that drops
everything.

`hew_node`'s `parked_reply_*` tests stood a `dangling_mut` pointer in for the
parked caller, which only worked while nothing read it; they now install a
tracked stub, because registration reads the caller's identity.

## Not in this change

`mailbox_detach_await_send` (`hew-runtime/src/mailbox.rs:2916`) still finds the
registration to remove by read-slot ADDRESS. That is sound today because the
detaching caller holds the slot's creator reference for the whole call, so the
address cannot be recycled underneath it - construction by convention, which is
what the issue asks to replace. It is a separate signature change (the detach
entry takes no actor) with its own oracle, and it is not on the wake edge this
change is about. Worth a follow-up rather than a rider here.

Fixes #3069

### DIFF
--- a/docs/internal/waivers/runtime-unsafe-count-43907.md
+++ b/docs/internal/waivers/runtime-unsafe-count-43907.md
@@ -102,15 +102,15 @@ Remediation that must land with the merge: update
 .github/hew-runtime-unsafe-count.txt to total=43907. (Not done in this review â
 the reviewed worktree is read-only.)
 
-NOT waived, tracked as a follow-up lane, not a blocker: 11 enqueue_resume call
-sites still resolve pointer-only through with_live_actor â
-supervisor.rs:1838,2454; reactor.rs:1053,1147,1288,1315; await_cancel.rs:226;
-channel_core.rs:248; task_scope.rs:933,1118; reply_channel.rs:419;
-hew_node.rs:824. Residual defect is a wrong-incarnation wake (a replacement CAS'd
-SuspendedâRunnable with no readiness behind it â the fabricated-timeout path at
-scheduler.rs:1310-1315), NOT a use-after-free: with_live_actor never derefs an
-untracked pointer. Three of these carry SAFETY comments still asserting
-"enqueue_resume re-validates waiter.actor", which is now the weaker claim; update
-them when the sites are converted. Also: mailbox_detach_await_send
+The follow-up lane landed (#3069). Every readiness source now records an
+ActorIncarnation (actor id + never-reissued spawn_serial) at park time and wakes
+through scheduler::enqueue_resume_by_incarnation, which resolves the live
+incarnation, refuses a serial mismatch before touching the actor, and holds the
+pin across the enqueue. The pointer-only entry survives as the pub(crate)
+enqueue_resume_pinned, reachable in production only from that resolver, so the
+GROUP B reasoning above (id-keyed registry + pin across the wake) is now the
+whole-crate rule rather than one site's. The three SAFETY comments that asserted
+"enqueue_resume re-validates waiter.actor" are rewritten to the incarnation
+claim. Also: mailbox_detach_await_send
 (mailbox.rs:2926) still matches its waiter by slot ADDRESS, safe only because the
 detaching caller holds the creator ref â convention, not construction.
--- a/hew-runtime/src/await_cancel.rs
+++ b/hew-runtime/src/await_cancel.rs
@@ -14,6 +14,7 @@ use std::ptr;
 use std::sync::atomic::{AtomicBool, AtomicI32, AtomicPtr, AtomicU64, AtomicUsize, Ordering};
 
 use crate::actor::HewActor;
+use crate::lifetime::live_actors::ActorIncarnation;
 use crate::task_scope::{hew_cancel_token_release, hew_cancel_token_retain, HewCancellationToken};
 use crate::timer_wheel::{hew_timer_wheel_cancel, hew_timer_wheel_schedule_handle};
 use crate::timer_wheel::{HewTimerEntry, HewTimerWheel};
@@ -97,7 +98,10 @@ pub struct HewAwaitCancel {
     timer_entry: AtomicPtr<HewTimerEntry>,
     timer_generation: AtomicU64,
     timer_ref_held: AtomicBool,
-    actor: AtomicPtr<HewActor>,
+    /// The waiting actor's incarnation, set once at registration and never
+    /// rewritten. `ActorIncarnation::NONE` means the wait has no resumable
+    /// continuation (a sleep, or a foreign-thread waiter).
+    actor: ActorIncarnation,
     cleanup: HewAwaitCleanup,
     source: *mut c_void,
     /// Test-only per-instance final-free probe. When a test binds one via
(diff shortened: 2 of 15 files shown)


slepp:
## CI evidence — not a new cause of failure

Jobs compared: PR run [33279062287](https://github.com/hew-lang/hew/actions/runs/33279062287) (head `a847f9589`) vs newest completed push run on `main`, [33277923207](https://github.com/hew-lang/hew/actions/runs/33277923207) (head `d6802d8f2`, one commit ahead of the `7ad50cd1` main-head noted for this task).

Failing checks on the PR: `Build & test (Linux)`, `Build & test (Windows)`, `Build & test (macOS arm64)`, `Linux gates (shard 1/3)`, `Linux gates (shard 2/3)`, `Linux gates (shard 3/3)`, plus GitHub's native `Compiled Hew Test Results` (Code Quality/CodeQL check, not a repo-defined gate).

Per-job verdict (failing test IDs extracted with `gh run view --job <id> --log-failed`, ANSI-stripped, `FAIL \[`/`TIMEOUT \[` lines, sorted, `comm -23` against main):

- **Linux gates (shard 1/3)**: same single cause on both — `make: *** [Makefile:967: test-opaque-resource-lifecycle-matrix] Error 1` (wasm-lifecycle ratchet drift, unrelated crate). Not the PR's.
- **Linux gates (shard 2/3)** (`make test`, the Rust nextest run — covers hew-runtime): 60 failing IDs on the PR, 60 on main, **identical sets, 0 unique to the PR**. No `hew-runtime` ID in either set.
- **Linux gates (shard 3/3)**: same single cause on both — `make: *** [Makefile:622: fuzz-oracle] Error 1`, same panic site `hew-runtime/src/cont.rs:1960:9`. Not the PR's.
- **Build & test (Windows)**: 47 failing/timeout IDs on the PR, 47 on main, **identical sets, 0 unique**.
- **Build & test (macOS arm64)**: 40 failing/timeout IDs on the PR, 40 on main, **identical sets, 0 unique**.
- **Build & test (Linux)**: pure roll-up gated on the three Linux-gate shards above; fails only because they fail.
- **`Compiled Hew Test Results`** (GitHub-native Code Quality/CodeQL check, `dynamic/github-code-quality/codeql`, PR-only, not part of `ci.yml`, no main-run equivalent to diff against): 8 failures, all one compile error — `E_MIR_CHECK: obligation balance` (`ObligationUnderReleased`) in `tests/hew/imported_machine_ownership_test.hew:47:20`, function `imported_payload_machine_covers_construction_calls_returns_and_record_storage`. The PR's diff touches only `hew-runtime/src/*` (wake/incarnation) and one waiver doc — no `.hew` fixtures, no `hew-mir`/`hew-sir`/`hew-hir`. This is the ownership/IR overhaul class of red that's independently in flight on `main`, not caused by this change.

Local spot-run in the PR worktree (`.claude/worktrees/runtime-lane`, exactly at PR head `a847f95897d9ca2bd885bb214143f6b55deed9b4`) — the touched crate:

```
cargo nextest run -p hew-runtime
```
`2978 tests run: 2978 passed (1 leaky), 0 skipped` — includes the new `wake_incarnation_tests` module and every touched wake source (`supervisor`, `task_scope`, `reactor`, `await_cancel`, `channel_core`, `reply_channel`, `hew_node`, `mailbox`).

**Verdict: not a new cause of failure.**

PR #3144, opened 2026-08-29
fix(types): grant Send to a machine event enum exactly when its payloads are

### LINKED ISSUE #3122: Make a machine's generated event enum sendable so the spec's actor example compiles
## Problem

The worked example in HEW-SPEC-2026 §3.11.7 — an actor with a machine field
whose handler is
`receive fn handle(event: TcpStateEvent) { tcp.step(event); }` — does not
compile:

```
spec_3_11_7.hew:30:20: error: cannot send `TcpStateEvent` to actor: type is not Send
```

The compiler-synthesized `<Machine>Event` companion enum is not `Send`, so it
cannot be a `receive fn` parameter. Advancing a machine from a message today
requires sending a scalar code and constructing the event inside the handler.

## Expected

A machine's event enum is an ordinary enum for sendability: `Send` exactly when
every payload field is. The spec's example compiles and runs, and a reject
fixture pins that an event carrying a non-sendable payload is still refused.

## Notes

`hew-types/src/check/registration.rs` (companion `<Machine>Event` synthesis and
its value class), the `Send` classification for enums in `hew-types`
(`check/admissibility.rs`), `docs/specs/HEW-SPEC-2026.md` §3.11.7,
`docs/specs/MACHINE-SPEC.md` §6.

#3073 would make this fall out of the enum desugar; this should not wait for
it.

slepp:
## Why

The compiler-synthesized `<Machine>Event` companion enum did not go through the same `Send` classification as an equivalent hand-written enum. It never registered its variant-payload field types with the trait registry, so every marker — including `Send` — derived false regardless of payload, and there was no machine-path allowance to compensate. HEW-SPEC-2026 section 3.11.7's worked example (an actor with a machine field whose handler takes the machine's own event enum) could not compile: `cannot send TcpStateEvent to actor: type is not Send`.

## What

- **types**: register the event companion's field types with `TraitRegistry`, the same call an ordinary `enum` declaration gets, at both initial declaration and re-resolved-import commit paths.
- **mir**: project a machine layout's event companion into the enum-layout view (`machine_event_enum_view`/`_views`), mirroring the existing machine-state-enum view, and thread it through the drop-synthesis seed collectors.
- **codegen**: two follow-on gaps surfaced once the checker actually let a `Send` event enum reach codegen —
  - the actor-message CBOR codec and dyn-drop synthesis pass were still built from the un-augmented enum-layout list, so a machine-owned enum's registered layout was invisible past the checker boundary.
  - the wire codec's key lookup (`xnode_registry_key`) never applied the `mc$$Name$$` machine class tag a machine's own state enum and event companion are registered under, so it silently missed every machine-owned enum, hand-written-enum-shaped lookups included. Added `resolve_machine_enum_key`, mirroring the existing MIR-side lookup authority (`find_enum_layout`)'s dual-key probe.

## Test

`hew-cli/tests/machine_event_send_e2e.rs`: `payload_free_machine_event_enum_is_send_and_runs` (the spec's worked example compiles and runs) and `machine_event_enum_with_non_send_payload_is_rejected` (negative control — an `Rc<i64>` event payload keeps the companion enum non-Send, refused at the send call site).

Fixes #3122


slepp:
## CI evidence (PR #3144)

CI does not fire on `pull_request` for this PR — `gh pr checks 3144` shows only the CodeQL dynamic check and reads misleadingly green. This evidence comes from the manual `workflow_dispatch` CI run **33280353481** on branch `fix/machine-event-enum-send` (head `3548c560f`), compared against the newest completed push run on `main`, run **33277923207** (head `d6802d8f2`).

Both runs have the same job set. Failing jobs on both sides: `Linux gates (shard 1/3)`, `Linux gates (shard 2/3)`, `Linux gates (shard 3/3)`, `Build & test (Windows)`, `Build & test (macOS arm64)`, `Build & test (Linux)` (roll-up).

Method: `gh run view <id> --log-failed` per failing job on both runs; non-empty sets confirmed; diffed by test ID separately for (a) assertion-style `... FAILED` lines and (b) crash/timeout-style nextest status lines (`SIGSEGV`/`SIGABRT`/`TIMEOUT`), since a crashing test never prints the libtest `FAILED` line. Extracted-ID totals were cross-checked against each job's nextest `Summary` line (`N failed, M timed out`) to confirm the extraction was non-lossy.

- **Linux gates (shard 1/3)**: both fail identically at `test-opaque-resource-lifecycle-matrix` (Makefile:967) — `AssertionError: Wasm lifecycle evidence failures: ... remove their ratchets: ['std.encoding.json.Value::public-explicit', 'std.encoding.json.Value::public-implicit', 'std.encoding.toml.Value::public-explicit', 'std.encoding.toml.Value::public-implicit', 'std.encoding.yaml.Value::public-explicit', 'std.encoding.yaml.Value::public-implicit']`. Byte-identical list on both. No unique-to-PR IDs.
- **Linux gates (shard 2/3)**: PR (60 failed + 2 timed out = 62) and main (61 failed + 3 timed out = 64) share the same failure set by ID — `nested_returned_string_scope_exit_is_discharged_exactly_once`, `normal_goto_prevents_later_loop_cancellation_from_releasing_index_twice`, `returned_result_branch_releases_only_the_nonreturned_string_sibling`, `mir_pins_retain_and_noncompeting_drop_authorities`, `link_remote_crash_cascade_on_{clean_exit,crash,partition}`, `remote_monitor_and_link_target_exit_during_setup_deliver_once`, `source_read_after_selection_does_not_over_release`, `raw_and_elaborated_mir_attribute_only_unselected_resource_members`, `debugger_reports_untaken_conditional_reassignment_unavailable_not_wrong`, the `extern_ownership_double_release_e2e` family, and others — plus a shared `hew-codegen-rs::exec machine_emit_exec::machine_emit_push_populates_thread_queue_in_fifo_order` **SIGSEGV** present in both runs' shard-2 logs (also reproduced locally, see below). Main additionally fails one test the PR does not (`a_machine_step_is_keyed_by_the_machine_declaration_and_its_type_arguments`) — this PR's fix removes that failure, it does not add one. Zero unique-to-PR IDs in either the assertion-style or crash/timeout-style diff.
- **Linux gates (shard 3/3)**: PR fails at `checked-mir-verify` (Makefile:996) with `E_MIR_CHECK: discharge-authority corroboration` drift on `examples/v05/checked-mir/vec_index_assign_consumed_move.hew`. Main's push run 33277923207 fails **earlier** in the same target chain, at `fuzz-oracle` (Makefile:622), and never reaches `checked-mir-verify` — that target's state is uncompared from this specific main run (`make ci-shard-*` stops at first failing prerequisite). The immediately preceding completed main run (33272364998, head `25565381d8`, 2026-08-29T20:01Z) *does* reach `checked-mir-verify` and fails with the byte-identical `E_MIR_CHECK: discharge-authority corroboration` / `vec_index_assign_consumed_move.hew` signature — confirming this is a pre-existing main failure from the ownership/IR overhaul, not caused by this PR. The `fuzz-oracle` asymmetry itself (fails on main's reference run, passes on the PR run) is a main-only failure, not a PR effect.
- **Build & test (Windows)**: PR (43 failed + 3 timed out) and main (44 failed + 3 timed out... — counts within one of each other, verified against Summary lines) share an identical FAILED-test-name set; the crash/timeout-style diff is also empty. No unique-to-PR IDs.
- **Build & test (macOS arm64)**: PR (38 failed + 1 timed out) and main (39 failed + 1 timed out) share an identical FAILED-test-name set; crash/timeout-style diff is empty. No unique-to-PR IDs.
- **Build & test (Linux)**: pure roll-up gate (`RUST_GATES_RESULT != success`) on both; not an independent evidence source.

**Local spot-run** (PR worktree `.claude/worktrees/oneshot-3122`, `cargo nextest run -p hew-types -p hew-codegen-rs -p hew-mir --no-fail-fast`): 6078/6096 passed, 18 failed. All 18 failing IDs match the shared CI failure set above (`for_in_vec_string_exec`, `generator_exec`, `sink_owned_element_exec`, `vec_owned_element_routing_exec` families, plus the `machine_emit_exec` SIGSEGV, reproduced in isolation) — none are new.

**Verdict: not a new cause of failure.** Zero test IDs are unique to the PR across every comparable job, checked separately for assertion-style and crash/timeout-style failures.

Issue #3147, opened 2026-08-30
Match blocked-sender detach registrations by id, not by slot address


slepp:
## Problem

`mailbox_detach_await_send` (hew-runtime/src/mailbox.rs:2916) finds the registration to remove by comparing raw slot addresses:

```rust
waiters.iter().position(|waiter| waiter.slot == slot)
```

`waiter.slot` is a `*mut HewReadSlot`, and read-slot allocations are recycled like any other. Nothing in the entry point itself establishes that the address it was handed still names the registration that made it. What keeps this correct today is a property of the caller rather than of the lookup: the detaching caller holds the slot`s creator reference for the whole call, so the allocation cannot be freed and handed to a new registration underneath it. That is correctness by convention, and it is the same shape of argument that #3069 replaced everywhere else.

## Expected

The detach entry point identifies its registration by an identity the registration itself carries - a registration id minted at `mailbox_await_send` and handed back to the caller - so removing a waiter never depends on an address, and a recycled slot allocation cannot make one detach remove somebody else`s registration. The signature change is the substance: the entry takes the id rather than the slot pointer, and the creator-reference discipline stops being load-bearing for the lookup.

## Notes

This is a separate change from #3069, which converted the wake edges (timer, channel, reply, node, scope, reactor, supervisor restart-await, mailbox blocked-sender) to incarnation ids and left this lookup address-keyed. It is not on the wake edge: the blocked-sender wake in `wake_blocked_sender` already resolves `waiter.sender` as an `ActorIncarnation`. Only the detach-side registration match is still by address.

Needs its own oracle: a test that frees a registration`s slot, has a second registration receive the same allocation, and asserts the first caller`s detach removes nothing.

## Outcome

A blocked-sender detach removes the registration the caller made, identified by a registration id minted when the waiter parked, never another waiter on a recycled allocation at the same address; keeping the slot alive stays the creator reference's job but stops being load-bearing for the lookup. Lands together with #3152, since both finish #3069's conversion from address identity to actor-incarnation identity and the C-ABI change is landed once.

## Acceptance

- `grep -n 'waiter.slot == slot' hew-runtime/src/mailbox.rs` returns nothing
- A hew-runtime unit test frees one registration's slot, forces a second onto the same allocation, and asserts the first caller's detach by id removes nothing; a positive control removes exactly the right waiter
- The existing mailbox_await_send / mailbox_detach_await_send tests pass, changed only for the id-carrying signature

slepp:
[...]

Issue #3074, opened 2026-08-29
Unwind with cleanup on panic() in main context, as we already do in actors


slepp:
## Problem

`panic()` behaves differently depending on where it runs. Inside an actor it
unwinds with cleanup: drop obligations discharge, resources close. In main
context the process goes down without running that cleanup.

## Expected

`panic()` unwinds with cleanup in main context too, matching actor behaviour
and the resource story in `docs/v05/ownership.md`. A trap path may leak at
abort; a `panic` is a controlled unwind and already runs as one in an actor.

## Notes

A346/A347 rejected specifying a main-context carve-out, which would make
`defer` and `close` conditional on call context.

- unwind lowering in `hew-codegen-rs/src/llvm.rs` (`C-unwind` exports, the
  landing-pad/personality path) and the main-context entry in `hew-cli`
- the `hew_panic` path in `hew-runtime/src/scheduler.rs` and
  `hew-runtime/src/cont.rs`
- `docs/specs/failure-philosophy.md`, `docs/v05/ownership.md`
- ledger Q347 / A347

slepp:
Decided (ledger A352): unify — main context unwinds with cleanup exactly like an actor. Runtime lane, sequenced with #3069.

PR #3128, opened 2026-08-29
Restore exact cross-platform test gates

slepp:
## Why

Test gates classified results by host-sensitive signals: LLVM spelling checks,
host scheduling and temporary-path assumptions, and filesystem encoding. A
filtered nextest run could also drop an active test without the gate noticing.
Gates now classify by semantic outcome and canonical test identity, validate
filtered runs against both the full workspace inventory and the selected shard
inventory, and assert target-capability cleanup rather than LLVM spelling.

## What

- Classify compiled Hew and Rust test results by semantic outcome and canonical
  test identity.
- Validate filtered nextest runs against the full workspace inventory and the
  selected shard inventory, so an active test cannot disappear silently.
- Replace LLVM spelling-sensitive checks with target-capability cleanup
  assertions.
- Keep the native and WASM lifecycle corpus authoritative, removing duplicate
  checker-only fixtures and stale expected-failure ledgers.
- Ratchet active compiler defects to exact tests and issues without allowing
  neighbouring regressions.
- Remove host scheduling, temporary-path, and filesystem-encoding assumptions
  from gate tests.

## Test

- `make test` serially, `make lint` serially, and actionlint.
- Compiled Hew O0/O2 parity and the C ABI suite.
- Windows: the ratchet units, the WebSocket suite, and the package-name oracle.
- macOS: string ownership, WebSocket accept, and the ratchet units; the
  selected end-to-end ledger matched only the exact #3079 failure.

The ratchets keep the active ownership and compiler defects visible under
#2865, #3079, #3126, and #3127. Any changed outcome or additional failure fails
the gate. #3129 tracks cleanup of generated stdlib documentation after its test
process exits.

PR #3151, opened 2026-08-30
test(runtime): control the wake edges #3143 left untested

slepp:
## Why

#3143 converted every readiness source to wake by incarnation, but its controls
covered five of the wake edges it changed. Three converted edges shipped with no
test that a stale wake is dropped: the reactor - the background-thread family,
where the registering worker and the firing thread are genuinely different
threads - the supervisor restart-await registry, and the scope-join arm, whose
every existing test passes a null actor and so only ever drove it with
`ActorIncarnation::NONE`.

An `ActorIncarnation` is an actor id paired with a spawn serial, and every
negative control shipped so far reincarnated the registrant under a wholly new
id. The resolver looks the id up first, so all of them were refused there and
none of them ever reached the serial comparison - deleting that comparison would
have left every one of them green.

Reviewing that gap turned up a third thing. A wake that wins the
`Suspended -> Runnable` CAS raises the target's `send_pin_count` and leaves a raw
pointer to it on the scheduler's global queue. Nine runtime tests fired a wake,
never took it off the queue, and then freed the stub actor - dropping the box
while a queue entry still held a pin lease on it, which is the exact allocation
invariant the incarnation-keyed wake rests on.

## What

- **runtime tests**: incarnation pairs for the reactor (readiness over a real
  socket, and both arms of the orphan close), the supervisor restart-await
  registry (the `notify_restart` drain and the teardown drain), and the
  scope-join arm.
- **runtime tests**: a replacement that keeps the dead incarnation's actor id
  and differs only in serial, so the resolver's id lookup succeeds and only the
  serial comparison can refuse the wake. It asserts the mechanism, not just the
  outcome: the reused id resolves to the live replacement, and the dead pair is
  still refused.
- **runtime tests**: `TrackedTestActor::drop` asserts the pin lease is clear, so
  a test that fires a wake and walks away now fails loudly. The nine tests it
  caught - four `parked_reply_*` in `hew_node`, two scheduler round-trips, three
  stop-path tests - dequeue before they activate, which is what a worker does.
- **runtime tests**: stub actor ids move to a disjoint high base. The live-actor
  registry is keyed by id and refuses a duplicate, and the stub counter started
  at 1 alongside a spawn allocator that also counts from 1, so a stub installed
  next to a really-spawned tree could fail to track.
- **runtime**: `enqueue_resume_pinned` becomes module-private. Only a doc comment
  kept readiness sources off it; its callers are `enqueue_resume_by_incarnation`
  and the scheduler's own tests, both in that module, so the single wake
  authority is now structural rather than advisory.
- **runtime docs**: comments that still credited the deleted pointer-revalidation
  step now state the incarnation rule - a `Waiter` SAFETY block describing a raw
  actor pointer that is two scalars now, a module doc linking a deleted symbol,
  the W6.010 block asserting the reply channel stores a caller pointer, and an
  `ActorIncarnation::NONE` doc crediting `hew_pid_make` with a refusal that lives
  in `next_actor_id`.
- **docs**: the wasm runtime twin still wakes by address. Its soundness argument -
  one activation at a time, no allocator turn between registering and firing - is
  now stated at the wake edge and catalogued in `docs/sandbox-vm-divergences.md`,
  with the conditions that would invalidate it.

## Test

Seven new pairs on the previously uncontrolled edges: `reactor_read_wake_*`,
`reactor_orphan_close_wake_*`, `reactor_accept_orphan_close_wake_*`,
`restart_await_notify_*`, `restart_await_teardown_*`, `scope_join_wake_*`, plus
`timer_wake_does_not_resume_a_reused_id_under_a_new_serial`.

The negatives are not vacuous, and the two halves of the identity are separated:

- Give the replacement the destroyed incarnation's id AND serial - the pre-fix
  world, where the wake edge cannot tell the two apart - and every
  fresh-identity negative fails.
- Delete the `spawn_serial` comparison in `with_actor_send_by_identity` and every
  fresh-identity negative stays green; only the new serial control fails. That is
  what it exists to catch.

`cargo nextest run -p hew-runtime --no-fail-fast`: 2988 run, 2988 passed. The
tripwire was committed before the tests it caught were fixed, so each of those
nine has a recorded failing state.

Two local `make test-runtime-unit` runs each failed one different test -
`actor::tests::native_ask_timeout_sets_timeout_error`, then
`supervisor::tests::role_ask_masked_id_alias_refuses_closed_never_enqueues`.
Neither touches anything this change alters. The first passes five runs out of
five in isolation and turns on a 1 ms ask deadline beating a 20 ms handler, which
a machine running several builds at once does not guarantee. Both are load, and
CI's failing-ID diff against main is the judge rather than a local run on a busy
box.

## Known remaining

The reactor's Dekker liveness handshake is still address-keyed - the
`DELIVERING_ACTOR` publication at `reactor.rs:841` and `:852` and the
`actor_snapshot_alive` gate at `:871` all match on the actor's address, not its
incarnation. That is a separate seam from the wake edge this change controls, and
#3147 covers only the mailbox half of the same pattern.

`mailbox_detach_await_send` matching waiters by slot address is #3147. The
main-context panic unwind (#3074) belongs to its own lane.


slepp:
Local full-suite evidence at `70a34bcc5`, run in a disposable worktree so the lane's own edits could not race it.

`make test`: 66 failing ids. Diffed against `baseline-failures-15f47a43e.txt`, six are not in the baseline:

```
hew-compile tests::source_in_root_cwd_outside_any_root_with_std_not_ambiguous
hew-mir::callable_identity a_machine_step_is_keyed_by_the_machine_declaration_and_its_type_arguments
hew-observe::bin/hew-observe client::tests::get_json_rejects_bare_body_without_envelope
hew-observe::bin/hew-observe client::tests::get_json_rejects_envelope_with_mismatched_schema_version
hew-pkg checksum::tests::checksum_rejects_non_utf8_names
hew-types module_registry::tests::find_enclosing_hew_root_returns_none_outside_checkout
```

That baseline predates the base commit, so those six needed a control. All six were re-run at the base `2b82c5726` in the same worktree and all six fail there as well (`6 tests run: 0 passed, 6 failed`). They arrived on main between the baseline and the base; none is in `hew-runtime`, and `hew-types`, `hew-mir` and `hew-pkg` do not depend on it. Zero failing ids are unique to this branch.

`make test-runtime-unit` at `70a34bcc5`: 2988 run, 2988 passed.

The reactor's address-keyed Dekker handshake named under "Known remaining" is now filed as #3152, alongside #3147 for the mailbox half.

slepp:
Gate evidence at `6d22fe242`, run in a disposable worktree. The head has since moved to `a9682006e`, so these cover the branch up to the accept-arm commit rather than its tip.

- `make lint`: exit 0.
- `make test-cabi`: exit 0. Neither had been run against this branch before.
- FreeBSD 15 (native, `cargo test -p hew-runtime --lib -- wake_incarnation_tests`): 11 run, 11 passed, including `timer_wake_does_not_resume_a_reused_id_under_a_new_serial`. Not a platform CI covers.

`make test-runtime-unit` failed once here, on `shutdown::tests::drain_converges_repeatedly_when_active_work_finishes_within_budget` at iteration 59. It passes five runs out of five in isolation in the same worktree at the same commit. Its oracle is a 100 ms wall budget waiting on a thread that sleeps 1 ms, repeated 256 times, and the box was running several concurrent cargo builds. Nothing in the wake path is involved - it moves `scheduler::ACTIVE_WORKERS` and calls `drain_until_idle`. Filed as #3165 together with the two load-sensitive tests already named in the Test section above.

slepp:
All four review items are addressed on `d38dc3dde`.

**1. Reactor resume arm's sleep (`reactor.rs`).** The 20 ms sleep is gone. Both incarnation arms now block on real readability (`libc::poll` with EINTR retry) and then assert the deposit landed - `hew_read_slot_status(slot) == ReadStatus::Data` - so the refusal arm can only pass when incarnation resolution is what dropped the wake. The accept arm gained the same wait and the same assertion; the shared `tcp_listener_with_pending_conn_for_test` helper keeps its own sleep, which is now backed by an assertion in these tests rather than being the thing they rely on. One timing constant survives, the poll timeout, and it is marked as a failure bound rather than a correctness guess: the wait is on actual readiness, and the bound only turns a socket that never becomes readable into a failed run instead of a hung one.

A correction on the mechanism as filed: the arm was not vacuous. A `WouldBlock` read deposits nothing and leaves the one-shot registration in place, so the pre-existing `registration_count_for_test() == 0` would have failed the no-readiness case. The sleep was still an unmarked guess and a flake source, and readiness was still never asserted directly - that is what changed.

**2. Block-send admission wake (`mailbox.rs`).** Added `blocked_sender_wake_resumes_the_registering_incarnation` beside the refusal test: a parked producer on a full bounded `Block` mailbox is resumed when the consumer frees a queue slot. Mutation evidence - with `enqueue_resume_by_incarnation(waiter.sender)` replaced by a comment, the new test fails at `test_actor.rs:340`, `blocked-sender: the registering incarnation must be resumed by its own wake, left: 3 right: 1`, while the refusal sibling still passes. The call was restored and the full crate re-run.

**3. Wasm address-keyed wake justification.** The doc section and the `scheduler_wasm::enqueue_resume` comment now name the guard that actually holds: the single worker thread, plus the E10 actor-decl admission gate that keeps `caller_actor` null so `reply_channel_wasm::hew_reply` never takes the parked branch. Both say plainly that single-threadedness alone would not carry it, because that reply fires from the callee's activation rather than the one where the caller parked. The revisit trigger now lists E10 lifting alongside threads, preemption, and out-of-activation readiness sources.

**4. Stale pointer-liveness prose.** `reactor.rs` (both sites), `hew_node.rs`, and `lifetime/live_actors.rs` now describe incarnation resolution instead of a pointer re-validation. The `live_actors` docstring is a deadlock argument, so it was re-derived rather than renamed: `with_live_incarnation` takes `LIVE_ACTORS` to resolve and pin, releases it, and only then does `enqueue_resume_pinned` -> `with_live_actor` take it again, so no lock is held across either step.

`cargo nextest run -p hew-runtime` is green and `make lint` is clean.

slepp:
## CI evidence

Compared PR run [33286581486](https://github.com/hew-lang/hew/actions/runs/33286581486) (head `d38dc3dde`) against the newest **completed** push run on `main`, run [33283677658](https://github.com/hew-lang/hew/actions/runs/33283677658) (head `bdf5412fa`, an ancestor of current main tip `6bff9d2dd`).

**Failing jobs on the PR (3):** `Build & test (Linux)`, `Compiled Hew Test Results`, `Linux gates (shard 3/3)`.
**Failing jobs on the same main run:** `Build & test (Linux)`, `Linux gates (shard 3/3)` (identical set, modulo the roll-up check-run naming).

Per job:

- **`Build & test (Linux)`**: pure roll-up (`Require every shard to complete`); fails only because `RUST_GATES_RESULT: failure` is inherited from `Linux gates (shard 3/3)`. No independent content.
- **`Linux gates (shard 3/3)`**: both runs die at the same step, `make ci-shard-3`, in the vertical-slice ORACLE gate, exit code 2. `grep -oE 'FAIL\s+\S+\.hew'` on both `--log-failed` outputs, sorted and diffed, is byte-identical: the same 74 `.hew` files fail with the same signal/verdict breakdown (`ORACLE done total=756 clean=682 frontend-reject=34 FAIL=74 (runtime-abort=12, runtime-crash=28)`) on both the PR and main. Zero unique-to-PR IDs.
- **`Compiled Hew Test Results`**: a reporting wrapper around the same shard results above; no independent failure content.

**Checks specific to this PR's changes:**
- `enqueue_resume_pinned` narrowing (`pub(crate)` → private) in `hew-runtime/src/scheduler.rs`: every call site is inside `scheduler.rs` itself (verified via `git grep`); no out-of-module or `scheduler_wasm.rs` caller. `Playground WASM build` passed on both runs.
- Local spot-run in the PR worktree (`d38dc3dde`): `cargo nextest run -p hew-runtime` → 2994/2994 passed, including `shutdown::tests::drain_converges_repeatedly_when_active_work_finishes_within_budget` (tracked as #3165) which passed cleanly here and does not appear in either CI run's failure log.
- No new failing test ID belonging to this PR's own added tests (reactor readiness/accept-readiness/orphan-close, supervisor restart-await/teardown, scope-join, serial control, block-send admission control) appears in either run's failure output.

**Verdict: not a new cause of failure.** All red jobs on the PR reproduce byte-for-byte on the newest completed main run (`bdf5412fa`); main is red on purpose from the ownership/IR overhaul (#3128 ratchet gates unaffected — no unlisted failure appeared).

Issue #3084, opened 2026-08-29
Graduate or retire the bare-variant deprecation


slepp:
## Problem

A bare enum variant in expression position emits a deprecation warning:

```
E_BARE_VARIANT_EXPR: bare variant `Foo` is deprecated; use `.Foo` when the
surrounding type selects the enum, or qualify the variant with its type
```

Nothing forces the migration, so the warning fires across real code on every
build and no build fails if it is ignored.

## Expected

The deprecation is settled for v0.6.0, one way:

- **Error.** Bare variants stop compiling; `.Variant` and `Type.Variant` are
  the language, and the diagnostic keeps the same help text.
- **Accepted.** Bare variants are legal where the surrounding type selects the
  enum, and the warning is deleted along with the deprecation.

The guide and the spec are updated to match. If the answer is "error", the
stdlib and examples move in the same change.

## Notes

v0.6.0 is where surface decisions land (D274).

- `hew-types/src/check/diagnostics.rs:14`, `hew-types/src/error.rs:1401,2036`
- `hew-wasm/src/lib.rs` (the same code surfaces in the playground)
- `hew-cli/tests/diagnostic_suggestion_truth_e2e.rs`
- `docs/specs/HEW-SPEC-2026.md` enum expression section

slepp:
Let's decide to make this an Error. Let's make all Variants need the . for now, following our dot-path changeover, this is what we'll try to normalize on.

We will see how those ergonomics work out.

PR #3154, opened 2026-08-30
Resolve a machine's own .Variant bodies when its file is imported

slepp:
## Why

A machine's own transition bodies stopped resolving the contextual dotted variant
spelling the moment the file declaring the machine was imported by another module.
The file checked clean as the compilation entry point and failed inside its own
transitions as a dependency, reporting `E_CONTEXT_VARIANT_NO_TYPE` against source
the importer cannot change. The bare spelling has been a hard error since #3084, so
a machine's transitions had no working spelling cross-module short of qualifying
every return with the type name.

## What

- **types**: the machine-body check minted its expected type from the bare
  declaration name, while the import path retires that bare `type_defs` row once
  the module-qualified owner is published (`retire_imported_type_keys`), leaving
  the contextual-variant arm with no enum-or-machine definition to find.
  Registration and the body check now mint the declaration identity through one
  helper, `declaration_identity`, so the expected type names the machine the same
  way in both positions. At the root program the helper yields the bare name, so
  the entry path is unchanged by construction.

## Test

- `tests/hew/imported_machine_context_variant_test.hew` and
  `tests/hew/aliased_import_machine_context_variant_test.hew` run a machine declared
  in `tests/hew/machine_context_variant_support/light_defs.hew`, whose transition
  bodies use the dotted spelling, through a plain and an aliased import. All four
  fail on `origin/main` with the reported diagnostic and pass here.
- `tests/vertical-slice/reject/machine_context_variant_unknown/` is the negative
  control: an undeclared variant in the same position is still refused, and the
  harness pins `E_PATH_MEMBER_NOT_FOUND` naming the qualified owner. It also fails
  on `origin/main`, where the same file is rejected for the wrong reason under a
  bare owner name.


slepp:
## CI evidence — PR #3154 vs main baseline

**PR run:** 33284739537 (head 0b7011033)
**Main baseline:** latest *completed* push run 33283677658 (head bdf5412fa — "ci: restore exact cross-platform test gates"). Main's newer head a104d8589 has a run still queued at review time, so it is not a valid comparison point per the ID-diff protocol.

### Failing checks (3, identical set on both sides)
- `Build & test (Linux)` — roll-up job; fails only because `RUST_GATES_RESULT: failure` (same cascade on PR and main).
- `Linux gates (shard 3/3)` — fails at `make ci-shard-3`'s vertical-slice ORACLE gate (`ORACLE gate: FAIL`, `ORACLE done total=756 clean=682 frontend-reject=34 FAIL=74`).
- `Compiled Hew Test Results` — test-reporter check tied to the same `Build & test (Linux)` job; no independent content.

### ID diff — vertical-slice ORACLE (`.hew` fixture IDs)
Extracted failing fixture names from `--log-failed` on both runs (69 unique IDs each, e.g. `arena_nested_vec_reuse.hew`, `assert_eq_fail.hew`, `zoo.hew` undefined-symbol-main cases, `wire_cbor_*_traps.hew`, etc.).
`comm -23 pr_ids main_ids` → empty. `comm -13 pr_ids main_ids` → empty. **Byte-identical sets, same failure reasons (same signals/panics) on both sides.** `scripts/hew-corpus-expected-failures.txt` and `scripts/hew-suite-expected-failures.txt` are unchanged by this PR (`git diff origin/main...HEAD` on both files is empty), so this is a pre-existing main red, not a ratchet regression introduced here.

### `hew-mir::callable_identity a_machine_step_is_keyed_by_the_machine_declaration_and_its_type_arguments`
Checked by ID, not by count, per the specific instruction to verify this one carefully:
- Ran on PR HEAD (`0b7011033`, worktree `machine-variant`): `cargo nextest run -p hew-mir --test callable_identity -E 'test(a_machine_step_is_keyed_by_the_machine_declaration_and_its_type_arguments)'` → **PASS**.
- Ran on the main baseline commit (`bdf5412fa`, worktree `baseline-main` checked out detached to that SHA): same command → **PASS**.
- The test is not present in either run's `--log-failed` output (`Linux gates` shards 1/2, where cargo nextest for hew-mir runs, passed on both sides).
- Verdict: this test currently passes on both the PR and the main baseline commit. The lane's earlier local baseline showing it failing on main does not reproduce against `bdf5412fa` — treat that prior observation as stale, not a live blocker.

### Blast radius (touched crate: `hew-types` only — `hew-types/src/check/methods.rs`, `hew-types/src/check/registration.rs`)
- `cargo nextest run -p hew-types` on PR HEAD → 3265/3265 passed.
- `make playground-check` (hew-wasm/playground impact-table target for lexer/parser/types changes) on PR HEAD → succeeded (wasm-pack build clean).

### Verdict
**Not a new cause of failure.** Zero IDs unique to the PR. All three failing checks trace to the same pre-existing main-red ORACLE gate failure (identical fixture-ID set, identical failure reasons). `callable_identity`'s target test passes on both sides by direct ID-level run.

PR #3164, opened 2026-08-30
fix(runtime): fail closed on a channel send the drain cannot complete

### LINKED ISSUE #3146: Native send on a full channel with no receiver parks a handler until the drain times out
## Problem

A native actor handler that sends on a full channel whose only receiver is gone
blocks forever. The handler never returns to its loop top, so the shutdown drain
cannot converge and the process exits non-zero through the drain timeout in
`hew-runtime/src/shutdown.rs`. Nothing reports the actual cause; the only signal
is `hew: shutdown drain timed out after 5s; abandoning in-flight work`.

Reduced case: a periodic handler sends into a one-slot channel that main reads
once before returning. The second send fills the slot and the third blocks.

```hew
actor Pulse {
    let ready: channel.Sender<i64>;
    var count: i64 = 0;

    #[every(1ms)]
    receive fn tick() {
        count += 1;
        println(f"tick {count}");
        ready.send(count);
    }
}

fn main() {
    let (ready_tx, ready_rx): (channel.Sender<i64>, channel.Receiver<i64>) = channel.new(1);
    let _p = spawn Pulse(ready: ready_tx, count: 0);
    let _ = ready_rx.recv();
    sleep(200ms);
}
```

Observed on Linux and Windows: exactly three ticks print, then the drain timeout,
exit 1.

The WASM twin does not behave this way; `wasm_channel_send_full_traps_instead_of_dropping`
records that wasm traps on a send to a full channel.

## Expected

A send that can never complete because the channel is full and no receiver
remains should fail closed at the send, with a diagnostic naming the channel and
the actor, rather than parking a handler forever and surfacing later as an
unattributed drain timeout. Native and wasm should agree on the disposition.

## Notes

Found while root-causing the periodic-timer quiescence probe in
`hew-cli/tests/wasi_run_e2e.rs` (#3128). The probe was rewritten to send only on
the first tick, so it no longer depends on this behaviour, but the underlying
liveness question is untouched.

slepp:
## Why

A native actor handler that sends on a full channel whose receiver never closes blocks forever. The handler never returns to its loop top, and the shutdown drain deliberately refuses to join a worker still inside a handler, so the drain cannot converge: the process exits through `hew: shutdown drain timed out after 5s; abandoning in-flight work` with nothing naming the channel or the actor.

The mechanism, ground out on main's tip: `channel.Sender.send` lowers to `hew_channel_send_layout` -> `ChannelCore::blocking_send`, which parks the scheduler worker on the core's condvar. The only edge that wakes such a producer is the receiver's `close_stream`. In the reduced case that edge never fires before the drain deadline: the receiver's close runs at `main`'s exits, and `main` cannot reach an exit while the drain is waiting on the handler that is parked inside it. Closing the receiver by hand makes the same program run indefinitely, which isolates the park as the liveness defect rather than the close protocol.

## What

A send parked on a full ring now fails closed once the runtime reaches the drain phase, naming the channel's capacity, its queue depth, and the sending actor. That is the verdict the drain reaches five seconds later anyway, delivered deterministically and with attribution instead of as an anonymous timeout. Backpressure while the runtime is running or quiescing is untouched: a full ring with a receiver still draining it still parks and resumes.

- **runtime/channel**: `ChannelCore::blocking_send` reads the shutdown phase in its full-ring branch and fails closed at `PHASE_DRAIN` or later. The condvar wait becomes a bounded wait matching the drain's own poll cadence, marked as a shortcut with its WHY / WHEN obsolete / WHAT - shutdown publishes a phase, not a wake edge, and the real fix is a suspending ramp for `channel.Sender.send` like the one `channel.Receiver.recv` already has.
- **runtime/channel**: the native/wasm32 disposition on a full channel is stated once, at the wasm trap site. Both targets fail closed and neither drops the element silently; wasm32 has no send park at all, so it is strictly stricter than native, and the two converge when the suspending send ramp lands. `docs/sandbox-vm-divergences.md` is the sandbox VM's contract rather than the wasm32 runtime twin's, and carried no entry for this, so the statement lives beside the code it constrains.

`PHASE_DRAIN` becomes `pub(crate)` so the predicate is a named phase rather than a literal.

**#3147 is not fixed here and this PR does not close it.** The registration id the issue asks for has to be minted at `mailbox_await_send` and handed back through `hew_actor_await_send_by_id` / `hew_actor_detach_await_send_by_id` (`actor.rs`), `actor_await_send_pinned`, and `hew_supervisor_role_await_send` (`supervisor.rs`), then through the codegen suspend ramp (`suspend.rs`, `runtime_abi.rs`, `llvm.rs`) and the JIT symbol classification. Two of those files belong to other live lanes. No mailbox-local scheme substitutes: any identity read back through the slot pointer collapses to address identity the moment the allocation recycles, which is the case the issue's oracle is about. `ChannelCore::detach_consumer` and `detach_producer` are the same defect class and want the same treatment.

## Test

- `hew-cli/tests/channel_send_shutdown_e2e.rs::full_send_at_shutdown_fails_closed_with_named_channel` - the issue's reduced case, asserting the named diagnostic and the absence of the drain timeout. Fails on `origin/main`'s tip with `hew: shutdown drain timed out after 5s` after 6.5s; passes on this branch in 1.7s.
- `...::full_send_with_draining_receiver_still_backpressures` - the negative control: eight items through a one-slot channel with the handler parking between each receive. Passes on both, so the phase check cannot be firing while the runtime is running.

Fixes #3146

### DIFF
--- a/hew-cli/tests/channel_send_shutdown_e2e.rs
+++ b/hew-cli/tests/channel_send_shutdown_e2e.rs
@@ -1,0 +1,130 @@
+//! Executed regression for #3146: a blocking channel send that is still full
+//! when the runtime enters shutdown must fail closed with an attributed
+//! diagnostic instead of parking its handler until the drain deadline.
+//!
+//! The fixture spawns a periodic handler that sends into a one-slot channel
+//! whose only receiver is a `main` local that reads once and never closes. The
+//! third send finds the ring full and, before the fix, parked the scheduler
+//! worker inside the handler: the shutdown drain refuses to join a worker still
+//! in a handler, so the process died through the 5s drain timeout with nothing
+//! naming the channel or the actor.
+
+mod support;
+
+use std::fs;
+use std::process::Command;
+
+use support::{hew_binary, repo_root, require_codegen};
+
+/// A periodic handler filling a one-slot channel that `main` reads once.
+const FULL_SEND_AT_SHUTDOWN_SOURCE: &str = r#"import std.channel.channel;
+
+actor Pulse {
+    let ready: channel.Sender<i64>;
+    var count: i64 = 0;
+
+    #[every(1ms)]
+    receive fn tick() {
+        count += 1;
+        ready.send(count);
+    }
+}
+
+fn main() {
+    let (ready_tx, ready_rx): (channel.Sender<i64>, channel.Receiver<i64>) = channel.new(1);
+    let _p = spawn Pulse(ready: ready_tx, count: 0);
+    let _ = ready_rx.recv();
+    sleep(200ms);
+    println("main done");
+}
+"#;
+
+/// A send parked on a full channel when shutdown starts fails closed, naming
+/// the channel, rather than surfacing as an unattributed drain timeout.
+#[test]
+fn full_send_at_shutdown_fails_closed_with_named_channel() {
+    require_codegen();
+
+    let dir = support::tempdir();
+    let source = dir.path().join("full_send_at_shutdown.hew");
+    fs::write(&source, FULL_SEND_AT_SHUTDOWN_SOURCE).expect("write full-send fixture");
+
+    let mut command = Command::new(hew_binary());
+    command.arg("run").arg(&source).current_dir(repo_root());
+    let output = support::run_bounded_command(command, "hew run full_send_at_shutdown");
+
+    let stdout = String::from_utf8_lossy(&output.stdout);
+    let stderr = String::from_utf8_lossy(&output.stderr);
+
+    assert!(
+        stdout.contains("main done"),
+        "main must reach its end; stdout:\n{stdout}\nstderr:\n{stderr}",
+    );
+    assert!(
+        !stderr.contains("shutdown drain timed out"),
+        "a full send must not strand the drain; stdout:\n{stdout}\nstderr:\n{stderr}",
+    );
+    assert!(
+        stderr.contains("channel send abandoned during shutdown"),
+        "the abandoned send must name itself; stdout:\n{stdout}\nstderr:\n{stderr}",
+    );
+    assert!(
+        stderr.contains("capacity 1"),
+        "the diagnostic must name the channel's capacity; stdout:\n{stdout}\nstderr:\n{stderr}",
+    );
+}
+
+/// Negative control: backpressure under a live, draining receiver in the
+/// running phase still parks and resumes. A shutdown-phase check that fired
+/// while the runtime is running would turn this into a failure.
+#[test]
+fn full_send_with_draining_receiver_still_backpressures() {
+    require_codegen();
+
+    let dir = support::tempdir();
+    let source = dir.path().join("full_send_backpressure.hew");
+    fs::write(
+        &source,
+        r#"import std.channel.channel;
+
+actor Pump {
+    let out: channel.Sender<i64>;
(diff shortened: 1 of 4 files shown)


slepp:
Correcting one claim in the Why above. It said the emitted LLVM for `main` carries no `hew_channel_receiver_close` at all. That was a grep scoped to `@main`'s printed body: when the pair returned by `channel.new` stays a tuple local, the drop is indirected through a generated helper (`__hew_tuple_drop_inplace_tuple_Sender...Receiver...`) that is called from every live exit and does call `hew_channel_receiver_close` then `hew_channel_sender_close`. Verified on this base and on the ownership lane's build, and a destructured `tx`/`rx` binding emits the closes inline instead.

So there is no missing-drop defect, and nothing was filed for one. The liveness defect this PR fixes is unchanged: the close runs at `main`'s exits, and `main` cannot reach an exit while the drain waits on a handler parked inside a full-ring send. The tests and the fix stand as written.

slepp:
CI evidence — jobs compared: PR run 33284850414 (head 2e26e0583) vs main run 33283677658 (head bdf5412fa), both completed.

**Build & test (Linux)** (FAIL, both): pure roll-up (`Require every shard to complete`); fails only because `RUST_GATES_RESULT=failure` — no independent signal, downstream of Linux gates (shard 3/3).

**Linux gates (shard 3/3)** (FAIL, both): `gh run view --job 99186149076/99186920108 --log-failed`, grepped `FAIL <name>.hew` lines from the corpus ratchet step. PR: 69 IDs; main: 69 IDs. `comm -23`/`comm -13` diff: empty both directions — identical failing `.hew` set on both runs (SIGSEGV/SIGABRT/SIGILL traps and `undefined symbol: main` linker failures unrelated to channels, e.g. `arena_nested_vec_reuse.hew`, `bytes_index_oob_traps.hew`, `wire_cbor_decode_malformed_traps.hew`).

**Compiled Hew Test Results** (FAIL, both — a posted check-run, not a workflow job): `gh api check-runs/{99196559174,99195939647}` → both title "2808 tests run, 2800 passed, 0 skipped, 8 failed." Annotations diffed by test id (`gh api .../annotations`): same 4 failing test names on both — `hew.imported_machine_transitions_and_supports_dotted_unit_construction`, `hew.imported_payload_machine_covers_construction_calls_returns_and_record_storage`, `hew.imported_private_machine_crosses_an_owned_call_boundary`, `hew.module_qualified_variant_pattern_selects_the_matching_arm_for_both_states` — all `E_MIR_CHECK: obligation balance … envelope … leak` in `tests/hew/imported_machine_ownership_test.hew`, unrelated to channels/shutdown. (Annotation ordering of the "never called" warnings differs run-to-run — cosmetic reordering of the same fixed test set, not a new failure.)

No ID unique to the PR run in any job. Zero-failure ratchet check: the corpus-selection failing set is byte-identical to main's, so this PR does not surface an unlisted failure against the #3128 expected-failure manifests.

Local spot-run in the PR worktree (head 2e26e0583, `bash in-wt.sh .claude/worktrees/channel-fix …`):
- `cargo nextest run -p hew-runtime` — 2978/2978 passed.
- `make observe-functional-test` (impact-table target for `hew-runtime` internals) — 2/2 passed.

**Verdict: not a new cause of failure.**

PR #3170, opened 2026-08-30
fix(runtime): keep crash reclamation the sole owner of an abandoned coro frame

### LINKED ISSUE #2865: receive-generator fault/teardown path double-frees (glibc tcache) on Linux
## Summary

The receive-generator fault and teardown-fault paths double-free heap memory. glibc detects it on Linux as `free(): double free detected in tcache 2` (SIGABRT on a worker thread); macOS's allocator does not flag it, so it was latent until a native-Linux run surfaced it. This is a real memory-safety bug on a crash/teardown path, present on the release line (not introduced by any recent change).

## Reproduce

The vertical-slice fixtures `receive_gen_fn_fault_trap` and `receive_gen_fn_teardown_fault` both abort with the double-free on a native x86_64 Linux host. Running the compiled fixture binary directly:

```
free(): double free detected in tcache 2
```

with or without a virtual-memory ulimit — the cap is not involved. gdb backtrace lands in `malloc_printerr` / `tcache_double_free_verify` on the worker thread. The fixture binary is a release build (unsymbolized), so the exact HIR/MIR drop site needs a debug/`-g` rebuild to pin.

## Where to look

A receive `fn` (generator) that faults, or whose teardown faults, appears to release a payload or frame owner twice on the fault-cleanup path — the same class of exactly-once-discharge concern as the enum/record composite drops, but on the generator fault/teardown lowering. Symbolize the aborting frame with a debug build, identify the two release sites, and make the fault-cleanup path discharge each owner exactly once.

## Related harness gap (separate, smaller)

`tests/vertical-slice/run.sh` masks this poorly: the double-free exits 134, which is also the status the harness expects for a *designed* trap, so the status check reports PASS; the next bare `grep -qF ... "$stderr"` content assertion then fails with no message under `set -euo pipefail`, silently killing the whole run. Give those top-level content assertions an explicit failure message so a real crash is reported at its fixture instead of looking like the run silently stopped.

Target: fix before the v0.6.0-rc1 tag (memory safety).

slepp:
## Why

`receive_gen_fn_fault_trap` aborted before the stream-fault diagnostic it is
supposed to produce, with `tracked coroutine destroy returned with a mismatched
active-frame stack` followed by a misaligned-pointer dereference whose address
decoded as ASCII filesystem-path bytes - reused heap read back as a
cleanup-registry pointer. A `receive gen fn` producer that traps mid-stream is
abandoned by the language unwind while its coro frame's activation record is
still standing, and on that same unwind edge the pump's generated cleanup
landing pad drops its `Generator` and drives `hew_cont_destroy` into the frame.
The destroy re-entered the abandoned frame's cleanup outline and freed it, so
crash recovery then read the frame header of freed storage. Two authorities
claimed one frame: the abandoned activation record the drain reclaims, and the
destroy edge. The suspended-teardown twin was green because a parked generator
is off the active-frame stack, so only the running-producer leg was affected.

The drain's quarantine (#2865's earlier fix) closes the same hole, but only for
the window after `catch_unwind`; the landing pad reaches the destroy before that
window exists.

## What

- **hew-runtime**: `hew_cont_destroy` now refuses any frame the crash-reclamation
  authority owns, expressed as `frame_is_crash_owned` - a live activation record
  on this worker's active-frame stack, or membership in the drain's quarantine.
  The activation record is the frame's owner from the unwind onward and the
  quarantine takes over from it, so there is one authority across the whole
  recovery instead of two windows to coordinate. The quarantine-only skip it
  replaces is gone.
- **tests/vertical-slice**: `assert_receive_gen_stream_faulted` loses its
  known-failure branch and its `2865` argument; both fault fixtures are now held
  to the same disposition - the stream-fault diagnostic, exit 134, and the exact
  values yielded before the fault.
- **tests/fuzz-oracle**: `receive_gen_fn_fault_trap`'s expected-failures entry
  now describes the designed abort rather than the corruption; the fixture still
  terminates on SIGABRT by design, so it stays listed.
- **LESSONS.md**: extends `crash-recovery-frame-owner-is-single-authority` with
  the unwind-edge leg and the activation-record keying.

## Test

- `hew_cont_destroy_skips_a_frame_abandoned_mid_execution` (hew-runtime
  `cont::tests`) plants a `CoroSplit`-shaped destroy outline on a tracked frame
  that still carries its activation record and asserts the outline is never
  entered, that the record survives untouched, and - as a counterfactual - that
  the same handle with no record still reaches its outline, so the guard is the
  record and not the pointer. With the predicate reverted to quarantine-only it
  fails with the exact production panic, `tracked coroutine destroy returned with
  a mismatched active-frame stack`.
- `receive_gen_fn_fault_trap` and `receive_gen_fn_teardown_fault` both reach
  `receive-gen stream: producer actor ... stream faulted` and exit 134 under
  `make test-vertical-slice`.

### Sanitizer evidence

The fixture compiled against an ASan-instrumented `libhew.a` and run with
`ASAN_OPTIONS=detect_leaks=0:use_sigaltstack=0` (the fixture aborts by design, so
LSan never runs). Before is this branch with `frame_is_crash_owned` reverted to
`frame_is_quarantined`, which reproduces main's behaviour per the bisect.

Before - `heap-use-after-free`, read of 8 bytes:

```
ERROR: AddressSanitizer: heap-use-after-free
READ of size 8 ... in core::ptr::read_unaligned::<*mut CrashCleanupRegistry>
    cont::take_frame_cleanup_registry      cont.rs:669
    cont::run_frame_crash_cleanups         cont.rs:821
    cont::drain_active_coroutine_frames_excluding
    scheduler::activate_queued_actor       scheduler.rs:3605
freed by thread T1 here:
    cont::free_frame_allocation            cont.rs:599
    hew_cont_frame_free                    cont.rs:581
    __hew_gen_body_Emitter__recv__ticks_0.destroy
    hew_cont_destroy                       cont.rs:1992
    hew_gen_coro_destroy                   cont.rs:2068
    Emitter__recv__ticks
```

After - no ASan diagnostic; exit 134 with stdout `1` and the designed
`receive-gen stream: producer actor 1 crashed or was torn down; stream faulted`.
Exit 134 is the fixture's intended disposition, not a failure. This is one-off
evidence; the fixture is not wired into the `asan-fixtures` gate.

### Failing-ID diff

`make test` on this branch against a control run of `origin/main` (bdf5412fa) in
a scratch worktree on the same host. Both sides report the same two unexpected
failures and nothing else:

```
hew-compile :: tests::source_in_root_cwd_outside_any_root_with_std_not_ambiguous
hew-types   :: module_registry::tests::find_enclosing_hew_root_returns_none_outside_checkout
```

Both are artifacts of running inside a worktree that is itself in a checkout.
The delta is empty. `make lint` and `cargo clippy --workspace --tests` in CI JSON
mode are clean.

## Quality Checklist

- [x] No new `.ok()?` or `unwrap_or_default()` in codegen without `// JUSTIFIED` comment - n/a, no codegen change
- [x] New allocations have cleanup paths for sync, async, and actor shutdown contexts - n/a, no new allocation; this removes a free, leaving the drain as the frame's single free authority
- [x] Serialization changes include round-trip encode/decode tests - n/a, no serialization change
- [x] New runtime behavior has native/WASM execution coverage where supported - n/a for WASM: the skip is keyed on an activation record, and on wasm32's panic=abort build no unwind can abandon a frame with its record standing, so a destroy never observes one. No new `hew_*` export.
- [x] No duplicated logic - `frame_is_crash_owned` composes the existing `active_coroutine_contains` and `frame_is_quarantined` rather than adding a third ledger

Fixes #2865


slepp:
## CI evidence: not a new cause of failure

Compared PR run [33286310766](https://github.com/hew-lang/hew/actions/runs/33286310766) (head `8e2f5a22`) against the newest **completed** push run on `main`, run [33283677658](https://github.com/hew-lang/hew/actions/runs/33283677658) (head `bdf5412fa`, 2026-08-30T00:35:27Z). `main` tip is `4f0c50fea`, but its push CI run was still `pending`/in-progress at check time, so per instructions the newest completed run was used instead.

### Per-job verdict

- **Build & test (Linux)** (fail on both): pure roll-up job whose only body is `for leg in BUILD_RESULT SHARD_RESULT RUST_GATES_RESULT`; on both PR and main it fails solely because `RUST_GATES_RESULT: failure` — i.e. it mirrors "Linux gates (shard 3/3)" below. No independent test IDs here.
- **Linux gates (shard 3/3)** (fail on both): the whole shard is one `make ci-shard-3` invocation ending in the `fuzz-oracle` target. On both PR and main the gate fails with the identical driver:
  ```
  UNEXPECTED REJECTIONS (not in expected-reject.txt):
    UNEXPECTED-REJECT: tests/vertical-slice/accept/file_import_fn_value_and_call_lib.hew
      ...:1:1: warning: function `twice` is never called
  ORACLE gate: FAIL
  make: *** [Makefile:667: fuzz-oracle] Error 1
  ```
  Same file, same warning-vs-linker mismatch, on both commits — pre-existing red on `main`, not introduced by this PR.
- **Compiled Hew Test Results** (fail on both): a check-run that mirrors the "Build & test (Linux)" job conclusion; all four "Compiled Hew suites" shards passed on the PR (and are not separately red on main either). No independent signal.

**ID diff:** `Linux gates (shard 3/3)` runs the `fuzz-oracle` corpus scan (756 fixtures). Extracted the failing `.hew` IDs from both logs (`grep -oP 'FAIL\s+\K\S+\.hew'`, 69 IDs each) and diffed with `comm -3`: the sets are **identical** on PR and main. No `.hew` ID is unique to the PR.

### Fault-fixture disposition (the actual subject of this PR)

`test-vertical-slice` (`tests/vertical-slice/run.sh`, with its `assert_receive_gen_stream_faulted` assertions) runs in `ci-shard-1`, which is **green** on the PR — the assertions actually executed and passed, not merely absent from a red log:
```
RUN receive_gen_fn_fault_trap
PASS receive_gen_fn_fault_trap
RUN receive_gen_fn_teardown_fault
PASS receive_gen_fn_teardown_fault
```
(`Linux gates (shard 1/3)` job log, PR run.)

Both fixtures also appear in the `ci-shard-3` fuzz-oracle corpus scan on both commits:

- **On `main` (bdf5412fa)**: `receive_gen_fn_fault_trap.hew` does **not** trap cleanly there — it hits
  ```
  thread 'hew-worker-0' (53761) panicked at hew-runtime/src/cont.rs:1960:9:
  tracked coroutine destroy returned with a mismatched active-frame stack
  ```
  i.e. the pre-fix UAF/mismatched-ownership assertion from #2865 is still live.
- **On the PR (8e2f5a22)**: both fixtures fault cleanly with `killed by signal 6 (SIGABRT)` — the expected fault-close disposition documented in `tests/fuzz-oracle/expected-failures.txt:57-58`.
- Neither fixture appears in `UNEXPECTED REJECTIONS` on the PR, and the `fuzz-oracle/regressions (2 files)` section printed no entries — no unexpected-pass regression, the ratchet is satisfied.

### Local spot-run (PR head `8e2f5a22`)

- `cargo nextest run -p hew-runtime` (touched crate): 2979 passed, 0 failed.
- `make observe-functional-test` (impact-table target for hew-runtime internals): 2 passed, 0 failed.

### Verdict

**Not a new cause of failure.** All three failing checks on the PR reduce to the same pre-existing `main` red (`file_import_fn_value_and_call_lib.hew` UNEXPECTED-REJECT in the fuzz-oracle gate). No test ID is unique to the PR. The PR's own fault fixtures (`receive_gen_fn_fault_trap`, `receive_gen_fn_teardown_fault`) run and pass their expected fault-close disposition, replacing the prior UAF panic in `hew-runtime/src/cont.rs`.

Issue #3120, opened 2026-08-29
Make owned Vec iteration joins and cleanup guards path-sensitive


slepp:
## Problem

Three owned Vec iteration shapes fail at distinct points in the same ownership
route: continue-edge joining, iterator guard assignment, and destructured
loop-binding cleanup.

```console
RUSTC_WRAPPER= cargo test -p hew-codegen-rs --test for_in_vec_string_exec -- --nocapture --test-threads=1
RUSTC_WRAPPER= cargo test -p hew-codegen-rs --test vec_owned_element_routing_exec vec_array_literal_owned_tuple_runs -- --nocapture --test-threads=1
```

- `for_in_continue_skips_empty_runs_correct` reports `DischargeAuthorityDrift`
  for an invalid generation join across continue.
- `vec_iter_cursor_transfers_are_path_sensitive_and_overwrite_safe` panics
  because one owner generation receives conflicting `Overwrite` and `VecIter`
  physical guards.
- `vec_array_literal_owned_tuple_runs` reports `UseAfterConsume` after
  destructuring consumes `pair` and iteration cleanup records another use.

## Expected

Continue edges join a valid live generation, each generation receives one
path-correct physical guard, and destructuring ends the consumed loop binding's
lifetime before iteration cleanup.

## Notes

#3079 tracks the broader ownership-event and drop-plan architecture.

PR #3171, opened 2026-08-30
fix(codegen): resolve JIT test runtime symbols from the test binary

### LINKED ISSUE #3142: Machine emit queue test segfaults on main
## Problem

`hew-codegen-rs::exec machine_emit_exec::machine_emit_push_populates_thread_queue_in_fifo_order` dies with SIGSEGV on current main. It is absent from the 15f47a43e failing-ID baseline, so it regressed on main between that commit and today's head; a clean origin/main worktree reproduces it with no local changes (found during #3141's baseline diff).

## Expected

The fixture compiles and runs, the per-thread emit queue preserves FIFO order, and the test passes - or, if a today's-merge behaviour change made the pin stale, the test is re-anchored on the current contract in the same change that names the cause.

## Notes

The machines skill records the emit queue as per-type with a reentrancy cap (`hew-runtime/src/machine_emit.rs`); a segfault rather than an assertion failure points at the crash-cleanup or queue-lifetime seam, not the ordering. Bisecting today's merges (#3133, #3138, #3139 docs-only, #3128-era test changes) should be quick - only two touch code.

slepp:
## Why

`hew-codegen-rs::exec machine_emit_exec::machine_emit_push_populates_thread_queue_in_fifo_order`
dies with SIGSEGV on Linux CI. It is not a recent regression: the same
`SIGSEGV` row is present in every `main` CI run sampled back to 2026-08-28.
It stayed invisible because failing-ID extraction that matches only `FAIL`
rows drops a crash row from both sides of a baseline diff.

The test JIT-compiles the emitted module and calls it in-process, so MCJIT has
to resolve the module's `hew_*` references back to the runtime linked into the
test binary. It hand-listed four symbols through `add_global_mapping` and left
every other reference to the engine's process-symbol resolver. Mach-O puts
every global symbol in the dynamic symbol table, so that resolver finds them
and the test passes on macOS. An ELF executable exports nothing unless the
link asks - `nm -D` on the test binary reports zero `hew_*` symbols - so on
Linux the resolver finds nothing, MCJIT leaves the reference at address 0, and
the JIT-compiled call faults with no diagnostic of any kind.

## What

- **codegen-rs (build)**: `export_runtime_symbols_to_jit_hosts` links this
  crate's test targets with `--export-dynamic` on Linux, giving ELF the
  property Mach-O already had. The test binary's own dynamic symbol table is
  now the single authority for which runtime symbols a JIT-executed test can
  see, on both platforms.
- **codegen-rs (test)**: deleted the four `add_global_mapping` calls and the
  inline `extern "C"` block they needed - the authority above replaces them,
  and covers every symbol the binary links rather than the four someone
  remembered.
- **codegen-rs (test)**: before invoking JIT-compiled code, every referenced
  declaration in the module is probed with `dlsym`, the same process lookup
  the engine's resolver uses. An unresolvable symbol now names itself in an
  assertion instead of faulting on a null address.

## Test

`machine_emit_exec::machine_emit_push_populates_thread_queue_in_fifo_order` is
unchanged in what it asserts and is not weakened or skipped. Both halves of
the fix carry a negative control, run locally:

- with `--export-dynamic` and no `add_global_mapping` at all, the test passes -
  the process symbol table alone resolves the module.
- with `--export-dynamic` disabled, the new `dlsym` guard fails naming
  `hew_machine_emit_push`; with the guard also removed, the same build
  reproduces the CI signature exactly (`signal: 11, SIGSEGV`), confirming the
  mechanism rather than assuming it.

A note on evidence: the crash does not reproduce on this Linux host, and the
usual suspects are ruled out - it still passes built against CI's exact LLVM
(22.1.5 upstream tarball), under `mold`, and inside an `ubuntu:24.04` container
matching CI's distro, glibc and toolchain. What the change removes is the
whole failure class: resolution no longer depends on a hand-list being
complete or on `add_global_mapping` taking effect, so whichever symbol CI
resolves to null is covered, and if any gap remains it now fails by name
instead of as a bare signal.

## Quality Checklist

- [x] No new `.ok()?` or `unwrap_or_default()` in codegen without `// JUSTIFIED` comment — n/a, no codegen source changed; the change is a build script flag and test code
- [x] New allocations have cleanup paths for sync, async, and actor shutdown contexts — n/a, no new allocations
- [x] Serialization changes include round-trip encode/decode tests — n/a, no serialization change
- [x] New runtime behavior has native/WASM execution coverage where supported; intentional WASM limitations have a typed manifest disposition and a focused diagnostic test; new `hew_*` exports are classified in `scripts/jit-symbol-classification.toml` — n/a, no new runtime behaviour and no new `hew_*` export; the symbols this test resolves are already classified `codegen-stable`
- [x] No duplicated logic — checked for existing helpers before adding new ones; this deletes a hand-list rather than adding a second one, and `add_global_mapping` had no other caller in the workspace

Fixes #3142


slepp:
## CI evidence

Compared PR head `d4fb48787` (run 33293492989) against the newest **completed** push run on `main`, `b9bf757e4bb8520e9d20785d1fc73fd3e89a9352` (run 33291112844) — main is red on purpose (ownership/IR overhaul).

PR failing checks (5): `Build & test (Linux)`, `Build & test (macOS arm64)`, `Compiled Hew Test Results`, `Linux gates (shard 3/3)`, `Test Results (macOS)`.
Main failing checks (3): `Build & test (Linux)`, `Linux gates (shard 3/3)`, `Compiled Hew Test Results`.

Per-job verdict:

- **`Linux gates (shard 3/3)`** — shared, not a new cause. Both runs fail on `make fuzz-oracle` with the same set of `.hew` library-style corpus files (`zoo.hew`, `greeting.hew`, `file_import_trait_default_local_type_lib.hew`, etc.) failing to link with `hew: linker: ld.lld: error: undefined symbol: main` — same files, same error text, on both PR and main logs.
- **`Compiled Hew Test Results`** — shared, not a new cause. Both runs report the same file, same dead-code-as-error failure mode, same count (8 annotations) in `tests/hew/imported_machine_ownership_test.hew` (functions such as `imported_private_machine_crosses_an_owned_call_boundary`, `carry_hidden`, `module_qualified_variant_pattern_selects_the_matching_arm_for_both_states`).
- **`Build & test (Linux)`** — shared roll-up failure; `RUST_GATES_RESULT=failure` on both runs, downstream of `Linux gates (shard 3/3)` above.
- **`Test Results (macOS)`** — **unique to this PR** (main: success). Ratchet reports `unexpected failure: hew-codegen-rs::exec :: machine_emit_exec::machine_emit_push_populates_thread_queue_in_fifo_order`. Raw log (`Build & test (macOS arm64)`, job 99209086954) shows the cause directly:
  ```
  thread 'machine_emit_exec::machine_emit_push_populates_thread_queue_in_fifo_order' panicked at hew-codegen-rs/tests/exec/machine_emit_exec.rs:404:9:
  the JIT-executed module references `hew_machine_emit_push`, which this test binary does not export for the JIT to resolve; calling it would fault on a null address. See `export_runtime_symbols_to_jit_hosts` in hew-codegen-rs/build.rs
  ```
  Confirmed from the PR diff itself: the deleted comment in `hew-codegen-rs/tests/exec/machine_emit_exec.rs` states `add_global_mapping` was required precisely because "Rust test binaries on macOS (and Linux with default linker flags) do not export all `#[no_mangle]` symbols to the dynamic symbol table." The PR replaces `add_global_mapping` with the `dlsym` probe and links `--export-dynamic` for the Linux test binary, which resolves the Linux case, but the macOS test binary still does not export `hew_machine_emit_push` by name — so the new fail-closed probe now (correctly) refuses instead of faulting, naming the same symbol the old hand-list used to cover on macOS.
- **`Build & test (macOS arm64)`** — unique-to-PR roll-up, downstream of the same `Test Results (macOS)` failure.

Linux confirmation (the PR's actual point): `machine_emit_exec::machine_emit_push_populates_thread_queue_in_fifo_order` **passed** on Linux, confirmed directly in the raw log (not just the ratchet summary) — `Linux gates (shard 2/3)` job 99209087090, 2026-08-30T05:12:51Z: `PASS [0.041s] hew-codegen-rs::exec machine_emit_exec::machine_emit_push_populates_thread_queue_in_fifo_order`. Reproduced locally too: `cargo nextest run -p hew-codegen-rs --test exec machine_emit_exec` → 2 passed, 0 failed, in the PR's worktree (`d4fb48787`, based on `b9bf757e4`).

**Verdict: introduces 1 unique-to-PR failing test ID** — `hew-codegen-rs::exec machine_emit_exec::machine_emit_push_populates_thread_queue_in_fifo_order` fails on macOS (`Build & test (macOS arm64)` / `Test Results (macOS)`), even though it now passes on Linux as intended. Not merging.

slepp:
## The macOS path, and a premise corrected

The earlier commit's reasoning was wrong about Mach-O. It claimed Mach-O puts every global in the dynamic symbol table, so macOS never needed help - but the comment that same commit deleted said the opposite, that Rust test binaries on macOS do not export their `#[no_mangle]` symbols for a JIT host to look up. The deleted comment is the one that matches what the machine does. Mach-O writes an export trie for an executable only when the link asks, exactly as ELF does.

So macOS gets the same single authority, not a restored hand-list: `export_runtime_symbols_to_jit_hosts` in `hew-codegen-rs/build.rs` now emits ld64's spelling of the flag, `-Wl,-export_dynamic`, alongside GNU/BSD ld's `--export-dynamic`. The test binary's own dynamic symbol table decides which runtime symbols a JIT-executed test can see, on both platforms, and the `dlsym` probe stays the thing that turns a gap into a named failure. The freebsd arm follows the same principle and is unverified - no acceptance target runs there.

## Evidence

On a macOS arm64 box, in a disposable worktree at `d4fb48787` - the pre-fix state - the probe refuses by name rather than faulting, reproducing CI verbatim:

```
thread 'machine_emit_exec::machine_emit_push_populates_thread_queue_in_fifo_order' panicked at hew-codegen-rs/tests/exec/machine_emit_exec.rs:404:9:
the JIT-executed module references `hew_machine_emit_push`, which this test binary does not export for the JIT to resolve; calling it would fault on a null address. See `export_runtime_symbols_to_jit_hosts` in hew-codegen-rs/build.rs
test machine_emit_exec::machine_emit_push_populates_thread_queue_in_fifo_order ... FAILED
```

That run is also the fail-closed proof: with the export absent, the failure names the symbol.

Same worktree moved to `c0e8e6c98` and rebuilt:

```
test machine_emit_exec::machine_emit_push_populates_thread_queue_in_fifo_order ... ok
```

And on Linux at `c0e8e6c98`:

```
test machine_emit_exec::machine_emit_push_populates_thread_queue_in_fifo_order ... ok
```

The Linux arm of the match is byte-identical to what `d4fb48787` shipped, so nothing about the Linux link changed. `make lint` is clean. The rest of the `exec` suite locally shows the known `E_MIR_CHECK` shapes main is already red on.

slepp:
## CI evidence

Compared against the newest completed push run on `main`: `3c9604734` (run [33295156422](https://github.com/hew-lang/hew/actions/runs/33295156422), also failure). PR head: `c0e8e6c98` (run [33295943164](https://github.com/hew-lang/hew/actions/runs/33295943164)).

PR failing checks (3): `Build & test (Linux)`, `Linux gates (shard 3/3)`, `Compiled Hew Test Results`.

**`Build & test (Linux)`** — pure roll-up; its "Require every shard to complete" step reports `BUILD_RESULT: success`, `SHARD_RESULT: success`, `RUST_GATES_RESULT: failure`. It fails only because `Linux gates (shard 3/3)` fails; no independent cause.

**`Linux gates (shard 3/3)`** — fails in the `fuzz-oracle` step of `make ci-shard-3`, not nextest. `ORACLE done total=758 clean=685 frontend-reject=34 FAIL=73` on both the PR and the main comparison run, and the set of failing `.hew` corpus IDs is byte-identical between the two logs (68 unique IDs, `diff` empty). Shared pre-existing red, not caused by this PR.

**`Compiled Hew Test Results`** — `2816 tests run, 2808 passed, 0 skipped, 8 failed` on both the PR and on main's head commit (`3c960473...`, check run `99219322416`). Same 4 unique failing IDs on both (each reported twice, once per shard reporter):
- `hew.imported_payload_machine_covers_construction_calls_returns_and_record_storage`
- `hew.imported_machine_transitions_and_supports_dotted_unit_construction`
- `hew.module_qualified_variant_pattern_selects_the_matching_arm_for_both_states`
- `hew.imported_private_machine_crosses_an_owned_call_boundary`

All are E_MIR_CHECK obligation-balance failures in `tests/hew/imported_machine_ownership_test.hew`, unrelated to this PR's `hew-codegen-rs/build.rs` change. Shared pre-existing red.

**Unique-to-PR failing IDs: none.**

**Named test on both platforms:**
- Linux: `PASS [0.030s] hew-codegen-rs::exec machine_emit_exec::machine_emit_push_populates_thread_queue_in_fifo_order` inside `Linux gates (shard 2/3)`'s `make test` step (job `99215495262`).
- macOS: `PASS [0.094s] hew-codegen-rs::exec machine_emit_exec::machine_emit_push_populates_thread_queue_in_fifo_order` inside `Build & test (macOS arm64)` (job `99215495307`).

**Blast radius / local spot-run** (worktree `.claude/worktrees/emit-segv` at `c0e8e6c98`, changed files: `hew-codegen-rs/build.rs`, `hew-codegen-rs/tests/exec/machine_emit_exec.rs`, `LESSONS.md` — no gate/manifest/ratchet file touched):
- `cargo nextest run -p hew-codegen-rs -E 'test(machine_emit)'` → 2/2 passed.
- `cargo nextest run -p hew-codegen-rs --no-fail-fast` (full crate, all 872 tests) → 856 passed, 16 raw `FAIL`s; all 16 IDs are listed in `scripts/nextest-expected-failures.tsv` (the ownership-overhaul ratchet — #3120 Vec-iterator path-sensitivity, generator extraction W3.053 refusals, contextual-sink guard shapes). None are new; this matches CI's green ratchet verdict in `Linux gates (shard 2/3)` where `make test` ran.
- No new link failures, no duplicate-symbol errors observed in either the PR's or main's `Linux gates` / `Build & test (macOS arm64)` logs from the `--export-dynamic`/`-export_dynamic` build.rs change.

**Verdict: not a new cause of failure.** Merging via admin (CI's `main is green` roll-up will stay red for the unrelated shared reds above; no `fix-main` label needed since this PR does not add to them).

Issue #3079, opened 2026-08-29
Derive drop plans from the ownership event stream instead of checking a template against it


slepp:
## Problem

Three authorities answer "which owners are live here":

1. **Builder ledgers** — per-feature side sets on `struct Builder`
   (`hew-mir/src/lower/mod.rs`), from which `derive_elaboration`
   (`hew-mir/src/lower/drop_plan.rs:93`) builds a LIFO drop template through a
   family of `derive_*_drop_allowed` provers plus per-exit-kind re-admission
   loops.
2. **HIR binding dataflow**, still threaded through `enumerate_exits`
   (`drop_plan.rs:14108`) although the filter now ignores it (`_state_map`).
3. **The Checked-MIR `OwnershipEvent` stream**, replayed by
   `exact_owner_states` / `validate_ownership_events`, and declared by
   `seal_checked`'s own doc to be the sole authority.

The plan is produced by (1) and then checked against (3). Composite admission
excludes anything the prover cannot positively clear, accepting a leak to avoid
a double free (`hew-mir/src/lower/composite_own.rs:2341`,
`derive_owned_record_drop_allowed`), while the verifier treats a leak as a
compile error. Every admission gap becomes a compiler-invariant failure, fanned
out once per (owner, exit edge) at `drop_plan.rs:4313`.

Repro (`.tmp/ownership-seams/rec_index_field4.hew`, reduced):

```hew
type Task { name: string; command: Vec<string>; }
type Workflow { name: string; tasks: Vec<Task>; }

fn digest_list(prefix: string, items: Vec<string>) -> string {
    var s = "";
    for i in 0 .. items.len() { s = s + prefix + items[i] + "\n"; }   // indexes
    s
}

fn render(w: Workflow) -> string {
    var d = "";
    for i in 0 .. w.tasks.len() {
        d = d + digest_list("  cmd=", w.tasks[i].command);
    }
    d
}
```

`w.tasks[i]` materializes a `Task` temp (`__hew_produced_value`) that mints an
owner; passing `temp.command` to `digest_list` counts as an owning sink, so the
temp is dropped from the drop template and the replay reports
`owned value __hew_produced_value : Task ... never released on Unwind`, once
per later call. The same program with `for item in items` in the callee
compiles (`rec_index_field3.hew`), so the callee's loop shape decides whether
the caller is legal.

## Expected

- Exit plans are derived from the event replay: `required(exit)` = the exact
  owners live at that exit, minus those whose inline release dominates it. The
[...]

PR #3140, opened 2026-08-29
refactor(mir): key callables by resolver-anchored identity

slepp:
## Why

The MIR pipeline identified functions by presentation name alone: `RawMirFunction`, `CheckedMirFunction`, and `ElaboratedMirFunction` carried `name: String` and every downstream join compared names, so two callables with equal names (imported twice, generic instances, a synthesized closure of a same-named parent) were indistinguishable. The name-keyed joins and the codegen fail-open exist to paper over that.

## What

- **mir**: `MirCallableKey { declaration: DefId, instance }` on all three MirFunction types, authored by both producers - the legacy lowerer from `HirFn.declaration` plus the mono key, the SIR bridge from `SemCallable`; synthesized callables derive a typed child identity with deterministic per-parent ordinals; a duplicate key across a module is a fail-closed diagnostic naming both symbols. `name` stays as the presentation and linkage alias.
- **mir**: two producer defects the key exposed are fixed at the producer - an imported function was lowered under two spellings of one declaration, and synthesized actor handler declarations were minted from the bare actor name, so same-named actors in different modules collided.
- **hir**: `HirMachineDecl` carries a declaration identity minted beside the function mint; machine steps no longer reconstruct identity in MIR.
- **docs**: `identity.rs` and the IR-ladder doc state what holds today - the key is anchored on the declaration identity hew-hir currently reconstructs; resolver-native minting is the tracked upstream fix.

## Test

`hew-mir/tests/callable_identity.rs`: cross-producer key equality (SIR bridge vs legacy lowerer), generic instances carry their exact type arguments, per-parent closure ordinals, the first-referencer shim pin with a two-referencer case, the actor-handler declaration pin, and the duplicate-key collision firing on a hand-built module and on the naturally occurring import shape. `compile-determinism-selftest` covers ordinal stability. Full-suite failing IDs match the main baseline.

Issue #3090, opened 2026-08-29
Arena<Vec<T>>::remove frees an invalid pointer and slot reuse segfaults


slepp:
## Problem

Removing a nested `Vec` payload from `Arena<T>` frees an invalid pointer, and
reusing the removed slot can then segfault.

```hew
import std.arena.{ Arena };

fn main() {
    let store: Arena<Vec<i64>> = arena.new();
    let key = store.insert([1]);
    store.remove(key);
}
```

```console
make hew-native
target/debug/hew build -o arena-remove repro.hew
./arena-remove
```

The program aborts with `free(): invalid pointer`. The tracked
remove-and-reuse case in
`tests/vertical-slice/accept/arena_nested_vec_reuse.hew` exits on SIGSEGV
before producing its expected output.

## Expected

`Arena<T>::remove` transfers and releases a nested `Vec` payload exactly once.
Removing a value and reusing the vacant slot completes without an invalid free
or segmentation fault.

## Notes

`Arena<Vec<string>>` reproduces the nested-payload failure. `Arena<i64>` reuse
and `Arena<string>` removal exit normally.

PR #3173, opened 2026-08-30
refactor(mir): make the ownership event replay the only cleanup authority

### LINKED ISSUE #3078: Stop calls from consuming the caller's binding
## Problem

`docs/v05/ownership.md` states that passing an owned value to a function is a
borrow and that there is no use-of-moved-value error for function calls. The
compiler does not do that. `compute_call_param_consumption`
(`hew-mir/src/lower/facts.rs:865`) scans the callee body; if the parameter is
stored, returned, sent, captured or forwarded, `call_param_owned_carrier`
(`hew-mir/src/lower/mod.rs:4042`) flips the parameter to `OwnedCarrier`, and
`hew-mir/src/lower/move_value.rs:242,285` moves the caller's argument in.

Whether the caller keeps its binding then depends on the binding's provenance:
a function parameter gets a CoW share and survives; a `match` payload is
consumed and every later use is an error. Same call, same callee, different
answer.

Repro (`.tmp/ownership-seams/owned_param_reuse3.hew`, reduced):

```hew
fn digit(path: string) -> i64 {
    let args: Vec<string> = Vec.new();
    args.push(path);                      // parameter stored -> OwnedCarrier
    match process.try_run_argv("stat", args) { .Err(_) => -1, .Ok(_) => 0 }
}

fn describe(s: Source) -> Result<string, CredError> {
    match s {
        Source.File(path) => {            // match payload, not a fn param
            let others = digit(path);
            Err(CredError.Malformed("could not determine '" + path + "'"))
            //                                                ^^^^ binding `path`
            //                                                is used after it
            //                                                was consumed
        },
        Source.Env(name) => Ok("env " + name),
    }
}
```

`.tmp/ownership-seams/owned_param_reuse2.hew` is the same program with `path`
as a function parameter, and it compiles. The callee is byte-identical in both.

## Expected

A call never consumes a non-resource argument. `call_param_owned_carrier`
remains an escape summary used to pick a cost strategy — transfer on a proven
last use, otherwise a CoW share or a materialized copy for the callee — and
decides neither legality nor the caller's lowering path. `UseAfterConsume` is
reachable only for `consume` parameters and resource kinds, per the three-walls
contract in `docs/v05/ownership.md`.

## Notes

- `hew-mir/src/lower/facts.rs:865` (`compute_call_param_consumption`),
  `hew-mir/src/lower/mod.rs:4042,14543` (`call_param_owned_carrier`),
  `hew-mir/src/lower/move_value.rs:242,285`
- `docs/v05/ownership.md` — the borrow-by-default rule and the three walls
- `docs/plans/2026-08-28-ownership-seams.md` §3 C1, §6 F3

Being fixed on `refactor/ownership-seams` as Track A3 / F3; this issue is the
tracking record.

### LINKED ISSUE #3080: Make MIR lowering linear: compute the owner-state replay once per function
## Problem

MIR lowering is superlinear in function size. `prepare_body_transfers`
(`hew-mir/src/lower/mod.rs:8150`) runs
`canonicalize_terminal_transfer_owner_ids` (`:9057`) and
`canonicalize_stale_relocation_and_reset_owner_ids` (`:9098`), each of which
calls `exact_owner_states(blocks)` / `maybe_owner_states(blocks)` — a full-CFG
replay that clones its `HashMap` state per block — inside a
`for block_index in 0..blocks.len()` loop, giving O(blocks² × instructions).
The fixpoint region around `:9373` recomputes both replays every iteration.

Measured with the release-lib compiler,
`hew check src/workflow/workflow_machine.hew` in the goobers dogfood takes
33.3 s wall and 33.25 s user CPU for one 2 K-line module, against 0.57 s for
`journal.hew` in the same project. Symbolized stack samples at 45 s and 120 s
both sit in `prepare_body_transfers`.

The dogfood Makefile also defaults to `target/debug/hew`, turning a 108 s
release build into 809 s.

## Expected

Compile time linear in function size. The owner-state replay is computed once
per function and passed by reference, updated incrementally or re-run at most
once per pass after a mutation, never per block. `exact_owner_states` and
`maybe_owner_states` are called from many sites across `drop_plan.rs` and
`mod.rs` per function and get the same treatment.

A ratcheted `bench-mir` target names the regressing function: `hew check` on
`workflow_machine.hew` under 3 s release, a full `native-loop` build under 30 s
release, with per-function lowering time under `HEW_MEASURE_TIMINGS`.

The dogfood build points at the release compiler.

## Notes

- `hew-mir/src/lower/mod.rs:8150,9057,9098,9373`
- `hew-mir/src/lower/ownership.rs` — `exact_owner_states`, `maybe_owner_states`
- `docs/plans/2026-08-28-ownership-seams.md` §5, §6 F5

Being fixed on `refactor/ownership-seams` as Track A5 (F5); this issue is the
tracking record.

slepp:
## Why

Two authorities decided cleanup. Exit drop plans came from a builder-ledger
LIFO template filtered by allow-sets, while the ownership event replay
computed the exact owner state at each exit independently. When the two
disagreed the compiler refused ordinary code, and the refusals moved whenever
either side was patched, so a program that compiled today could stop
compiling after an unrelated lowering change. A user could neither predict the
refusals nor work around them.

## What

The Checked-MIR ownership event replay is the only authority for cleanup.

- **mir**: exit drop plans are derived from the replay - the exact owners live
  at an exit whose inline release does not dominate it. `build_lifo_drops`,
  `owned_locals_exit_candidates`, the `derive_*_drop_allowed` plumbing into
  plans, and the per-exit re-admission loops are gone rather than disabled.
  Anything an allow-set used to decide moved to the mint site: a value that
  cannot be safely dropped does not mint an owner.
- **mir**: a place has at most one live exact owner generation at any program
  point, enforced by a verifier rule. Every produced-value temp gets a fresh
  local instead of reusing another owner's slot.
- **mir**: a call never consumes a non-resource argument. The callee summary
  may still choose a cost strategy, and it never changes caller legality;
  `UseAfterConsume` is reachable only through explicit consume and resource
  parameters.
- **mir, cli**: an unbalanced owner is reported once, keyed by its mint site,
  rather than once per exit. Diagnostics that state a lowering invariant go to
  an internal-error channel with a stable code, so raw `OwnerId` and
  `BindingId` debug formatting no longer reaches users.

## Test

`hew-mir` verifier tests reject a hand-built two-owners-one-place event stream
and accept the fixed lowering; generation-protocol tests cover the
transfer-before-join and distinct-places cases. The repro corpus at
`.tmp/ownership-seams` compiles with the tightened per-file expectations in
`expected.txt`, checked by `validate.sh`. Full `make test` failing IDs diff
against a control run on main's tip; the corpus and harness are mirrored in
the orchestration repo.

Closes #3078, #3079.

The compile-time work follows in its own PR and closes #3080. It is an
optimization rather than a correctness dependency, and holding this series for
it would ship nothing sooner: the pre-series baseline checks the dogfood
workflow module in 33.3 s, this series checks it in 16-28 s, and the
compile-time work drives that toward a 2 s budget.


slepp:
Rebased onto `3c9604734`. The branch was conflicting, so GitHub built no merge ref and nothing had run here; the PR now reports mergeable and CI has a head to work with.

Main's arena fix from #3090 needed no re-derivation. `a78a69e11` added `publish_consuming_match_projection` and `projected_enum_payload_is_handle_transfer` to `ownership.rs` and its call site to `pattern.rs`, and it leans on `classify_field_load`, `FieldLoadClass` and `publish_projection_source_transfer` - none of which this branch deletes. The fix replayed into the rewritten tree untouched, and its fixture `match_record_enum_field_payload_owner` passes on the new tip, as do `panic_main_unwind_runs_resource_close` from the main-panic unwind change and `machine_context_variant_unknown` from the imported-machine variant fix.

Three conflicts were real. At the five `seal_checked` call sites main had renamed `thir.statements` to `body_statements` while this branch removed the dataflow and allow-set arguments; the resolution keeps both, passing `body_statements` alone. `drop_plan.rs` resolved to the branch's rewrite, which drops roughly 18k lines main had only touched to add the resolver-anchored callable key from #3140 - that key is re-applied on top in the tip commit, so `seal_checked` and `derive_elaboration` carry it onto `CheckedMirFunction` and `ElaboratedMirFunction` and the replay-plan fixtures mint theirs through `for_test`. Third, the branch carried a merge of main, so flattening replayed the pre-merge copy of the consuming-call handoff fix and re-added its `affine_call_consume` helpers to `expr.rs` beside the pair main had already landed in `ownership.rs`; the duplicates are gone and the shape now matches both main and the pre-rebase branch tip.

Two smaller reconciliations. Main and this branch had independently retargeted the resource-marker reject fixture at different invalid targets - an alias on main, a callable here - so `run.sh` now asserts both boundaries, and both fire the diagnostic. And `style(examples): format type declarations with field terminators` dropped out as already upstream; its files are byte-identical to the pre-rebase tip.

Verified on the rebased tip: `make hew` builds, `cargo nextest run -p hew-mir` is 1868 passed with none skipped, the ownership-seam repro corpus is 41 of 41 within its expected-error ceilings, and `make lint` exits clean. A contamination audit over the thirteen files the duplicated commit touched shows no added line that is not attributable to a change main made since the merge base. The `make test` gates are left to CI.

One note for bisectability: the callable-key reconciliation is a single commit at the tip, so the replayed commits between the drop-plan rewrite and that commit do not build in isolation.

slepp:
## CI evidence — not a new cause of failure verdict: **blocked, introduces new failures**

**Compared against:** newest completed CI run on `main`, run [33300481590](https://github.com/hew-lang/hew/actions/runs/33300481590), head `585b9711273123a27a6fa19177510a16d0ad6cb7` (matches the stated `585b97112`).
**PR run:** [33306730976](https://github.com/hew-lang/hew/actions/runs/33306730976), head `ec02c145e5651fbb4610f558b172a718698af4ce`.

### Failing checks, both runs

| Check | PR | main |
|---|---|---|
| `Linux gates (shard 3/3)` | fail | fail |
| `Build & test (Linux)` | fail (roll-up of the above) | fail (roll-up) |
| `Compiled Hew Test Results` | fail (26 failed) | fail (8 failed) |

### 1. `Linux gates (shard 3/3)` / `Build & test (Linux)` — not a new cause

`make ci-shard-3` stops at the first failing prerequisite, so the PR and main died at different targets:
- **main** dies earlier, at `fuzz-oracle` (`make: *** [Makefile:700: fuzz-oracle] Error 1`), never reaching `hew-check-all`.
- **PR** gets past `fuzz-oracle` and dies at `hew-check-all`, with two unexpected corpus failures:
  - `tests/pkg-import/pkgs/qmachbad/qmachbad.hew` — `E_PATH_MEMBER_NOT_FOUND: expected type 'Gate' has no variant 'Nope'`
  - `tests/pkg-import/qualified_machine_event_access.hew` — `module 'hew.qmach' not found`

Since CI on main never reaches `hew-check-all`, I verified locally rather than inferring: checked out `main` @ `585b9711273123a27a6fa19177510a16d0ad6cb7` (clean tree, no PR changes) and ran `HEW_BIN=target/release/hew scripts/corpus-ratchet.sh hew-corpus` directly. Result: **byte-identical** failure — the same two fixtures, same error text, both absent from `scripts/hew-corpus-expected-failures.txt` on main's own head. This is main's own regression (main's head added these fixtures without listing them), not the PR's. `Build & test (Linux)` is a derivative roll-up (`RUST_GATES_RESULT: failure`) with no independent content.

**Verdict for this job: not a new cause of failure.**

### 2. `Compiled Hew Test Results` — introduces 13 new failing IDs

Per-job JUnit annotation counts: PR 26 failed / main 8 failed. The failing test *names* do not overlap at all:

- **main's 8** (4 distinct × 2 shard duplicates) are all `imported_machine_ownership_test.hew` — `imported_payload_machine_covers_construction_calls_returns_and_record_storage`, `imported_machine_transitions_and_supports_dotted_unit_construction`, `module_qualified_variant_pattern_selects_the_matching_arm_for_both_states`, `imported_private_machine_crosses_an_owned_call_boundary` — all "unused function" warnings from other tests in the same file, unrelated to the PR.
- **PR's 26** (13 distinct × O0/O2) are all in `tests/hew/iter_test.hew`, `kind=runtime`, none present in main's set, none listed in `scripts/hew-suite-expected-failures.txt`:

```
tests/hew/iter_test.hew::test_vec_iter_any_method_short_circuits_to_true
tests/hew/iter_test.hew::test_vec_iter_find_method_returns_the_exact_item
tests/hew/iter_test.hew::test_vec_iter_count_method_counts_every_item
tests/hew/iter_test.hew::test_vec_iter_skip_method_keeps_exact_suffix
tests/hew/iter_test.hew::test_vec_iter_fold_method_accumulates_exact_value
tests/hew/iter_test.hew::test_vec_iter_take_method_clamps_to_exact_prefix
tests/hew/iter_test.hew::test_filter_map_collect_two_deep_chain_preserves_values
tests/hew/iter_test.hew::test_terminal_into_iter_method_chain_executes
tests/hew/iter_test.hew::test_vec_iter_map_method_delegates_to_iter
tests/hew/iter_test.hew::test_vec_iter_filter_method_keeps_exact_matching_values
tests/hew/iter_test.hew::test_vec_iter_all_method_detects_a_non_matching_value
tests/hew/iter_test.hew::test_vec_iter_enumerate_method_keeps_indices_and_values
tests/hew/iter_test.hew::test_vec_iter_collect_method_preserves_every_value
```

**Local reproduction** (PR worktree, head `ec02c145e`, via `hew build` — the same synthetic-`main`-wrapper compile `hew test` performs internally):

```
cp tests/hew/iter_test.hew .tmp/iter_test_repro.hew
printf '\n\nfn main() {\n    test_vec_iter_any_method_short_circuits_to_true();\n}\n' >> .tmp/iter_test_repro.hew
build/bin/hew build .tmp/iter_test_repro.hew -o .tmp/iter_test_repro_bin
.tmp/iter_test_repro_bin
```
Result: **Segmentation fault (exit 139)**, on the PR binary.

Same repro against `main`'s binary (`target/release/hew` @ `585b9711273123a27a6fa19177510a16d0ad6cb7`): **exit 0**, prints correctly.

This is a genuine PR-introduced regression, not present on main.

**Pattern observed (fact, not a confirmed root cause):** every failing test in the list calls the terminal/adapter method directly on a temporary `.into_iter()` receiver (e.g. `values.into_iter().any(...)`). The one comparable test in the same file that binds the iterator to a `let` first, `test_let_bound_into_iter_method_chain_executes`, is **not** in the failing set, and a standalone repro of that let-bound shape ran correctly. This points at the temporary-receiver path without asserting the mechanism — worth the ownership lane's attention given the PR's stated change to have calls borrow non-resource arguments by default.

### Verdict

**Introduces new failures — do not merge as-is.** 13 unique-to-PR test IDs in `tests/hew/iter_test.hew`, reproduced locally as a SIGSEGV not present on main. The `hew-check-all` and `fuzz-oracle` reds are main's own, verified locally against main's own head.

slepp:
Final-head repair note — this supersedes the earlier “introduces new failures” verdict.

Head 94863fc77413f8bb2075f625662ece20d5d4f575 now contains three bounded follow-ups:

- 16477561a completes temporary VecIter owner mint/handoff. make_vec().iter() compiles, evaluates the producer once, prints MK once, and returns count 4. The same producer-side neutralization fixes the Windows named/temporary into_iter cleanup path without a codegen special case.
- 70420fb59 removes four stale imported-machine Hew-suite expected-failure rows; all four tests pass on this branch.
- 94863fc77 refreshes only the attributable vec_dyn_iter raw-MIR golden and its manifest hash for the two AggregateMemberConsume neutralizations.

Local evidence: make lint passes; focused MIR/provenance 21/21; host plus Windows codegen oracle passes; clean full Rust run has zero candidate-only failing IDs versus main; hew-check-all is 1,754 checked with zero ratchet drift; all four imported-machine tests pass; checked-MIR verify is 62 fixtures x 2 stages byte-identical after refresh. The full local Hew aggregate has only 19 loopback/network failures from this restricted environment and zero NOW-PASSES.

CI evidence on the product-code head before the snapshot-only commit: Linux shard 1 passed, all four compiled-Hew O0/O2 shards passed, Clippy/format passed, and CodeQL/artifact/license/playground passed. The replacement final-head run is now in progress; both CodeQL analyses already pass.

Issue #3105, opened 2026-08-29
Release imported machine records on normal and unwind exits


slepp:
## Problem

The imported-machine ownership suite cannot compile a record that stores an
imported payload machine and later projects that field.

```console
target/debug/hew test tests/hew/imported_machine_ownership_test.hew --format text
```

The file's
`imported_payload_machine_covers_construction_calls_returns_and_record_storage`
function leaves `Envelope` owner `envelope` live without a matching destructor
on every call unwind edge and on return. Checked MIR reports
`ObligationUnderReleased` for the record at line 47, and every test case in the
file fails on that one compile error.

## Expected

Storing the imported machine in `Envelope`, reading its scalar field, and
transferring or projecting the machine field produce one complete cleanup plan
on normal and unwind exits. The imported-machine tests then execute.

slepp:
Closing as fixed. The exact repro from this issue is now covered by `tests/hew/imported_machine_ownership_test.hew` (case `imported_payload_machine_covers_construction_calls_returns_and_record_storage`), which passes cleanly with no ObligationUnderReleased on the imported record. #3173 replaced the two-authority split that this class of bug came from.

slepp:
Closing as fixed. The exact repro is covered by `imported_payload_machine_covers_construction_calls_returns_and_record_storage` in `tests/hew/imported_machine_ownership_test.hew`, which passes today with no ObligationUnderReleased on the Envelope record. #3173 replaced the two-authority split this class of bug came from.

PR #3180, opened 2026-08-31
test: add Hewbers compiler regressions

slepp:
## Summary

- preserve four minimized dogfood compiler failures as executable desired-
  behavior tests with exact `compile` ratchet rows
- add two passing canaries for Vec record allocation and permissive unknown-field
  wire JSON decoding
- keep advisory/API findings out of the corpus when their contract is undecided

The retained failures cover an indirect owned-`Result` scrutinee, imported
supervisor-child layout publication, and the two generic loop-generation shapes
seen in journal path and JSON builder reassignment. The latter two are evidence
for #3120; they are not shape-specific fix requests.

The stale #3105 rows are not reintroduced, and the proposed
`Node.register(ChildRef)` test is excluded pending Q356's explicit role-lookup
contract.

## Validation

- focused JUnit run: 6 discovered tests
  - 4 expected compile failures with current `E_NOT_YET_IMPLEMENTED`, `E_MIR`,
    or `E_MIR_ICE` diagnostics
  - 2 passing executable canaries
- support worker checks but is not independently discovered as a test
- deterministic/local-only audit: no sleeps, process, filesystem, network, or
  platform dependency
- `git diff --check`
- two independent fixture reviews: no P0-P3 findings

The full local corpus exceeded a bounded five-minute run. Hosted CI's four
compiled-Hew shards and inventory/union ratchet are the exhaustive gate for this
PR.

### DIFF
--- a/scripts/hew-suite-expected-failures.txt
+++ b/scripts/hew-suite-expected-failures.txt
@@ -15,3 +15,14 @@
 # the ratchet: a known compiler refusal cannot silently become a hung,
 # unstartable, or crashing binary under the same test identity.
 #
+# Priority-1 upstream reductions from Goobers-Hew.  Each is intentionally an
+# executable desired-behaviour test: remove its row when the failure is fixed.
+# #3173 currently refuses indirect owned call scrutinees with E_NOT_YET_IMPLEMENTED.
+tests/hew/call_scrutinee_indirect_ownership_test.hew::indirect_owned_result_call_scrutinee_preserves_payload compile
+# #3173 cannot attach an imported actor protocol descriptor and then reports the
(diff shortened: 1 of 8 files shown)


slepp:
Fresh rc3-main validation narrows the supervisor canary rather than removing it.

- A basic package/module import using a qualified `workers.Worker` child now
  builds and runs, returning 42.
- This PR's exact file-import route still fails on the same fresh compiler:
  the unqualified imported actor lacks its protocol descriptor/layout during
  MIR lowering, the supervisor child is reported as unknown, and the imported
  handler is absent from the actor layout.

So this is not a claim that imported supervised actors fail generally. The
fixture correctly preserves the narrower file-import publication/layout seam.

PR #3182, opened 2026-08-31
fix(wire): implement Option-first field presence

### LINKED ISSUE #3179: Implement Option-first wire field presence across codecs
## Problem

Wire field presence currently has no single runtime authority:

- the parser/checker retain an `optional` marker;
- CBOR codegen does not consult it and always selects/emits every tag;
- text descriptors infer omissibility from `Option<T>` itself, so required
  `Option<T>` is treated as omissible;
- compatibility can therefore classify a schema change as safe while runtime
  readers reject it.

PR #3178 is the bounded fail-closed admission slice: bare `T optional` is
rejected. It does **not** implement presence semantics or clear this issue.

## Required v0.6 contract

| Field form | Encode | Absent decode | Explicit `null` decode |
| --- | --- | --- | --- |
| `T` | key always present | error | error |
| `Option<T>` | key present; `None` is `null` | error | `None` |
| `Option<T> optional` | `Some` emits key/value; `None` omits key | `None` | `None` |
| `T optional` | rejected before codegen | n/a | n/a |

Field presence must be explicit metadata, independent of the existing Option
value descriptor. `Option<Option<T>>` remains fail-closed unless separately
specified.

## Implementation scope

- carry required/optional field presence into generated CBOR and text
  descriptors;
- add a non-latching optional CBOR tag selector while preserving the required
  selector's missing-tag failure;
- initialize absent optional fields to `None`, and retain normal Option decoding
  for present explicit null;
- omit `None` only for optional Option fields in JSON/YAML/CBOR;
- make descriptor parsing fail closed when presence metadata is absent;
- align schema compatibility with the same presence authority (at minimum,
  optional to required is breaking; prefer rejecting either presence flip until
  a proven directional rule exists);
- update the language guide, spec, doctrine, CDDL, grammar copies, and std wire
  docs together with the implementation.

## Acceptance matrix

- required scalar: present succeeds; absent and null reject;
- required Option: Some and explicit null succeed; absent rejects;
- optional Option: Some succeeds; None is omitted; absent and explicit null both
  decode to None;
- exercise JSON, YAML, and direct CBOR, including round trips and unknown-tag
  tolerance;
- compatibility tests distinguish value nullability from field presence and do
  not call a runtime-incompatible change safe;
- hard failures return nonzero and publish no requested artifact.

This is an rc3/v0.6 conformance blocker (coordination decision D300).

slepp:
## Summary

Implements the ratified Option-first wire-field presence contract from #3179.
The full-presence commit is rebased directly onto `main` after the admission
slice from #3178.

Presence and value shape are separate semantic facts:

| Declaration | Encode | Missing input | `null` input |
| --- | --- | --- | --- |
| `field: T` | always emit value | error | error |
| `field: Option<T>` | always emit; `None` is `null` | error | `None` |
| `field: Option<T> optional` | omit `None`; emit `Some` | `None` | `None` |
| `field: T optional` | rejected by #3178 | n/a | n/a |

The same contract applies across generated CBOR and the JSON/YAML text bridge.
Unknown distinct tags remain forward-compatible, while duplicate CBOR struct
tags fail deterministically rather than collapsing last-wins.

## Implementation

- Carries explicit `WireFieldPresence` metadata independently of `Option<T>`.
- Adds required and optional CBOR key selectors with fail-closed missing-key
  handling.
- Emits required keys unconditionally and optional keys only for `Some`.
- Requires explicit presence metadata in text descriptors.
- Rejects ambiguous `Option<Option<T>>` on both encode and decode paths.
- Rejects required/optional presence flips in `hew wire check` compatibility.
- Updates the C ABI/JIT surface, CDDL, language guide, and vertical-slice
  fixtures.

## Validation

- Rebased doc-fence gate: 208 passed, 85 expected failures, 23 skipped.
- FFI validation: 568 contracts.
- Optional-presence CLI suite: 4 passed.
- Wire-compatibility CLI suite: 28 passed.
- Runtime wire-text suite: 25 passed.
- Formatting and diff checks passed.
- Independent semantic review: GO with no P0-P2 findings.

Non-blocking test hardening follows separately: fuller CDDL negative cases, an
actual sandbox optional-field fixture, and a committed compiled-body duplicate
CBOR-tag fixture.

Closes #3179.


slepp:
Independent semantic gate: GO at `4c7d90ee4dffee1f08cd2c187b668840f1a9c53e`.

No P0-P2 findings. The ratified required / nullable / omissible matrix agrees across CBOR, JSON, YAML, compatibility, descriptors, generated ownership, and failure publication. Focused native O0/O2, required-key/null traps, duplicate-tag counterfactual, ABI/FFI, WASI, sandbox-admission, and doc-ratchet checks passed.

Non-blocking P3 follow-up after the rc3 conformance landing:

- extend committed CDDL cases to present optional null and negative missing-required/null-bare shapes;
- give sandbox parity an actual optional-field fixture rather than a required-only fixture;
- commit a compiled-body duplicate-CBOR-tag rejection fixture.

The branch is safe to rebase onto `main` and mark ready after #3178 lands.

PR #3210, opened 2026-09-02
refactor(types): mint declaration identity once in the checker

slepp:
## Why

Declaration identity was reconstructed from rendered paths at every consumer,
so a declaration reached through a root compile, an import, an alias or a
module graph could end up with different `DefId`s, and every downstream join
was name-keyed. The checker now mints a declaration's identity once and
publishes it; lowering carries that identity instead of re-deriving it.

## What

- **types**: `IdentityTable` gains a source-declaration axis. Every item
  declares a `DeclarationOccurrence` (module, item span, kind, child ordinal)
  against a canonical path and receives one `DefId`. The table is frozen into
  a lookup-only `IdentityView` on `TypeCheckOutput`, and checker consumers
  resolve through it rather than minting. The legacy root render at the
  `fn_sigs` publication boundary is deleted, so root free functions publish
  under their canonical `{root}.{name}` key with no bare alias beside it.
- **types**: `NominalNamespace` replaces the bare-vs-qualified boolean and
  names the three cases. A nominal's PRIMARY render is the spelling
  `ResolvedTy::Named` carries for it, because every declaration-derived
  downstream key (`HirTypeDecl::qualified_name`, the MIR layout tables)
  renders from it: bare for the root unit, owner-qualified for a module, and
  owner-qualified for a file import flattened into the root. The other
  reachable spelling — the root's own module path, or the flattened file's
  bare name — is recorded as a second path on the SAME occurrence.
  **This departs from the frozen slice's `declare` contract**, which refused a
  second canonical path for an established occurrence. Probing a package whose
  peer is also importable in its own right shows `fn_sigs` publishing BOTH
  `pkg.make` and `pkg.aaa.make` for one physical declaration, so the strict
  rule refused a spelling the checker itself hands out. Two DIFFERENT
  occurrences claiming one path — the direction that would equate two
  declarations — are still refused, with the one exception below.
- **types**: peer files of a directory module render their declarations under
  the assembling module, which is the spelling the checker publishes
  (`pkg.make`, not `pkg.helpers.make`). The occurrence axis stays per file so
  two peers whose items share a span remain distinguishable.
- **types**: an `extern "C"` symbol is the one declaration form where two
  occurrences under one path are genuinely one declaration — the linker binds
  every call to a single implementation, the checker keys every declaration by
  the same fn-sig path, and the extern table resolves the redeclaration
  against the established ABI contract. Peer files of a shipped directory
  module routinely re-declare a runtime symbol (`std/net/http` declares
  `hew_http_last_error` twice), so the further occurrence binds to the
  established declaration instead of being refused.
- **types**: every other collision is a redefinition, and the one registration
  cannot see is a collision ACROSS files that share one namespace: it visits
  each file alone, and only the shared path collides. That case now reports a
  duplicate definition naming both files, replacing the silent refusal that
  left the user with the identity table's internal "no declaration" wording
  from whichever consumer asked first.
- **types**: the `unsafe` gate is restored for the two extern surfaces that
  have no ABI contract. Registry-imported externs and the codegen-intercepted
  `hew_channel_*_layout` witnesses declare into the identity table and register
  in the one extern table, replacing the removed name-keyed side set. They are
  two independent inventories declaring into ONE module, so a single occurrence
  allocator per module keeps them apart; each counting from its own zero made
  their occurrences collide, and a colliding occurrence resolves to the
  established declaration, so the witness rows adopted the mirror rows'
  endpoints and `Sender::send` emitted a call to `hew_channel_pair_sender`.
  A mirror row also does not own a call endpoint (`owns_endpoint`): it gates
  `unsafe` and carries provenance, but the call it names is executed by its
  runtime family, so call-target resolution reads endpoint-owning rows only.
- **hir**: `LowerCtx` resolves each item's exact occurrence through the
  identity view and fails closed with `CheckerBoundaryViolation` on a miss.
  New required `declaration` fields on the const, actor, lifecycle-hook,
  machine, record and supervisor HIR nodes carry the checker's identity. A
  trait default the impl does not override has no source `fn` for the checker
  to inventory, so its identity is minted at the synthesis boundary and is now
  carried into body lowering rather than looked for in a table it was never in.
- **cli**: `hew machine list` and `hew machine diagram` run the checker
  instead of lowering against a default `TypeCheckOutput`, which would now
  drop every item. Their missing-import diagnostic is consequently the
  checker's (`undefined variable \`fs\``) rather than the HIR static check's,
  and the machine fixtures move to the current transition-target dialect
  (`.On` rather than `On`).
- **analysis**: `signature_help` resolves a bare root spelling through the
  declaration table in place of the deleted `IdentityTable::root_fn_identity`.
- **mir**: record-layout publication moves into its own module and publishes a
  shape under both of a declaration's spellings. They coincide for every
  ordinary module; a peer file that is importable in its own right is the case
  where a use site can name either one. The MIR-shape harnesses run the
  checker, because HIR lowering now needs its identities; what they assert
  about MIR shape is unchanged.
- **types/hir**: three joins that matched one render now resolve the spelling
  through the identity table first, which is the one name-resolution
  authority: the extern row a call target reads, `trait_method_ids` for a
  trait reference, and the impl receiver the resource-lifecycle admission
  compares. A span index is allocated per (module, file), so HIR maps every
  index a peer file is walked under rather than only the first.

Left for the next slice, unchanged here: the 18
`NominalId::legacy_reconstruct_from_full_path` sites in `hew-hir/src/dispatch.rs`
and the two free mints in `hew-types/src/resolved_ty.rs`
(`default_impl_method_declaration`, `nominal_instance`).

## Test

- `hew-types/src/identity.rs`: a second canonical path for one occurrence
  resolves to the established identity under both spellings and does not mint
  a second; two distinct occurrences claiming one path still fail closed,
  including when the second occurrence already owns another spelling.
- `hew-types/src/check/registration.rs`:
  `a_registry_mirror_row_does_not_adopt_a_layout_witness_declaration` declares
  both source-less extern inventories into one module and asserts each row
  keeps its own identity; `channel_layout_witnesses_gate_unsafe` keeps the
  `unsafe` proof with a plain `fn_sigs` entry as its negative control.
- `hew-types/src/check/calls.rs`:
  `a_contractless_witness_row_does_not_displace_the_runtime_family` — a mirror
  row gates `unsafe` yet the witness call still resolves to
  `Runtime(ChannelRecvLayout)`, with an endpoint-owning declaration of the same
  key as the counterfactual that DOES publish an extern target.
- `hew-types/src/check/tests/modules.rs`:
  `registry_imported_extern_call_requires_unsafe` — a bare registry extern
  call demands `unsafe`, with the same call inside `unsafe` as the positive
  control.
- `hew-types/src/check/tests/extern_fn.rs`:
  `divergent_same_named_peer_nominals_are_refused_as_a_redefinition` replaces
  the third-render-collision oracle, whose premise (per-file namespaces for
  peers) the checker does not publish, paired with a new positive control that
  distinctly named peers assemble cleanly.
- `hew-compile/src/lib.rs`:
  `flattened_file_import_declares_one_identity_per_item` puts one item on both
  the graph-module and root `program.items` surfaces and asserts one `DefId`;
  `directory_module_peers_claiming_one_name_report_a_duplicate_definition` and
  `a_root_and_its_file_import_claiming_one_name_report_one_duplicate` pin the
  cross-file report and that it fires once, with
  `directory_module_peers_may_redeclare_one_extern_symbol` as the negative
  control that a shared C symbol still type-checks. The direct-symbol matrix
  pins the root declaration under the root's own identity, since only the
  emitted body symbol stays bare.
- `hew-hir/tests/declaration_identity.rs`:
  `directory_peer_declarations_publish_under_the_assembling_module` pins the
  assembled spelling and that HIR carries the checker's identity rather than a
  re-render.

`make lint`, `make test-vertical-slice` and `make fuzz-oracle` are green.
After `make wasm-runtime-debug` and
`make libhew-cross-release-lib` (the cross-arch archive `make test-artifacts`
provisions; without it `native_link_linux_cross_arch_produces_foreign_elf_binary`
refuses rather than skips, because its sysroot IS present here),
`cargo nextest run -p hew-types -p hew-hir -p hew-mir -p hew-analysis -p hew-cli -p hew-compile --no-fail-fast`
leaves no failure outside `scripts/nextest-expected-failures.tsv`.


slepp:
The 37 failures this slice still owes, from
\`cargo nextest run -p hew-types -p hew-hir -p hew-mir -p hew-analysis -p hew-cli -p hew-compile --no-fail-fast\`
(none of them appear in `scripts/nextest-expected-failures.tsv`):

- `hew-cli::actor_resource_opaque_state file_imported_user_receiver_closes_once_without_runtime_close`
- `hew-cli::affine_resource_carrier_boundaries channel_handle_clone_terminals_match_runtime_semantics`
- `hew-cli::affine_resource_carrier_boundaries receiver_vec_move_is_descriptor_owned_and_read_copy_surfaces_reject`
- `hew-cli::channel_recv_e2e await_recv_actor_binds_some_then_none_under_both_pools`
- `hew-cli::channel_recv_e2e select_channel_recv_arm_picks_ready_under_both_pools`
- `hew-cli::channel_send_shutdown_e2e full_send_at_shutdown_fails_closed_with_named_channel`
- `hew-cli::channel_send_shutdown_e2e full_send_with_draining_receiver_still_backpressures`
- `hew-cli::codegen_failclosed_span_e2e file_imported_trait_default_method_unsupported_type_no_false_root_caret`
- `hew-cli::consuming_call_owner_transfer_e2e consumed_pair_carriers_run_to_clean_exit_with_sentinels`
- `hew-cli::eval_e2e channel_record_elements_roundtrip_and_early_exit_under_single_worker`
- `hew-cli::eval_e2e for_await_receiver_int_drains_to_completion_under_single_worker`
- `hew-cli::eval_e2e for_await_receiver_string_drains_to_completion_under_single_worker`
- `hew-cli::eval_e2e for_await_recv_f64_drains_to_completion_under_single_worker`
- `hew-cli::extern_ownership_double_release_e2e a_returned_channel_pair_is_not_closed_by_its_producer`
- `hew-cli::fmt_stdin_e2e fmt_migrate_root_typechecks_legacy_directory_module_peers`
- `hew-cli::machine_e2e machine_diagram_json_event_fields_present`
- `hew-cli::machine_e2e machine_diagram_json_no_wildcard_rows`
- `hew-cli::machine_transition_watch_e2e awaited_ask_select_machine_heap_payload_stays_clean_under_scribble`
- `hew-cli::machine_transition_watch_e2e channel_receiver_actor_message_arg_transfers_locally`
- `hew-cli::machine_transition_watch_e2e channel_sender_actor_message_arg_transfers_locally`
- `hew-cli::machine_transition_watch_e2e cross_actor_record_transition_watch_runs_clean`
- `hew-cli::machine_transition_watch_e2e machine_snapshot_select_watch_matches_state_variants`
- `hew-cli::machine_transition_watch_e2e nested_channel_handle_in_tuple_transfers_correctly`
- `hew-cli::machine_transition_watch_e2e select_after_genuine_expiry_takes_after_arm`
- `hew-cli::machine_transition_watch_e2e select_enum_element_thunks_resolve_and_run_clean`
- `hew-cli::machine_transition_watch_e2e select_on_transition_example_runs_clean`
- `hew-cli::machine_transition_watch_e2e select_record_element_cross_block_arm_runs_clean`
- `hew-cli::machine_transition_watch_e2e suspending_select_wake_gate_ir_shape_holds`
- `hew-cli::machine_transition_watch_e2e transition_watch_baseline_prints_transition`
- `hew-cli::opaque_resource_close_counterfactual opaque_close_with_extra_side_effect_is_not_a_canonical_release`
- `hew-cli::run_e2e check_dual_module_same_type_name_impl_resource_qualified_compiles`
- `hew-cli::run_e2e run_generic_display_cross_module_receiver`
- `hew-cli::run_e2e run_string_path_imported_record_annotated_param_field_access`
- `hew-compile tests::checking_directory_module_peer_loads_entry_namespace`
- `hew-compile tests::flat_imported_specialisation_does_not_claim_the_generic_dispatch_key`
- `hew-compile tests::mixed_file_and_package_impls_keep_declaration_owned_dispatch_in_both_import_orders`
- `hew-compile tests::nested_generic_free_calls_keep_exact_direct_symbols_across_all_module_origins`

Shapes, from the panic messages:

- channel and for-await fixtures fail to compile with `hew_channel_pair_sender has no declared parameter for Hew argument 1`;
- the machine transition-watch and diagram fixtures fail to build;
- `checking_directory_module_peer_loads_entry_namespace` refuses because a directly checked peer of a directory module makes that module the graph root, whose nominals the checker qualifies while minting still renders them bare — qualifying them there re-renders every ordinary root compile, so the root unit keeps its bare namespace and this stays open;
- the remaining three hew-compile cases are cross-module dispatch keys for flat-imported and package-imported impls.

One diagnostic-quality gap, marked in `check/mod.rs`: two peer files of one directory module claiming one canonical path are refused, but silently at the identity table, so the user sees whichever consumer asks first rather than a duplicate-definition message.

slepp:
Corrected list, replacing my earlier comment. After `make wasm-runtime-debug`,
`cargo nextest run -p hew-types -p hew-hir -p hew-mir -p hew-analysis -p hew-cli -p hew-compile --no-fail-fast`
leaves 62 failures; 22 are in `scripts/nextest-expected-failures.tsv` and these 40 are not:

- `hew-cli::actor_resource_opaque_state file_imported_user_receiver_closes_once_without_runtime_close`
- `hew-cli::affine_resource_carrier_boundaries channel_handle_clone_terminals_match_runtime_semantics`
- `hew-cli::affine_resource_carrier_boundaries receiver_vec_move_is_descriptor_owned_and_read_copy_surfaces_reject`
- `hew-cli::channel_recv_e2e await_recv_actor_binds_some_then_none_under_both_pools`
- `hew-cli::channel_recv_e2e select_channel_recv_arm_picks_ready_under_both_pools`
- `hew-cli::channel_send_shutdown_e2e full_send_at_shutdown_fails_closed_with_named_channel`
- `hew-cli::channel_send_shutdown_e2e full_send_with_draining_receiver_still_backpressures`
- `hew-cli::codegen_failclosed_span_e2e file_imported_trait_default_method_unsupported_type_no_false_root_caret`
- `hew-cli::consuming_call_owner_transfer_e2e consumed_pair_carriers_run_to_clean_exit_with_sentinels`
- `hew-cli::cross_target_e2e native_link_linux_cross_arch_produces_foreign_elf_binary`
- `hew-cli::eval_e2e channel_record_elements_roundtrip_and_early_exit_under_single_worker`
- `hew-cli::eval_e2e for_await_receiver_int_drains_to_completion_under_single_worker`
- `hew-cli::eval_e2e for_await_receiver_string_drains_to_completion_under_single_worker`
- `hew-cli::eval_e2e for_await_recv_f64_drains_to_completion_under_single_worker`
- `hew-cli::extern_ownership_double_release_e2e a_returned_channel_pair_is_not_closed_by_its_producer`
- `hew-cli::fmt_stdin_e2e fmt_migrate_root_typechecks_legacy_directory_module_peers`
- `hew-cli::machine_e2e machine_diagram_json_event_fields_present`
- `hew-cli::machine_e2e machine_diagram_json_no_wildcard_rows`
- `hew-cli::machine_transition_watch_e2e awaited_ask_select_machine_heap_payload_stays_clean_under_scribble`
- `hew-cli::machine_transition_watch_e2e channel_receiver_actor_message_arg_transfers_locally`
- `hew-cli::machine_transition_watch_e2e channel_sender_actor_message_arg_transfers_locally`
- `hew-cli::machine_transition_watch_e2e cross_actor_record_transition_watch_runs_clean`
- `hew-cli::machine_transition_watch_e2e machine_snapshot_select_watch_matches_state_variants`
- `hew-cli::machine_transition_watch_e2e nested_channel_handle_in_tuple_transfers_correctly`
- `hew-cli::machine_transition_watch_e2e select_after_genuine_expiry_takes_after_arm`
- `hew-cli::machine_transition_watch_e2e select_enum_element_thunks_resolve_and_run_clean`
- `hew-cli::machine_transition_watch_e2e select_on_transition_example_runs_clean`
- `hew-cli::machine_transition_watch_e2e select_record_element_cross_block_arm_runs_clean`
- `hew-cli::machine_transition_watch_e2e suspending_select_wake_gate_ir_shape_holds`
- `hew-cli::machine_transition_watch_e2e transition_watch_baseline_prints_transition`
- `hew-cli::opaque_resource_close_counterfactual opaque_close_with_extra_side_effect_is_not_a_canonical_release`
- `hew-cli::run_e2e check_dual_module_same_type_name_impl_resource_qualified_compiles`
- `hew-cli::run_e2e run_generic_display_cross_module_receiver`
- `hew-cli::run_e2e run_string_path_imported_record_annotated_param_field_access`
- `hew-cli::wasi_run_e2e native_actor_periodic_timer_reaches_quiescence`
- `hew-cli::wasi_run_e2e wasm_channel_send_full_traps_instead_of_dropping`
- `hew-compile tests::checking_directory_module_peer_loads_entry_namespace`
- `hew-compile tests::flat_imported_specialisation_does_not_claim_the_generic_dispatch_key`
- `hew-compile tests::mixed_file_and_package_impls_keep_declaration_owned_dispatch_in_both_import_orders`
- `hew-compile tests::nested_generic_free_calls_keep_exact_direct_symbols_across_all_module_origins`

Shapes, from the panic messages:

- the channel and for-await fixtures fail to compile with `hew_channel_pair_sender has no declared parameter for Hew argument 1`; the machine transition-watch and diagram fixtures fail to build;
- `checking_directory_module_peer_loads_entry_namespace` refuses because the checker qualifies a named graph root's nominals while minting renders them bare. Keying that off a dotted root path re-rendered every ordinary root compile and took hew-cli from 87 failures to 333, so the discriminating condition is not `graph.root` and is still open; the shortcut is marked at the rule in `check/mod.rs`;
- the other three hew-compile cases are cross-module dispatch keys for flat-imported and package-imported impls;
- `native_link_linux_cross_arch_produces_foreign_elf_binary` wants a cross-arch artifact this checkout does not have, so it may be provisioning rather than owed.

One diagnostic-quality gap, marked in `check/mod.rs`: two peer files of one directory module claiming one canonical path are refused, but silently at the identity table, so the user sees whichever consumer asks first rather than a duplicate-definition message.

PR #3214, opened 2026-09-02
build(installers): install the newest release candidate by default

slepp:
## Why

Until v0.6.0 ships, `curl | sh` installs whatever GitHub's `/releases/latest`
endpoint returns — the newest *final* release only, so the installer never
picks up a release candidate even when the rc is the newest thing published.

## What

- **install.sh**: default version resolution now walks the GitHub `/releases`
  list (rc tags included) and picks the newest tag by semver, instead of
  hitting the non-prerelease-only `/releases/latest` endpoint. Adds `--stable`
  (newest final release only) and `--dry-run` (print the resolved tag, archive
  URL, and prefix without downloading or installing).
- **installers/README.md**: documents `--stable`, `--dry-run`, and the two
  Homebrew channels.
- **release.yml**: the Homebrew tap trigger now fires for every tag, not just
  final releases — `Formula/hew.rb` in homebrew-hew tracks the newest release
  (rc included), `Formula/hew@stable.rb` tracks final releases only.

Companion changes in `hew-lang/homebrew-hew` (pushed to main): add
`Formula/hew@stable.rb` and repoint `Formula/hew.rb` at v0.6.0-rc2, then
update `update-formula.yml` to drop its rc skip and update both formulae —
`hew.rb` for every tag, `hew@stable.rb` only when the version has no `-`.

## Test

- `sh -n installers/install.sh`.
- `make test-install-version-resolution`: a committed offline fixture
  (`installers/test-pick-newest-tag.sh`) for the semver picker
  (`pick_newest_tag`) and the `--stable` filter, wired into the `make lint`
  gate CI already runs. Cases: 0.10.0 beats 0.9.9 (numeric, not lexical);
  0.6.0 beats 0.6.0-rc3 (the final-release sentinel); rc10 beats rc9
  (numeric coercion on the rc suffix); 0.6.0-rc2 beats 0.5.6 on the default
  path; `--stable` returns 0.5.6 over the same input as a negative control.
- Live against the GitHub API: `--dry-run`, `--dry-run --stable`,
  `--dry-run --version 0.5.6`, an unknown flag, and `--version` + `--stable`
  together (rejected as mutually exclusive).
- Full install run (`--stable --prefix <scratch>`) followed by `hew version`
  against the installed binary.
- Downloaded the real v0.6.0-rc2 linux-x86_64 tarball, verified its checksum
  and layout against the `hew@stable`/`hew` Homebrew formula's `install`
  block, and ran the formula's `test do` program (`import std.math;
  math.clamp`) against the extracted build.

by CI.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -24,8 +24,9 @@
 #     identity, e.g. `v0.6.0-rc2@sha256:<64 lowercase hex digits>`. The
 #     maintainer records it only after the pre-tag candidate publish succeeds.
 #
-# Homebrew tap updates use a separate optional credential and are skipped for
-# prereleases:
+# Homebrew tap updates use a separate optional credential and run for every
+# tag, prereleases included — hew.rb tracks the newest release (rc included),
+# hew@stable.rb tracks final releases only:
 #   HOMEBREW_TAP_TOKEN           — GitHub PAT with actions:write scope for hew-lang/homebrew-hew
 #
 # Required secrets for VS Code extension publishing (optional — skipped if absent):
@@ -1502,8 +1503,10 @@ jobs:
   homebrew:
     needs: release
     runs-on: ubuntu-24.04
-    # Formula updates are for final releases; -rc tags must not touch the tap.
-    if: ${{ !cancelled() && needs.release.result == 'success' && !contains(github.event.inputs.tag || github.ref_name, '-rc') }}
+    # Triggers every tag: Formula/hew.rb tracks the newest release including
+    # release candidates, Formula/hew@stable.rb tracks final releases only —
+    # update-formula.yml in homebrew-hew decides which formula(e) to touch.
+    if: ${{ !cancelled() && needs.release.result == 'success' }}
     timeout-minutes: 5
 
     steps:
--- a/Makefile
+++ b/Makefile
@@ -74,7 +74,7 @@
 #   make clean        — remove generated build and test artifacts
 # ============================================================================
 
-.PHONY: all build bootstrap install-hooks help shell-script-lint actionlint hew hew-debug hew-profile-check hew-native shared-host-debug hew-lsp observe observe-functional-test mqtt-broker-e2e libhew-link-race-test runtime stdlib wasm-runtime wasm wasm-capability wasm-capability-check playground-manifest playground-manifest-check sandbox-fixtures sandbox-fixtures-check sandbox-vm-deps sandbox-parity playground-check playground-wasi-check preflight ci-preflight ci-preflight-smoke ci-local-linux wasm-dist release licenses licenses-check baselines baselines-check
+.PHONY: all build bootstrap install-hooks help shell-script-lint test-install-version-resolution actionlint hew hew-debug hew-profile-check hew-native shared-host-debug hew-lsp observe observe-functional-test mqtt-broker-e2e libhew-link-race-test runtime stdlib wasm-runtime wasm wasm-capability wasm-capability-check playground-manifest playground-manifest-check sandbox-fixtures sandbox-fixtures-check sandbox-vm-deps sandbox-parity playground-check playground-wasi-check preflight ci-preflight ci-preflight-smoke ci-local-linux wasm-dist release licenses licenses-check baselines baselines-check
 .PHONY: test test-strict ratchet-accounting ratchet-accounting-nextest test-ratchet-accounting-runner macos-leak-oracle test-leak-oracle-selftest test-cabi test-compiler-pipeline test-compiler-lifecycle test-opaque-resource-lifecycle-matrix test-opaque-resource-lifecycle-matrix-external test-vertical-slice test-pkg-import test-package-install test-runtime-unit test-hew-ratchet test-core-matrix core-matrix-record funcupdate-mir-baselines-golden test-o2-differential o2-differential-selftest test-stdlib-ratchet test-ux-examples ux-examples-expect test-surface-examples surface-examples-expect test-example-expectations-selftest test-release-binary test-release-lib-link asan asan-fixtures test-asan-fixture-selftest tsan miri lint structural-lint structural-lint-bootstrap structural-lint-bootstrap-install test-ast-grep-contract stdlib-lint stdlib-errno-gate legacy-path-syntax-lint hew-fmt-check test-migrate-corpus doc-ratchet-selftest verify-sys-lane-closure test-sys-lane-closure hew-fmt-property test-build-harness forced-cancel-composite-check
 .PHONY: test-ownership-balance-corpus test-ownership-balance-runner-selftest
 .PHONY: stdlib-user-build-clean
@@ -94,6 +94,15 @@ help:
 LINT_GATES += shell-script-lint
 shell-script-lint:
 	@$(PYTHON) scripts/shell-script-lint.py
+
+# The installer's newest-tag semver picker (installers/install.sh's
+# pick_newest_tag()) has no coverage anywhere else in the build: it never
+# runs from a compiled binary, only from a shell function that curl | sh
+# executes directly. Wired into LINT_GATES so it runs on every PR the same
+# way shell-script-lint does, without a release download or network access.
+LINT_GATES += test-install-version-resolution
+test-install-version-resolution:
+	@sh installers/test-pick-newest-tag.sh
 
 
 # Local GitHub Actions syntax, expression, local-action input, and embedded-shell
--- a/installers/README.md
+++ b/installers/README.md
@@ -8,6 +8,18 @@ Installation options for [Hew](https://hew.sh).
 curl -fsSL https://hew.sh/install | sh
 ```
 
+Until v0.6.0 ships as a final release, this installs the newest published
+release _including release candidates_. Pass `--stable` to install the
+newest final release instead:
+
+```sh
+curl -fsSL https://hew.sh/install | sh -s -- --stable
+```
+
+Other `install.sh` flags: `--version <ver>` (pin an exact version),
+`--prefix <dir>` (installation directory), `--dry-run` (print what would be
+installed without installing it). `sh install.sh --help` lists all of them.
+
 ## Quick Install (FreeBSD)
 
 FreeBSD base includes `fetch(1)` — no extra packages needed:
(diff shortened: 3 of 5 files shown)


slepp:
The `install.sh` side of this is right. I ran the resolver at the head commit
against the live API: default resolves `v0.6.0-rc2`, `--stable` resolves
`v0.5.6`, `--version 0.5.6` pins, and `--version` with `--stable` is rejected.
I also ran `pick_newest_tag` under gawk, mawk, and busybox awk with identical
results, and the `grep '^v[0-9]'` filter correctly drops the
`toolchain/llvm-*` tags that share the releases listing. Two things block the
merge.

## 1. The cross-repo contract the PR asserts is not true

The release.yml diff says:

```
    # Triggers every tag: Formula/hew.rb tracks the newest release including
    # release candidates, Formula/hew@stable.rb tracks final releases only —
    # update-formula.yml in homebrew-hew decides which formula(e) to touch.
```

and installers/README.md now promises:

```
| macOS / Linux (Homebrew) | `brew install hew-lang/tap/hew` (newest release, rc included) or `hew@stable` (final releases only) |
```

`update-formula.yml` on `hew-lang/homebrew-hew` `main` does neither. It still
carries the rc skip:

```yaml
    # RC tags are intentionally excluded: the tap only tracks final releases.
    if: ${{ !contains(inputs.version, '-rc') }}
```

and its update step writes one formula:

```yaml
          sed -i "s/version \".*\"/version \"${VERSION}\"/" Formula/hew.rb
```

`hew@stable.rb` is never named anywhere in that workflow. The companion commit
`c1438d3c` ("build(formula): add hew@stable, track newest rc in hew") touched
`Formula/hew.rb` and `Formula/hew@stable.rb` only, and there is no open PR on
that repo carrying the workflow change. So with this merged:

- An rc3 tag dispatches `update-formula.yml`, the `if:` skips it, and
  `Formula/hew.rb` stays pinned at the hand-written `0.6.0-rc2` forever. The
  README row promises something the automation does not do.
- A final v0.6.0 tag rewrites `hew.rb` only, leaving `hew@stable.rb` orphaned
  at `0.5.6` with nothing that ever updates it - the opposite of the
  "final releases only" channel the README documents.

Fix either by landing the homebrew-hew workflow change first (drop the rc
skip, always update `hew.rb`, update `hew@stable.rb` only when `version`
contains no `-`), or by dropping the release.yml `if:` change and the
`hew@stable` README claim from this PR until that workflow exists. The PR
body's "updates `update-formula.yml` to maintain both formulae going forward"
also needs to become true or come out - it squashes into main's history.

## 2. The semver picker ships with no committed test

```sh
pick_newest_tag() {
```

This awk block is the whole point of the change and nothing in the repo would
catch a regression in it. The Test section describes an offline fixture that
was run but not committed. `installers/docker/test-install.sh` is precedent
for a shell test living in that directory; wire a fixture in wherever CI or a
`make` target will actually run it.

The cases that discriminate a plausible bug, rather than just re-asserting the
happy path:

- `0.10.0` beats `0.9.9` - catches a lexical comparison, the reason the
  `%06d` padding exists.
- `0.6.0` beats `0.6.0-rc3` - catches the `rc = 999999` sentinel for finals
  being dropped or inverted.
- `0.6.0-rc10` beats `0.6.0-rc9` - catches `substr($4, 3)` losing the `+ 0`.
- `0.6.0-rc2` beats `0.5.6` - catches an over-eager stable filter on the
  default path.
- `--stable` over the same input returns `0.5.6`, not the rc - the negative
  control for the `grep -v -- '-'` filter.

## Not blocking

- `install.ps1` still reads `/releases/latest`, so Windows diverges from the
  new Unix default. Disclosed in the PR body and the README's rc sentence sits
  under the curl section, so it reads correctly, but a line in the PowerShell
  section saying it tracks final releases only would save a support question.
- The SHORTCUT comment says a `-beta1` tag "would sort as an unrelated
  string". It does not - `substr("beta1", 3) + 0` is `0`, so it lands in the
  rc0 slot, below every rcN of the same core version and above any older
  final. That happens to match SemVer precedence, so the behaviour is fine and
  only the comment is wrong. I confirmed no `-dev` or `-beta` tag has ever
  been published, so the shortcut's WHY holds.
- No permissions widening. The `homebrew` job has no `permissions:` block and
  inherits the workflow's `contents: read`; the diff touches only its `if:`.

slepp:
Both blockers from the last review are applied and I verified each one.

**Cross-repo contract (was blocker 1) — resolved.** `update-formula.yml` on
`hew-lang/homebrew-hew` `main` no longer carries the rc skip, and its update
step handles both formulae:

```yaml
          # Formula/hew.rb tracks the newest release, rc included.
          update_formula Formula/hew.rb
          ...
          case "${VERSION}" in
            *-*) echo "Skipping Formula/hew@stable.rb: ${VERSION} is a pre-release" ;;
            *)
              update_formula "Formula/hew@stable.rb"
```

`Formula/hew@stable.rb` exists on that repo at `version "0.5.6"`. The
release.yml `if:` change and the README's `hew@stable` row are now true.

**Committed test (was blocker 2) — resolved.** `installers/test-pick-newest-tag.sh`
is committed and reachable: `lint: $$(LINT_GATES)` at Makefile:1547 with
`LINT_GATES += test-install-version-resolution`, and ci.yml:83 runs `make lint`.
I ran the fixture at this head — five cases pass. I also mutated
`install.sh` and confirmed the fixture catches both mutations that matter:
dropping the `%06d` padding fails two cases, and flipping the `rc = 999999`
sentinel to `0` fails one. `python3 scripts/shell-script-lint.py` also passes
at this head, so the other LINT_GATES member accepts the new script.

Three new things block the merge.

## 1. The branch conflicts with main

`mergeStateStatus` is `DIRTY`. `git merge-tree origin/main b02b7c04` reports
one conflict, in `installers/README.md`: main has since landed the PowerShell
side of this work, and its Windows section already says what this PR's diff
does not know about:

```
Until v0.6.0 ships as a final release, this installs the newest published
release _including release candidates_ — matching `install.sh`'s policy
(#3214). Pass `-Stable` to install the newest final release instead:
```

against this PR's rewrite of the same file — the new `install.sh` section and
the reflowed Package Managers table:

```
| macOS / Linux (Homebrew) | `brew install hew-lang/tap/hew` (newest release, rc included) or `hew@stable` (final releases only) |
```

Rebase onto `origin/main` and resolve. Keep both sections; the table row is
the one that needs a real merge, since main reflowed the same columns.

## 2. CI has never run on this head

`gh run list --commit b02b7c04` returns only CodeQL. The full CI run
(`33601243629`, success) is on `03b65579` — the first commit, before
`test-install-version-resolution` was added to `LINT_GATES`. CodeQL builds
`refs/pull/3214/head`; CI builds the merge ref, which GitHub cannot produce
while the PR conflicts. So this is absent, not pending: the gate the last
review asked for has never executed in CI. The rebase in (1) fixes this by
producing a mergeable head for CI to build.

## 3. The PR body's Deferred paragraph is no longer true

```
Deferred (tracked in `.tmp/TODO.md`, not touched here): `install.ps1` still
reads `/releases/latest` unconditionally
```

`git show origin/main:installers/install.ps1` has `-Stable` (line 62),
`-DryRun` (line 64), the mutual-exclusion check (line 68), and the same
rc-inclusive resolver comment this PR added to `install.sh` (line 245). That
paragraph squashes into main's history as a false statement — same standard
applied to the homebrew claim last round. Drop it, or restate what actually
remains deferred (`installers/homebrew/hew.rb` is still a stale scaffold).

## Not blocking

The fixture's comment on the rc10/rc9 case claims more than the case proves:

```
# 0.6.0-rc10 beats 0.6.0-rc9 — catches substr(\$4, 3) losing its + 0 numeric
# coercion and comparing "10" < "9" lexically.
```

I removed the `+ 0` from `install.sh` and the fixture still passed all five
cases. `sprintf("%06d", ...)` coerces its argument in awk, so the `+ 0` is
redundant and the case cannot catch its loss — what it actually catches is
the padding, which the padding mutation already covers. The behaviour is
correct either way; only the comment overstates. Same class as the `-beta1`
comment fix already in this PR, and worth folding into the rebase push.

slepp:
Re-review of the head commit (`2320f9f`). Both blockers from the last round
are applied and I verified each against the source, not the claim.

**Rebase (was blocker 1) — resolved.** `mergeStateStatus` is `BLOCKED` with
`mergeable: MERGEABLE`, not `DIRTY`. The README merge kept both sections: the
new `install.sh` `--stable` paragraph and main's PowerShell
`-Stable`/`-DryRun` section survive side by side, and the Package Managers
table carries the `hew@stable` row without dropping main's reflow.

**CI on this head (was blocker 2) — resolved.** Run `33682811672` has
`head_sha: 2320f9fa`, so the `test-install-version-resolution` gate is
executing on the code being merged rather than on the first commit.

**Cross-repo contract (blocker 1 of the first round) — still true.**
`update-formula.yml` on `hew-lang/homebrew-hew` `main` calls
`update_formula Formula/hew.rb` unconditionally and gates
`Formula/hew@stable.rb` behind `case "${VERSION}" in *-*)`, and
`Formula/hew@stable.rb` exists on that repo. The release.yml `if:` change and
the README row are backed by automation.

**rc10/rc9 comment (non-blocking last round) — folded in** as `2320f9f`; the
comment now attributes the case to the padding rather than to the redundant
`+ 0`.

One thing blocks the merge.

## The Deferred paragraph is false again

The PR body now reads:

```
Deferred (tracked in `.tmp/TODO.md`, not touched here):
`installers/homebrew/hew.rb` is a stale manual reference scaffold not read
by CI.
```

That file does not exist. `abdbde44` ("build(installers): apply the
release-candidate policy to Windows (#3217)") deleted it:

```
D	installers/homebrew/hew.rb
```

and its own body says why:

```
- installers/build-packages.sh, installers/homebrew/hew.rb: delete the
  stale local Homebrew formula scaffold and its syntax-check target
```

`git merge-base --is-ancestor abdbde44 2320f9fa` succeeds, so this PR's head
already contains the deletion; `gh api
repos/hew-lang/hew/contents/installers?ref=build/install-newest-rc` lists no
`homebrew/` entry at all. The paragraph describes deferred work on a file
this branch does not have, and it squashes into main's history as a false
statement - the same standard that took out the `install.ps1` paragraph last
round and the homebrew-automation claim the round before.

The suggestion I made last round ("restate what actually remains deferred -
`installers/homebrew/hew.rb` is still a stale scaffold") was stale by the
time it was applied, because #3217 landed the deletion. My error, not yours.
Drop the paragraph entirely unless something genuinely remains deferred, in
which case name that instead.

## Not blocking

The rewritten Package Managers table's separator row is wider than its
content rows:

```
| ------------------------ | ----------------------------------------------------------------------------------------------------- |
| macOS / Linux (Homebrew) | `brew install hew-lang/tap/hew` (newest release, rc included) or `hew@stable` (final releases only) |
```

GitHub renders this correctly and no CI gate checks Markdown formatting, so
it is cosmetic only.

PR #3216, opened 2026-09-02
docs(syntax): make syntax-data.json and the spec match the compiler

slepp:
## Why

docs/syntax-data.json and docs/specs/HEW-SPEC-2026.md had drifted from what the compiler actually accepts - a type it rejects, a separator it hard-errors on, keywords with no parser consumer, contextual identifiers missing from the data, a stale version, and spec text describing generator forms as deferred when they compile and typecheck today.

## What

- **syntax-data.json**: removed `never` from `types.primitive` (`hew check` rejects it - `Ty::Never` has no lexer token or user-facing spelling, only the internal `!` marker exists). Renamed `path_separators.deprecated` to `removed` - `::` is `E_PATH_LEGACY_SEPARATOR`, a hard error, not a warning. Moved `super`/`budget` from `declarations`/`supervisor_config` into `reserved_unused`; neither has a parser consumer beyond the generic contextual-identifier-fallback path. Added `consume`/`clone` to `contextual_identifiers`, both special-cased in `hew-parser/src/parser/core.rs`. Bumped the stale `version` field to the workspace version and added a `hew-lexer` test asserting they stay in sync going forward, since nothing in-tree generates this file today.
- **HEW-SPEC-2026.md**: `async gen fn` and `receive gen fn` were listed as deferred in two places (§4.12 and the edition changelog) while both compile and typecheck; only `Lazy<T>` and `#[prefetch(N)]` remain deferred, per HEW-FUTURE.md §1.6. The §3.11.1 machine transition example used a bare `StateB` variant body, which the checker rejects as `E_BARE_VARIANT_EXPR`; dot-qualified it.
- **editors/**: regenerated `editors/nano/hew.nanorc` from the corrected syntax-data.json via `tools/downstream/generate-nano.mjs` (this also drops `record`, which isn't in `ALL_KEYWORDS`/syntax-data.json either - stale in sublime/emacs too, left as a follow-up). Hand-synced `editors/sublime/Hew.tmLanguage.json`'s `super`/`budget`/`never` categories and its `clone`/`consume` contextual entries; dropped `never` from `editors/emacs/hew-mode.el`'s builtin-type list (emacs highlights reserved words like `try`/`catch` uniformly, so `super`/`budget` staying there is consistent with its existing pattern, not drift).
- **scripts/doc-test-expected-failures.txt**: the machine-example fence's content hash changed with the dot-qualify fix; re-verified its failure (a required `field: Type` on `StateB` makes the unit-state `.StateB` shorthand a constructor-kind mismatch, not the old bare-variant error) and updated the label.
- **workspace CLAUDE.md**: fixed the sync-order step that still pointed at hew.sh's `tools/generate-monarch.mjs`, retired when hew.sh moved to TextMate grammars, and noted that a version bump needs `docs/syntax-data.json` alongside `Cargo.toml`.
- vscode-hew and vim-hew now drift from the same reclassification; filed hew-lang/hew#3219 to sync them, since they are separate repos with their own push convention.

hew.sh: deleted the dead `tools/generate-monarch.mjs` script and its `.env.example` reference in a separate commit pushed directly to hew.sh main (`chore: remove the retired monarch generator`).

## Test

- `cargo test -p hew-lexer --lib`, including `syntax_data_json_covers_all_keywords` (extended with the version assertion) and `sublime_grammar_keywords_match_lexer_surface`.
- `HEW_BIN=<release hew> scripts/corpus-ratchet.sh doc-fences` - passes with the updated expected-failure label.
- `scripts/sync-downstream.sh --check` - confirms tree-sitter-hew needed no change (super/budget/never were never in its grammar rules) and editors/nano now matches the generator's output.
- Verified against a built `hew` binary that `never` and `::` are rejected, and that plain `gen fn`/`async gen fn` and the bare-vs-dot-qualified machine transition body behave as described above.

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -46,3 +46,12 @@ JOURNEY.md
 # WHY: the generator and prettier disagree about array formatting.
 # WHEN: remove once the generator emits prettier-shaped output itself.
 scripts/cabi-surface.json
+
+# Hand-maintained syntax authority; prettier collapsing its arrays turns a
+# one-keyword edit into a whole-file reflow that buries the real change.
+#
+# WHY: nothing generates this file's shape and prettier disagrees with the
+# hand-maintained array formatting.
+# WHEN: remove once a real generator writes this file and emits
+# prettier-shaped output.
+docs/syntax-data.json
--- a/docs/internal/versioning.md
+++ b/docs/internal/versioning.md
@@ -149,6 +149,9 @@ the Hew language edition; it governs how the compiler's own source compiles.
 
 - **Workspace `Cargo.toml`** (`[workspace.package] version = "..."`) — the
   source of truth. All crates inherit via `version.workspace = true`.
+- **`docs/syntax-data.json`** — its `version` field is hand-maintained and
+  checked against `CARGO_PKG_VERSION` by a `hew-lexer` test; a release bump
+  that skips it fails `cargo test -p hew-lexer`.
 - **`CHANGELOG.md`** — `[Unreleased]` becomes `[X.Y.Z]` at tag time.
 - **`docs/releases/vX.Y.Z.md`** — curated narrative release notes
   (when the release warrants them; PATCH releases may skip this).
--- a/docs/release-runbook.md
+++ b/docs/release-runbook.md
@@ -13,9 +13,9 @@ This is the concrete expansion of the `ci-full-run-pre-tag` todo.
 - [ ] CHANGELOG.md has either a populated `[Unreleased]` section or the dated
       `[X.Y.Z]` section for the intended release
 - [ ] Curated GitHub release notes are drafted at `docs/releases/vX.Y.Z.md`
-- [ ] `workspace.package.version` in `Cargo.toml`, `Cargo.lock`, the intended
-      changelog record, and `docs/releases/vX.Y.Z.md` all name the candidate
-      that will be tagged
+- [ ] `workspace.package.version` in `Cargo.toml`, `Cargo.lock`,
+      `docs/syntax-data.json`, the intended changelog record, and
+      `docs/releases/vX.Y.Z.md` all name the candidate that will be tagged
 
 ## Phase 1 — Assemble the candidate
 
@@ -57,9 +57,10 @@ git log --oneline -5  # confirm expected HEAD
 ## Phase 2 — Establish the release identity
 
 > **Prerequisite:** Any required version bump must update `Cargo.toml`'s
-> workspace version, every lockfile, the dated changelog record, and the exact
-> `docs/releases/vX.Y.Z.md` file together. Tagging a commit where any one of
-> those identities differs produces a split release record.
+> workspace version, every lockfile, `docs/syntax-data.json`, the dated
+> changelog record, and the exact `docs/releases/vX.Y.Z.md` file together.
+> Tagging a commit where any one of those identities differs produces a split
+> release record.
 
 ```bash
 # Set the intended version in the root [workspace.package], if needed.
@@ -81,8 +82,8 @@ release_version="$(scripts/workspace-version.py)"
 release_tag="v${release_version}"
 test -f "docs/releases/${release_tag}.md"
 
-git add Cargo.toml Cargo.lock hew-parser/fuzz/Cargo.lock CHANGELOG.md \
-  "docs/releases/${release_tag}.md"
+git add Cargo.toml Cargo.lock hew-parser/fuzz/Cargo.lock docs/syntax-data.json \
+  CHANGELOG.md "docs/releases/${release_tag}.md"
 git commit -m "chore(release): prepare ${release_tag}"
 ```
 
--- a/docs/specs/HEW-SPEC-2026.md
+++ b/docs/specs/HEW-SPEC-2026.md
@@ -2672,7 +2672,7 @@ machine Name {
     state StateB { field: Type; }         // state with data
 
     // Transitions: on Event: Source => Target { body }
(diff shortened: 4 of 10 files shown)


slepp:
The five things this PR claims all check out under re-verification, and two of them found a gap that has to close before it lands.

Verified good:

- `never` is genuinely rejected. `fn boom() -> never { panic("x") }` gives `error: unknown type 'never'`, so removing it from `types.primitive` is grounded.
- `super` and `budget` have no grammar consumer. `Token::Super` appears only in the lexer's token enum and its spelling table; `Token::Budget` appears there plus `contextual_keyword_name` in `hew-parser/src/parser/core.rs`, which is the identifier-fallback path, not a supervisor rule. Supervisor budgets are spelled `intensity: N within <duration>`, so `reserved_unused` is the right home for both.
- The new version assertion is not vacuous. Mutating `docs/syntax-data.json` to `0.6.0-rc4` and running `cargo test -p hew-lexer --lib syntax_data` fails with `left: Some("0.6.0-rc4") right: Some("0.6.0-rc3")`. A missing field would fail too, since `as_str()` on a null returns `None`.
- The doc-fence ledger relabels `spec-0081` in place rather than adding a row, the checksum matches the extracted fence, and `scripts/corpus-ratchet.sh doc-fences` passes on this branch. `.StateB` is also the spelling the compiler itself recommends: the `E_BARE_VARIANT_EXPR` diagnostic in `hew-types/src/check/diagnostics.rs` says "use `.{name}`", and the machine corpus under `tests/core-matrix/cells/` uses that form throughout.
- `async gen fn` compiles and typechecks today: `tests/hew/generator_iterator_trait_test.hew` declares `async gen fn async_counter(n: i64) -> i64` and runs it. The spec no longer calling it deferred is correct.
- `editors/nano/hew.nanorc` is regenerated, not hand-edited. `node tools/downstream/generate-nano.mjs docs/syntax-data.json` followed by `diff dist/hew.nanorc editors/nano/hew.nanorc` is byte-identical. Sublime and emacs have no generator, so hand-syncing them is the right move.

Two required changes:

**1. `void` fails the same test that removed `never`.**

```diff
       "void",
-      "never",
       "duration",
```

`fn f() -> void {}` gives `error: unknown type 'void'`, exactly like `never`. Both live in `Ty::is_well_known_type_name` in `hew-types/src/ty.rs` but neither is in `Ty::from_name`, so neither is a spellable type; the only Hew-level appearances of `void` in `std/`, `tests/`, and `examples/` are in comments and C signatures. The stated rationale, no lexer token and no user-facing spelling, is word for word true of `void`, and the emacs hunk drops `never` while keeping `void` two words to its left:

```diff
-    "bool" "char" "string" "bytes" "void" "never" "duration" "instant"
+    "bool" "char" "string" "bytes" "void" "duration" "instant"
```

Drop `void` the same way `never` went: out of `types.primitive`, regenerate nano, hand-sync the sublime `storage.type.primitive.hew` match and the emacs builtin-type list. Otherwise the commit title is not true of the array it edits.

**2. A release that follows the runbook now goes red with no in-repo instruction.**

```diff
+        // WHY: nothing generates docs/syntax-data.json's `version` field
+        assert_eq!(
+            data["version"].as_str(),
+            Some(env!("CARGO_PKG_VERSION")),
```

This makes `docs/syntax-data.json` a file that must move with the workspace version, and `cargo test -p hew-lexer` is the gate. `docs/release-runbook.md` Phase 2 states the prerequisite as "`Cargo.toml`'s workspace version, every lockfile, the dated changelog record, and the exact `docs/releases/vX.Y.Z.md` file together" and its `git add` line lists those files; `docs/internal/versioning.md` "Where the version lives" enumerates the same set. Neither names `docs/syntax-data.json`. The note the PR description mentions went into the workspace `CLAUDE.md` one directory above this repo, which is not in any git repo, so the only record of the new obligation is machine-local. Add the file to both lists. Both are in `.prettierignore`, so hand-wrap and check `git diff --stat` for reflow.

**3. Minor: give the `record` deferral an issue.**

`record` is still in sublime's `keyword.declaration.hew` match and in the emacs keyword list while being absent from `ALL_KEYWORDS` and `syntax-data.json`. The description calls it a follow-up, and #3219 covers only the vscode-hew and vim-hew sync. Either drop it here or file it, so the deferral is tracked rather than prose.

Two notes that do not block:

- The `spec-0081` label names `E_PATH_KIND_MISMATCH`. The ratchet only compares checksums, so that specific code is unverified from here; the built binary disappeared mid-review while another build ran. Worth a one-line confirmation in the thread.
- The prettier hook reflowed all of `docs/syntax-data.json`, which buries the real change in a whole-file diff. Consumers parse it as JSON so nothing breaks, but it is the same situation `.prettierignore` already documents for `scripts/cabi-surface.json`. A follow-up adding it there would keep the next edit reviewable.

slepp:
Confirming the one item I left unverified: the built binary came back, and `hew check` on the extracted `spec-0081` fence reports exactly what the new label says.

```
spec-0081.hew:18:35: error: E_PATH_KIND_MISMATCH: variant `Name.StateB` does not use this constructor form
 18 |     on EventX: StateA => StateB { .StateB }
```

It is the only blocking error in the fence, and it cannot be resolved while the illustration uses the metasyntactic `field: Type`: `StateB { field: 0 }` gives `type mismatch: expected `Type`, found `i64``, and `event.payload` does not exist on the payload-free `EventX`. So SNIPPET is the honest category and the IMPL-GAP label it replaces was wrong. That item is settled; the two required changes above still stand.

Issue #3218, opened 2026-09-02
Stream<string> actor-state field SIGSEGVs at teardown after for-await drain


slepp:
## Problem

`hew run` SIGSEGVs at teardown (exit 139, `hew: fatal synchronous hardware
fault`) for a program that holds a `Stream<string>` as an **actor state
field**, drains it with `for await` in a handler, and lets `main` write to
and close the producer side after the handler starts. Reproduces 3/3 runs.

## Expected

The program should complete and exit 0, printing the drained item count
(`1`) and the `after-close` thread dump, with no crash at teardown.

## Repro

```hew
import std.fs;
import std.stream;

fn dump(label: string) {
    println(label);
    let tids = fs.list_dir("/proc/self/task");
    for t in tids {
        let name = fs.read("/proc/self/task/" + t + "/comm");
        println("  " + name.trim());
    }
}

actor Consumer {
    var input: Stream<string>;
    receive fn drain() {
        var count = 0;
        for await item in input {
            count = count + 1;
        }
        println(count);
    }
}

fn main() -> i64 {
    let (sink, input) = stream.pipe(4);
    dump("before-spawn");
    let c = spawn Consumer(input: input);
    c.drain();
    sleep(300ms);
    dump("while-consumer-parked-on-empty-stream");
    sink.write("a");
    sink.close();
    sleep(200ms);
    dump("after-close");
    0
}
```

Command: `hew run repro.hew`

Output (3/3 identical):

```
before-spawn
  hew_run_bin
  hew-worker-0
  ...
while-consumer-parked-on-empty-stream
  hew_run_bin
  hew-worker-0
  ...
1
after-close
  hew_run_bin
  hew-worker-0
  ...
hew: fatal synchronous hardware fault
```

exit code 139 (SIGSEGV) after `after-close` prints, i.e. at process teardown
once the program's own output is complete.

## Notes

- Isolating the `Stream<string>` field placement: a program that builds the
  pipe *inside* the handler instead of taking it as actor state does not
  crash and completes cleanly at the same thread-count baseline. So holding
  the stream as an actor state field, not the `for await` drain itself, is
  the distinguishing factor.
- Found incidentally while probing OS-thread usage of `fork` and generators
  for U370 (unrelated investigation); not minimized further and not
  confirmed as a compiler defect versus a probe misuse of the `Stream`
  actor-state-field API.
[...]

PR #3220, opened 2026-09-02
test(e2e): fence the thread count of generators and forks

slepp:
## Why

U370 verified that the four generator surfaces already run on `llvm.coro`
frames with zero OS threads, while every `fork` child still gets a dedicated
`std::thread::spawn` OS thread, one per child. Neither fact had a test, so
nothing would catch a regression on the generator side or record the fork
gap as tracked rather than silent.

## What

- **hew-cli**: `hew-cli/tests/coro_thread_fence_e2e.rs` (Linux-only, reads
  `/proc/self/task` from inside the running program):
  - `generator_iteration_creates_no_os_threads` samples the thread count
    around `gen fn` + for-in, `gen {}` block, `async gen fn` + for-await, and
    `receive gen fn` + for-await, and asserts the peak equals the pre-loop
    baseline for each. Each block also counts and asserts its own loop
    iterations (500 / 3 / 50 / 200) so a surface that yields nothing cannot
    pass on a vacuous `peak == pre`. Passes today; a regression fence so the
    P4 concurrency lane cannot quietly re-thread the pump.
  - `fork_children_create_no_os_threads` asserts the same ideal for N fork
    children parked on a long sleep inside `scope{}`, parameterized over N
    (1 and 5) so a per-child leak is distinguishable from a constant
    per-scope overhead. Fails today by exactly `N + 1` (the N children plus
    the reporter fork that samples while they are all live).
- **scripts/nextest-expected-failures.tsv**: ledgers
  `fork_children_create_no_os_threads` as an expected failure
  (`P4-CONCURRENCY: fork bodies run on OS threads until CORO lands`) so the
  P4 pure-coroutine `fork` lane must flip the row, not delete it.
- Filed #3218 alongside for an unrelated crash noticed during the same
  investigation: a `Stream<string>` held as actor state SIGSEGVs at teardown
  after a `for await` drain (3/3 reproductions). Not fixed here — recorded so
  it isn't lost.

## Test

- `cargo nextest run -p hew-cli --test coro_thread_fence_e2e`:
  `generator_iteration_creates_no_os_threads` passes;
  `fork_children_create_no_os_threads` fails with `N=1: pre=34 during=36
  delta=2; N=5: pre=34 during=40 delta=6` (both exactly `N+1`), matching the
  ledgered expectation.
- Ratchet confirmation: ran `cargo xtask nextest-ratchet` against that
  binary's junit output and the ledger — the tool reports
  `hew-cli::coro_thread_fence_e2e :: fork_children_create_no_os_threads` as
  `skipped` with the ledgered reason, i.e. the row is honoured (the run's
  other reported mismatches are `expected failure is absent` for ledger rows
  belonging to test binaries this scoped run didn't execute, an artifact of
  running one test file rather than the full workspace suite — not a false
  positive against the new row).

### DIFF
--- a/hew-cli/tests/coro_thread_fence_e2e.rs
+++ b/hew-cli/tests/coro_thread_fence_e2e.rs
@@ -1,0 +1,306 @@
+//! Thread-count regression fences for generators and `fork` (U370).
+//!
+//! Both tests read `/proc/self/task` from inside the running Hew program to
+//! count OS threads at a known point, so the count is exact, not sampled from
+//! outside. Linux-only: `/proc/self/task` has no cross-platform equivalent
+//! here, and the portable alternative (a runtime-owned spawn counter) does
+//! not exist yet.
+//!
+//! - `generator_iteration_creates_no_os_threads` passes today: all four
+//!   generator surfaces (`gen fn`, `gen {}`, `async gen fn`, `receive gen fn`)
+//!   already run on `llvm.coro` frames, not OS threads. It is a regression
+//!   fence so the P4 concurrency lane cannot quietly re-thread the pump.
+//! - `fork_children_create_no_os_threads` FAILS today: every `fork` child
+//!   gets its own `std::thread::spawn` OS thread
+//!   (`hew-runtime/src/task_scope.rs:1479`), one per child. It is ledgered in
+//!   `scripts/nextest-expected-failures.tsv` as an expected failure so the P4
+//!   pure-coroutine `fork` lane must flip it, not delete it.
+
+#![cfg(target_os = "linux")]
+
+mod support;
+
+use support::{require_codegen, run_bounded_hew_run};
+
+/// Extracts the integer printed on the line immediately after a given label
+/// line (`println("label"); println(value);` in the fixture).
+fn value_after_label(stdout: &str, label: &str) -> i64 {
+    let mut lines = stdout.lines();
+    while let Some(line) = lines.next() {
+        if line.trim() == label {
+            let value_line = lines
+                .next()
+                .unwrap_or_else(|| panic!("no line after label {label:?}; stdout:\n{stdout}"));
+            return value_line
+                .trim()
+                .parse()
+                .unwrap_or_else(|e| panic!("label {label:?}: {e}; line: {value_line:?}"));
+        }
+    }
+    panic!("label {label:?} not found in stdout:\n{stdout}");
+}
+
+/// U370: all four generator surfaces already run on coroutine frames, not
+/// OS threads — the `cut-gen-coro` cutover held. Each block samples the
+/// thread count immediately before its loop starts, then samples again on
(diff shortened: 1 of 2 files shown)


slepp:
The fences are well aimed and the CI evidence backs the claims: `generator_iteration_creates_no_os_threads` passes in Linux gates shard 2/3, and `fork_children_create_no_os_threads` fails with `N=1: pre=6 during=8 delta=2; N=5: pre=6 during=12 delta=6` — exactly `N+1` both times. The Linux gate (`#![cfg(target_os = "linux")]` ahead of `mod support;`) is right, the ledger row is `linux`-scoped and names `P4-CONCURRENCY`, and the `run_fork_thread_probe` doc carries a proper WHY / WHEN / WHAT for the free-function sleep seam. One blocker, though.

**1. The second assertion loop makes the fork fence unpassable, so the ledger row can never be flipped.**

```rust
    for &(n, pre, during) in &results {
        assert_eq!(
            during - pre,
            n + 1,
            "N={n}: delta must be exactly N+1 (N nappers + 1 reporter), not a \
             constant; full run: {}",
            report()
        );
    }
```

`assert_eq!` panics, so this loop is only reached when every `during == pre` assertion above it held — that is, when every delta is 0. It then asserts `0 == n + 1` and fails. The test therefore fails today through the first loop (CI panics at `coro_thread_fence_e2e.rs:200`, the `during == pre` assert) and would fail again through this second loop the moment the P4 pure-coroutine `fork` lane makes the ideal true. There is no state in which it passes, which contradicts the PR's own intent that the lane "must flip the row, not delete it" — the lane would have to edit the test.

The comment above the loop ("A per-N breakdown that survives even when the assertion above never fires") reads `assert_eq!` as non-fatal. Delete the second loop: the per-N deltas it wants to surface are already in `report()`, which the first assertion's message prints in full. If both facts really need asserting, collect mismatches across all N and panic once at the end rather than asserting today's shape and the ideal shape in the same run.

**2. The generator fence goes vacuous if a generator yields nothing.**

`peak_genfn` and friends start at `pre` and only move inside the loop body:

```
    var peak_genfn = pre_genfn;
    for v in counter(500) {
```

A regression that makes a surface produce zero items leaves `peak == pre` and the block passes having sampled nothing. Please print and assert a per-block iteration count (500 / 3 / 50 / 200) alongside the peak so each block proves it actually ran. `v` is unused in all four loops today, so the counter is nearly free.

**3. Trim the Test section.** "clippy … clean for the new file" and "`cargo fmt` applied" are CI-enforced and listed under never-include in CLAUDE.md's PR guidance. The `delta=2` / `delta=6` numbers should stay — those are the observable being fenced, not pass counts.

Non-blocking: exact-equality thread counts sample a scheduler that can grow lazily. The timer-tick warm-up handles the one known source and the ledger row is `failure`-typed either way, so this is a note rather than a change request — but if the row ever needs to become `timeout`/`signal`-tolerant, that is why.

PR #3221, opened 2026-09-02
feat(std): every public error type prints itself

slepp:
## Why

None of std's public error enums implemented `Display`. `println(e)` on an
`IoError` was refused by the checker, and six modules had each invented
their own differently named message function (`io_error_message`,
`path_error_message`, `net_error_message`, `write_error_message`,
`jwt.error_message`, `encrypt.error_message`) while the rest had nothing.

## What

- Implemented `Display for <Error>` on every public std error enum that the
  compiler can dispatch to today: `fs.IoError`, `path.PathError`,
  `os.EnvError`, `process.ProcessError`, `time.cron.CronError`,
  `io.closable.CloseError`, `crypto.jwt.JwtError`,
  `crypto.encrypt.CryptoError`, `concurrency.lifecycle.Error`,
  `encoding.{toml,yaml,json}.ParseError`, `text.template.TemplateError`,
  `net.{NetError,AttachError,WriteError}`, `metrics.MetricsError`,
  `link_monitor.MonitorError`.
- Deleted the six per-module `*_error_message` functions (plus
  `process.hew`'s private one) and repointed every caller to the enum's own
  `Display` (either bare interpolation or `to_string(err)`), including the
  callers in `tests/hew/` (`encrypt_try_open_oracle`,
  `crypto_return_abi_oracle`) and the embedded Hew sources in
  `hew-cli`'s `measured_os_io_string_retention_canary`,
  `match_bound_payload_drop_leak_oracle`, and
  `match_result_record_destructure_oracle` integration tests.
- The three `impl Display` blocks added to `fs.hew`, `net.hew`, and
  `path.hew` carry a `//` comment rather than `///`: `hew fmt` deletes a
  doc comment immediately preceding an `impl` block. Filed as #3238;
  restore `///` once it's fixed.
- `SendError`, `LookupError`, `AskError`, `TimeoutError`, `LinkError`
  (declared in the `std/builtins.hew` prelude) cannot get a working
  `Display` impl without a compiler change: the checker accepts
  `impl Display for TimeoutError` and `println`/f-string use of it, but
  `hew run` fails closed at HIR dispatch (`E_NOT_YET_IMPLEMENTED: display
  dispatch: missing std.builtins.TimeoutError::fmt`) because these five
  types are registered through a synthetic-item path that never reaches the
  imported-impl-body pass. Filed as #3224 with the repro and the exact HIR
  symbols; each of the five enums is marked at its declaration in
  `std/builtins.hew` referencing the issue; not part of this PR.
- `concurrency.ScopeError<E>` is a `pub type`, not an enum, and was left out
  of this pass as a separate, deliberate decision.

## Test

- One `println`/`to_string` test per module proving the `Display` message,
  e.g. `test_process_error_display_prints_the_carried_detail`,
  `test_path_error_display_prints_the_carried_detail`.
- Negative-control vertical-slice reject fixtures proving the deleted
  public message-function names no longer resolve:
  `tests/vertical-slice/reject/{encrypt,fs_io,jwt,net,net_write,path}_error_message_removed.hew`.
  `process.process_error_message` was module-private (never externally
  reachable), so its deletion needed no reject fixture — confirmed instead
  with a scratch `hew check` showing `no function process_error_message in
  module process`.
- `make checked-mir-verify` confirms the committed
  `examples/v05/checked-mir/golden/MANIFEST.sha256` is still in sync at
  HEAD; no regen needed.
- `make test-hew-ratchet` passes with no new tracked failures.
- `cargo nextest run -p hew-cli --test measured_os_io_string_retention_canary
  --test match_bound_payload_drop_leak_oracle --test
  match_result_record_destructure_oracle`: the `cleanup_cfg_*` unit test
  passes, and the macOS-gated poisoned-allocator/leak-slope tests report
  skipped on Linux, not failed. The canary's
  `shipped_os_io_wrappers_emit_all_measured_calls_and_caller_releases`
  fails on `%local_95` (the `process.try_run` output's `stdout` field, a
  code path this PR does not touch) on Linux too - confirmed against
  `main`'s own CI run (`Linux gates (shard 2/3)` and
  `Build & test (macOS arm64)`, run 33647946852) with the identical
  panic. Pre-existing, not this PR's regression. `local_61`, the slot
  the `path.path_error_message` to `Display` swap touches, is checked
  by the same canary and passes before the loop reaches `local_95`.
- The macOS-gated `net.net_error_message`/`to_string(err)` callers in
  `match_bound_payload_drop_leak_oracle.rs` and
  `match_result_record_destructure_oracle.rs` were verified locally
  instead by extracting their embedded Hew sources and running
  `hew check`/`hew run` against the branch's own compiler: each checks
  clean and, where the source runs the `Err` arm (the cancellation-window
  shape), dispatches `Display` at runtime; the two record-destructure
  sources take the `Ok` arm on the happy path, so their `Err` arm print
  is confirmed only by `hew check`, not execution.


slepp:
Review of `feat/std-error-display` at `cfb40ab`. The `Display` work itself is in good shape: every public error enum in `std/` outside the prelude now has an impl, each impl covers every variant, and each module gained a test that pins the exact message rather than just asserting non-empty. The six reject fixtures are real negative controls. Four things block a merge.

### 1. Five call sites still call the deleted functions

The PR deletes the public message functions:

```diff
-pub fn error_message(err: CryptoError) -> string {
-pub fn io_error_message(e: IoError) -> string {
-pub fn net_error_message(e: NetError) -> string {
-pub fn path_error_message(err: PathError) -> string {
```

but these callers were not repointed and no longer resolve:

- `tests/hew/encrypt_try_open_oracle.hew:37` and `:51` - `encrypt.error_message(err)`
- `tests/hew/crypto_return_abi_oracle.hew:25` - `encrypt.error_message(err)`
- `hew-cli/tests/measured_os_io_string_retention_canary.rs:37` - `fs.io_error_message(error)` in `MEASURED` Hew source
- `hew-cli/tests/measured_os_io_string_retention_canary.rs:50` - `path.path_error_message(error).len()` in the same source
- `hew-cli/tests/match_bound_payload_drop_leak_oracle.rs:93` - `net.net_error_message(err)` in `NESTED_ENUM_CANCEL_SCRIBBLE_SOURCE`

The first three are in `tests/hew/`, which `make test-hew-ratchet` compiles as a directory; the last three sit in `hew-cli` integration tests that compile their embedded Hew at test time. The evidence quoted in the description is `scripts/corpus-ratchet.sh stdlib`, which only walks `std/` - it cannot see either lane, which is why these survived. Please repoint all five to `to_string(...)` or bare interpolation and run `make test-hew-ratchet` plus the two `hew-cli` tests before the next push.

The `path.path_error_message(error).len()` site is worth a second look on its own: that source is the string-retention canary, and swapping a `pub fn` call for a `Display` dispatch changes the caller-release shape the test asserts on.

### 2. The five prelude enums are unmarked in `std/builtins.hew`

`SendError`, `LookupError`, `AskError`, `TimeoutError` and `LinkError` are skipped, and nothing at their declaration says so. The `LESSONS.md` entry is excellent and does name the mechanism (`plan_imported_impl_bodies`, the synthetic-item path, the `is_builtin_display_impl` allowlist), but a reader of `std/builtins.hew:49`, `:83`, `:106`, `:166`, `:182` sees only five error enums that silently lack the impl every sibling now has. That is the unmarked-shortcut case: mark each with why it is blocked, what makes it obsolete, and what the real fix is.

The follow-up also needs a home outside `.tmp/TODO.md`. That path is gitignored, so the deferral disappears the moment the branch is cleaned up. File the compiler gap as an issue and reference its number from both the `builtins.hew` comments and `LESSONS.md`.

### 3. The branch is behind main

`mergeStateStatus` is `DIRTY` / `mergeable` is `CONFLICTING`. Rebase onto `origin/main` and regenerate `examples/v05/checked-mir/golden/MANIFEST.sha256` afterwards if the conflict touches it.

### 4. Drop the count from the description

"(19 enums)" is a count, and the actual set of `impl Display for` additions under `std/` is 18. Counts go stale and the house style keeps them out of descriptions, so the parenthetical can just go.

### Not blocking

`std/failure.hew`'s `CrashAction` and `CrashKind` are the only other public `std` enums without a `Display`. Neither is ever an `Err` payload, so leaving them out of an error-printing pass reads as correct - no change wanted, just noting it was checked.

slepp:
Re-review at `79e03d1`. All four items from the last pass are applied and verified:

1. **Call sites repointed.** `git grep error_message` over the branch finds no live caller of `encrypt.error_message`, `fs.io_error_message`, `path.path_error_message`, `net.net_error_message` or `net.write_error_message` outside the reject fixtures. The `measured_os_io_string_retention_canary` source now reads `Err(error) => to_string(error)` / `to_string(error).len()`, and `match_bound_payload_drop_leak_oracle.rs`'s cancellation window reads `let message = to_string(err);`. The two extra `match_result_record_destructure_oracle.rs` sites the first pass missed were caught too.
2. **Prelude enums marked.** All five declarations in `std/builtins.hew` carry WHY / WHEN obsolete / WHAT-the-real-fix referencing #3224, and `LESSONS.md:831` names the issue. #3224 is open.
3. **Rebase.** `mergeStateStatus` is `BLOCKED` on pending checks, `mergeable` is `MERGEABLE` — no longer `DIRTY`.
4. **Count dropped.** `(19 enums)` is gone from the description.

One new blocker, and it is one I got wrong last time — I called the six reject fixtures "real negative controls" without checking that anything runs them.

### The six reject fixtures are never executed

`tests/vertical-slice/run.sh` enumerates its reject fixtures by explicit path; there is no directory walk. On `main`, 269 of the 275 files in `tests/vertical-slice/reject/` are named in the harness (the six that are not are stale leftovers). `run.sh` does not appear in this PR's diff, and

```
git grep -n error_message_removed pr3221-rr
```

hits only the six fixture files themselves. So

```hew
+// `path.path_error_message` was replaced by `impl Display for PathError`; the
+// deleted function name must no longer resolve.
+import std.path;
+
+fn main() {
+    let e = PathError.GlobFailed("x");
+    println(path.path_error_message(e));
+}
```

is a file `make test-vertical-slice` never opens, and the description's

> Negative-control vertical-slice reject fixtures proving the deleted public message-function names no longer resolve

is not true as shipped. The deletion is still covered for callers by `make test-hew-ratchet` and the stdlib ratchet; what is missing is the control that the *names* stopped resolving.

The fixtures themselves are fine — I extracted all six and ran `hew check` on each with `main`'s compiler, where the deleted functions still exist: all six report `OK`. So on this branch the deleted function is the only error each one raises, and registration is the whole fix.

Please register them in `run.sh` with `expect_check_fail_contains`, pinning the diagnostic rather than a bare non-zero exit, so a fixture that later breaks for an unrelated reason cannot pass vacuously. The substring form is the one your own scratch check produced: `no function path_error_message in module path`. The `std_panic_wrapper_*` block at `run.sh:2367-2375` is the natural place to slot them.

PR #3223, opened 2026-09-02
test(std): compile the standard library doc fences

slepp:
## Why

The doc-fence ratchet (`make test-doc-examples`, run in `ci-shard-1`) only
sourced the language guide, the spec, and `docs/language/*.hew`, so std's
232 doc-comment fences never compiled as part of CI. A stale std example
could sit broken indefinitely with nothing catching it.

## What

- **corpus-ratchet.sh**: added a `std/**/*.hew` enumerator and a std-specific
  fence extractor (`doc_fence_extract_std`) alongside the existing
  guide/spec extractor. std fences differ from the guide/spec ones on two
  axes the existing extractor doesn't handle: item-level `///` comments (not
  only module-level `//!`), and a bare ` ``` ` open (implicit hew) instead of
  an explicit ` ```hew ` tag. The extractor tracks a third "inside a
  non-hew-tagged fence" state so a bare ` ``` ` that closes e.g. a
  ` ```text ` block is never mistaken for the open of the next real fence.
  It also trims leading whitespace before matching `///`/`//!` markers, so
  indented item-level docs on trait/impl members are extracted correctly
  instead of being blanked to nothing.
- **std/**: fixed every std doc fence that failed to compile. Most were
  terse call-and-comment snippets missing the `import` + `fn main()`
  wrapper the module-level examples already use, or bare `Ok`/`Err`/enum
  patterns missing their leading dot. A few had real content bugs — see the
  `docs(std): fix the remaining std doc fences` commits for the list
  (`Stream<string>` chunking vs. a nonexistent `.lines()`, an opaque handle
  passed through a `receive fn`, `respond_stream`'s return type, an
  untagged ASCII diagram, `panic` given a `NetError` instead of a string).
- **Makefile**: wired the new `scripts/tests/test_std_doc_fence_extraction.sh`
  self-test into `doc-ratchet-selftest` — it existed but no target ran it.

## Test

- `make test-doc-examples` passes: the doc-fence ratchet's expected-failure
  set is unchanged from `origin/main`, all pre-existing guide/spec fences,
  so every std fence either compiles or was already a tracked expected
  failure before this branch touched it.
- `scripts/tests/test_std_doc_fence_extraction.sh` (now run via
  `make doc-ratchet-selftest`) pins the tri-state extractor with a fixture
  containing a `//!` fence, a column-0 `///` fence, a `text`-tagged fence
  between them, and an indented `///` fence on a trait member. Forcing
  `in_other=0` reds the fence-count and post-`text`-fence content
  assertions; reverting the indentation trim (`local trimmed="$line"`)
  reds the fence-count and fourth-fence content assertions. Both negative
  controls confirmed.
- `make doc-ratchet-selftest` and `make hew-fmt-check` both pass.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1379,6 +1379,7 @@ test-doc-examples: hew-native
 # Drive matching and mutated doc-failure sets through the production harness.
 doc-ratchet-selftest:
 	@scripts/tests/test_doc_ratchet_membership.sh
+	@scripts/tests/test_std_doc_fence_extraction.sh
 
 # Shell/python only; no artifacts.
 
--- a/Makefile
+++ b/Makefile
@@ -1360,18 +1360,19 @@ test-example-expectations-selftest:
 
 # Python only; no artifacts.
 
-# Check ```hew fenced blocks in docs/ against hew check.
-# Extracts each fence from the Markdown guides and docs/language/*.hew module
-# doc blocks into .tmp/doc-fences/, runs `hew check` on each, and applies the
-# ratchet from scripts/doc-test-expected-failures.txt so known-failing fences
-# do not block the gate while new failures always do.
+# Check ```hew fenced blocks in docs/ and std/ against hew check.
+# Extracts each fence from the Markdown guides, docs/language/*.hew module
+# doc blocks, and every std/**/*.hew doc comment into .tmp/doc-fences/, runs
+# `hew check` on each, and applies the ratchet from
+# scripts/doc-test-expected-failures.txt so known-failing fences do not block
+# the gate while new failures always do.
 #
 # Skip-annotated fences (<!-- doctest: skip --> or preceding NYI callout) are
 # never compiled — they describe aspirational or not-yet-implemented surfaces.
 # Fail-closed default: a fence is compiled unless explicitly skipped.
 #
-# Run `make test-doc-examples` after any docs/ change to confirm no fence
-# regressions were introduced.
+# Run `make test-doc-examples` after any docs/ or std/ change to confirm no
+# fence regressions were introduced.
 test-doc-examples: hew-native
 	@HEW_BIN="$(DEBUG_HEW)" scripts/corpus-ratchet.sh doc-fences
 
--- a/scripts/corpus-ratchet.sh
+++ b/scripts/corpus-ratchet.sh
@@ -938,7 +938,8 @@ hew_corpus_diagnostic() {
 
 # ── Corpus: doc-fences ────────────────────────────────────────────────────────
 #
-# Extracts every ```hew fence from the language guide and the spec into
+# Extracts every ```hew fence from the language guide, the spec, the
+# docs/language/*.hew modules, and every std/**/*.hew doc comment into
 # individual files, then type-checks each one. A fence whose preceding five
 # lines carry a "Not yet implemented" callout or a `<!-- doctest: skip -->`
 # comment is SKIPPED — spec-ahead-of-implementation is not drift when a plan
@@ -962,6 +963,10 @@ DOC_FENCE_SOURCES=(
     "docs/specs/HEW-SPEC-2026.md:spec"
 )
 DOC_FENCE_LANGUAGE_DIR="$REPO_ROOT/docs/language"
+# Overridable so a test can point extraction at a small fixture tree instead
+# of the real std/ (hundreds of fences) to exercise the tri-state extractor.
+DOC_FENCE_STD_DIR="${DOC_FENCE_STD_DIR:-$REPO_ROOT/std}"
+STD_FENCE_SOURCES=()
 
 # Substrings that, in the five lines before a ```hew fence, mark it skippable.
 DOC_FENCE_NYI_PATTERNS=("Not yet implemented" "doctest: skip" "doctest:skip")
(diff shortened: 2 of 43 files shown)


slepp:
Review of the std doc-fence gate. One required fix; everything else checks out.

## Required: the indented-`///` extractor fix has no negative control

`c8637ae` fixes a second extractor bug alongside the tri-state one:

```
-        case "$line" in
+        local trimmed="${line#"${line%%[![:space:]]*}"}"
+        case "$trimmed" in
         '//!'*)
-            line="${line#//!}"
+            line="${trimmed#//!}"
```

Its own commit message states the failure mode: item-level `///` docs on trait/impl members "silently blanked to nothing". The self-test's fixture writes every `///` at column 0:

```
+/// Item doc with an example.
+///
+/// ```
+/// let b = 2;
+/// ```
+pub fn noop() {}
```

so nothing exercises the trim. Reverting the fix in place (`local trimmed="$line"`) and running `scripts/tests/test_std_doc_fence_extraction.sh` still reports `5 passed, 0 failed`.

That is the worst gap for this gate specifically: the ratchet only fires on *new* failures, so a re-broken trim drops every trait/impl-member fence out of coverage and no gate notices. Please extend the fixture with an indented member doc, e.g.

```
pub trait Example {
    /// Member doc with an example.
    ///
    /// ```
    /// let d = 4;
    /// ```
    fn member();
}
```

bump the count assertion to 4, assert the fourth fence's content, and confirm the `trimmed="$line"` mutation turns it red before pushing.

## Non-blocking

`std/builtins.hew`: the `string_from_int` and `substring` hunks drop the `# Examples` heading in favour of the note, while `to_float` keeps both:

```
-/// # Examples
+/// Not yet implemented: the checker admits the call but cannot lower it to
+/// an executable target yet, so this snippet does not compile today.
```

Worth restoring the heading so the three read the same.

## What verified clean

- The std diff touches only `///`/`//!` lines: filtering the diff for added/removed lines that are not doc comments returns nothing, so no compiler or library code moved.
- `scripts/doc-test-expected-failures.txt` is byte-identical to `origin/main` and carries zero `std-` rows, so no std fence was ratcheted in as an expected failure.
- The five new "Not yet implemented" skips are honest. Running the current `hew check` on the `to_float`, `string_from_int` and `substring` bodies wrapped in `fn main()` reproduces `E_HIR: checker admitted a call without an executable target`, and the `P.Msg` fence reproduces the A640 projection-bound refusal verbatim. The three new ```` ```text ```` tags are log output and an ASCII diagram, not hew.
- The tri-state negative control is real: forcing `in_other=0` turns the fence count and the post-`text`-fence content assertions red, 2 of 5, as the PR body claims.
- Both `test-doc-examples` and the newly wired `doc-ratchet-selftest` sit in `ci-shard-1` (Makefile:637), which `.github/workflows/ci.yml:484` runs, so the self-test lands in the same shard as the fences it guards.

slepp:
Re-review at `aa3455ec`. The required fix landed and I reproduced its negative control. One thing still blocks the merge, and it is in the PR body rather than the diff.

## Required: the Test section is stale, and it becomes the squash commit message

The body still describes the pre-fix version of the self-test:

> containing a `//!` fence, an `///` fence, and a `text`-tagged fence between them; breaking the tri-state tracking (forcing `in_other=0`) turns 2 of its 5 assertions from PASS to FAIL, confirming the fixture actually exercises the bug the extractor guards against.

There are six assertions now, not five, and the section does not mention the indented-marker control that `aa3455ec` added — the fix this re-review was for. Merging as-is writes a description of a superseded test into the permanent squash body.

Please rewrite that bullet without the counts and naming both controls, e.g. the fixture pins a `//!` fence, a column-0 `///` fence, a `text`-tagged fence between them, and an indented `///` fence on a trait member; forcing `in_other=0` reds the fence-count and post-`text`-fence content assertions, and reverting the trim to `local trimmed="$line"` reds the fence-count and fourth-fence content assertions. Both mutations confirmed.

Same class, recommended but not required: drop the `(same 85 entries, all pre-existing guide/spec fences)` parenthetical. "unchanged from `origin/main`" is the claim; the number is decoration and drifts the same way.

## Verified clean at `aa3455ec`

- The required fix is applied as asked: the fixture gains a `pub trait Greeter` member with an indented `///` fence, the count assertion is 4, and the fourth fence's content is asserted. Unmutated the self-test reports 6 passed, 0 failed; with `scripts/corpus-ratchet.sh:1125` reverted to `local trimmed="$line"` it reports 4 passed, 2 failed — the fence count drops to 3 and the fourth-fence assertion goes red. The control is real.
- The new fixture actually ran in CI, not just locally: `Linux gates (shard 1/3)` on run `33654849832` logs `std doc-fence extraction self-test: 6 passed, 0 failed`.
- The non-blocking `# Examples` note is addressed — the heading is back on `string_from_int` and `substring`, matching `to_float`.
- No allowlist growth: `scripts/doc-test-expected-failures.txt` is untouched between the merge base and `aa3455ec`, and the head version carries zero `std-` rows. The `spec-0081` delta against `origin/main` is main moving ahead, not this branch.
- `mergeStateStatus` is CLEAN, `mergeable` MERGEABLE, and every check on run `33654849832` at `aa3455ec` passes, including all three Linux gate shards and the Windows and macOS legs.
- No session markers in the commits, the diff, or the body.

Update the Test section and I will arm the auto-merge.

PR #3225, opened 2026-09-02

slepp:
## Why

`hew tool playground-verify` failed on every runnable playground file on a clean tree with `main is defined multiple times`, even though each file defines exactly one `fn main`.

## What

The tool runs each file through the REPL's whole-file eval path (`eval_file_captured` → `eval_source_file_cli` → `build_program_with_kind_and_auto_print`), which classifies a file's item and always appends a synthetic `fn main() {}` after it. When the file's own item is `fn main`, that produces a duplicate definition.

- **session.rs**: skip the synthetic `fn main() {}` when the combined source already defines one, reusing `program_defines_main` (via `source_defines_main`) as the single authority for "does this program already define main" rather than a second hand-rolled check.
- **repl.rs**: `eval_file_captured` now parses the file up front and, when it parses cleanly and defines its own `main`, routes to the whole-file `eval_parsed_source_file_cli` path instead of the item-chunking path — mirroring `load_file_output_cli`'s existing routing, so a file that is one program is compiled as one program.
- **classify.rs**: a bare attribute line (e.g. `#[wire]`) has balanced delimiters and parses in no standalone context, so the line-buffering classifier treated it as a complete-but-invalid chunk instead of buffering for the item it decorates (`ends_with_bare_attribute`). This surfaced as a second, unrelated parse failure once the duplicate-main issue was fixed.
- **playground.rs**: `cmd_playground_verify` now fails closed — an empty manifest or a run where every entry was skipped no longer exits 0, since a gate that goes green having verified nothing is worse than one that never ran.
- **Makefile**: added `make playground-verify` and wired it into `ci-shard-1` alongside the other native example-verification gates (`playground-wasm-build`'s existing playground job targets wasm32 and can't run a native `fn main`).

## Test

- `hew-cli/src/eval/session.rs`: `item_input_defining_main_is_not_duplicated`, `remembered_main_item_is_not_duplicated_by_a_later_item` (regression for a `main` remembered from an earlier REPL item), and `item_input_not_defining_main_still_gets_synthetic_main` (negative control).
- `hew-cli/src/eval/classify.rs`: `input_completeness_buffers_bare_attribute_awaiting_its_item`.
- `hew-cli/src/playground.rs`: `empty_manifest_fails_closed_distinctly_from_all_skipped`, `all_skipped_run_fails_closed`, `a_failed_entry_fails_closed_distinctly_from_all_skipped`, and `run_with_a_pass_does_not_fail_closed` (negative control), plus `verify_entry_passes_for_a_file_defining_its_own_main` end to end through `verify_entry`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -49,6 +49,7 @@
 #   make sandbox-parity            — native hew run ↔ sandbox VM parity harness
 #   make playground-check          — manifest freshness + full hew-wasm test suite + build hew-wasm
 #   make playground-wasi-check     — focused curated manifest WASI runtime preflight
+#   make playground-verify         — native run of every runnable playground example vs. its .expected
 #   make licenses-check            — verify THIRD-PARTY-LICENSES is current (used in CI)
 #   make preflight                 — run every unconditional Linux gate, fail-fast
 #   make ci-preflight              — compatibility alias for make preflight
@@ -74,7 +75,7 @@
 #   make clean        — remove generated build and test artifacts
 # ============================================================================
 
-.PHONY: all build bootstrap install-hooks help shell-script-lint test-install-version-resolution actionlint hew hew-debug hew-profile-check hew-native shared-host-debug hew-lsp observe observe-functional-test mqtt-broker-e2e libhew-link-race-test runtime stdlib wasm-runtime wasm wasm-capability wasm-capability-check playground-manifest playground-manifest-check sandbox-fixtures sandbox-fixtures-check sandbox-vm-deps sandbox-parity playground-check playground-wasi-check preflight ci-preflight ci-preflight-smoke ci-local-linux wasm-dist release licenses licenses-check baselines baselines-check
+.PHONY: all build bootstrap install-hooks help shell-script-lint test-install-version-resolution actionlint hew hew-debug hew-profile-check hew-native shared-host-debug hew-lsp observe observe-functional-test mqtt-broker-e2e libhew-link-race-test runtime stdlib wasm-runtime wasm wasm-capability wasm-capability-check playground-manifest playground-manifest-check sandbox-fixtures sandbox-fixtures-check sandbox-vm-deps sandbox-parity playground-check playground-wasi-check playground-verify preflight ci-preflight ci-preflight-smoke ci-local-linux wasm-dist release licenses licenses-check baselines baselines-check
 .PHONY: test test-strict ratchet-accounting ratchet-accounting-nextest test-ratchet-accounting-runner macos-leak-oracle test-leak-oracle-selftest test-cabi test-compiler-pipeline test-compiler-lifecycle test-opaque-resource-lifecycle-matrix test-opaque-resource-lifecycle-matrix-external test-vertical-slice test-pkg-import test-package-install test-runtime-unit test-hew-ratchet test-core-matrix core-matrix-record funcupdate-mir-baselines-golden test-o2-differential o2-differential-selftest test-stdlib-ratchet test-ux-examples ux-examples-expect test-surface-examples surface-examples-expect test-example-expectations-selftest test-release-binary test-release-lib-link asan asan-fixtures test-asan-fixture-selftest tsan miri lint structural-lint structural-lint-bootstrap structural-lint-bootstrap-install test-ast-grep-contract stdlib-lint stdlib-errno-gate legacy-path-syntax-lint hew-fmt-check test-migrate-corpus doc-ratchet-selftest verify-sys-lane-closure test-sys-lane-closure hew-fmt-property test-build-harness forced-cancel-composite-check
 .PHONY: test-ownership-balance-corpus test-ownership-balance-runner-selftest
 .PHONY: stdlib-user-build-clean
@@ -627,6 +628,12 @@ playground-wasi-check: wasm-runtime hew-native
 	$(TEST_RUN_ENV) cargo test -p hew-cli --test wasi_run_e2e curated_playground_examples_run_under_wasi -- --exact
 	$(TEST_RUN_ENV) cargo test -p hew-cli --test wasi_run_e2e supervisor_stays_on_the_unsupported_diagnostic_path_under_wasi -- --exact
 
+# Native run of every runnable playground example against its checked-in
+# `.expected` file (`hew tool playground-verify`), catching drift the
+# analysis-only WASM/manifest checks above don't exercise.
+playground-verify: hew-native
+	$(DEBUG_HEW) tool playground-verify
+
 # Standard per-branch gate: validate workflow syntax locally, then run the lint
 # graph and the same three Make-owned test groups used by hosted Linux CI. One
 # Make graph lets shared prerequisites build once instead of being replanned by
@@ -644,7 +651,7 @@ ci-preflight: preflight
 ci-shard-1: observe-functional-test test-cabi test-compiler-lifecycle \
 	test-vertical-slice test-pkg-import test-runtime-unit test-ux-examples \
 	test-doc-examples doc-ratchet-selftest test-migrate-corpus \
-	o2-differential-selftest
+	o2-differential-selftest playground-verify
 
 ci-shard-2: hew-profile-check libhew-link-race-test test \
 	test-leak-oracle-selftest test-opaque-resource-lifecycle-matrix-external \
--- a/hew-cli/src/eval/classify.rs
+++ b/hew-cli/src/eval/classify.rs
@@ -83,11 +83,32 @@ pub fn input_completeness(input: &str) -> InputCompleteness {
         return InputCompleteness::Complete;
     }
 
-    if has_unclosed_delimiters(trimmed) || parses_with_continuation(trimmed) {
+    if has_unclosed_delimiters(trimmed)
+        || parses_with_continuation(trimmed)
+        || ends_with_bare_attribute(trimmed)
+    {
         return InputCompleteness::Incomplete;
     }
 
     InputCompleteness::Invalid
+}
+
+/// True if `input` is a syntactically complete attribute list (`#[wire]`,
+/// stacked `#[a]\n#[b]`, …) with no item yet attached to it.
+///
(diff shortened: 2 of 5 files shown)


slepp:
First review at `7c6b77fa`. The gate is a real addition and it runs — but the branch does not merge as it stands, and the fix is placed in the shared REPL builder rather than at the routing seam that already owns this decision, which leaves the same duplicate-`main` failure live on a second path and regresses the interactive REPL.

## Required: rebase

`mergeStateStatus` is `DIRTY` and `mergeable` is `CONFLICTING` at `7c6b77fa`. Rebase onto `origin/main` (`e24a447d4`) before anything else; CI reruns from there.

## Required: name the root cause, and fix it at the routing seam

The body says the tool "runs each file through the REPL's whole-file eval path … and always appends a synthetic `fn main() {}`". That is the symptom. The cause is one layer up: `eval_file_captured` (`hew-cli/src/eval/repl.rs:1796`) goes straight to `eval_source_file_cli`, bypassing the whole-file routing that already exists for the same job on `:load` (`repl.rs:1124-1133`):

```rust
fn load_file_output_cli(&mut self, source: &str, path: &str) -> Result<String, LoadFileError> {
    let parse_result = hew_parser::parse(source);
    if parse_result.errors.is_empty() && program_defines_main(&parse_result.program) {
        return self
            .eval_parsed_source_file_cli(parse_result.program, source, path)
            .map_err(load_file_error_from_cli);
    }
    self.eval_source_file_cli(source, path, path)
```

A file that parses and defines its own `main` is already meant to be compiled as one program, not chunked line-by-line into REPL inputs. `hew tool playground-verify` never gets that routing. Meanwhile the PR adds a second predicate for the same fact:

```rust
+fn defines_main(input: &str) -> bool {
+    let parse_result = hew_parser::parse(input);
+    if !parse_result.errors.is_empty() {
+        return false;
+    }
+    parse_result
+        .program
+        .items
+        .iter()
+        .any(|(item, _)| matches!(item, Item::Function(decl) if decl.name == "main"))
+}
```

`program_defines_main` at `repl.rs:376` is that function, already. Two authorities for one fact, and the new one lives in the builder every REPL input passes through in order to fix a playground-only caller.

### Why the chunk-local check is not enough

`build_program_with_kind_and_auto_print` replays prior session items above the input:

```rust
     for item in &self.items {
         source.push_str(&item.source);
         source.push('\n');
     }
```

and the new guard only inspects the current chunk:

```rust
         source.push_str(input);
         source.push('\n');
-        source.push_str("fn main() {}\n");
+        if !defines_main(input) {
+            source.push_str("fn main() {}\n");
+        }
```

`eval_source_file_cli` chunks a file one complete item at a time and `eval_file_cli` records each successful `InputKind::Item` into the session (`repl.rs:967` → `record_success` → `session.add_item`). So for

```hew
fn main() { println("hi"); }
fn helper() -> i32 { 1 }
```

chunk one defines `main`, runs, and is recorded; chunk two replays that real `main` from `self.items`, `defines_main("fn helper…")` is false, the synthetic `fn main() {}` is appended, and the program is back to `main is defined multiple times`. Same symptom, one file shape away.

That shape is in the corpus — `examples/lambda_actors.hew` and `examples/lambda_actor_factory.hew` both carry top-level `fn`s after `main`. None of the 44 curated manifest entries do today, which is why the new gate is green; it is one curated example away from red, and catching exactly that drift is the gate's purpose.

The same replay makes this a regression in shipped REPL surface, not only in the tool. Before this change, `fn main() { println("hi"); }` typed at the REPL errored on the duplicate and was never recorded. After it, that input succeeds and is recorded, and the next item definition fails with a duplicate `main` until `:clear`.

3237's `already_has_main || self.items.iter().any(item_defines_main)` does not resolve this either: with the second half, chunk two compiles the replayed real `main` alongside `helper` and runs `main` a second time, printing `hi` again. Neither chunk-level variant is the fix.

Please route instead: have `eval_file_captured` reuse (or mirror) `load_file_output_cli`'s parse-once-then-`program_defines_main` decision so a whole file with its own `main` is compiled as one program, and drop `defines_main` from `session.rs`. If you keep a session-level guard as REPL surface in its own right, it needs the replay case tested, not just the single-chunk case. Either way the test that should exist is a whole-file eval of `fn main() { println("hi"); }` followed by a second top-level item, asserting one run of `main` and no duplicate-definition error — the current `item_input_defining_main_is_not_duplicated` builds one chunk against an empty session and cannot see this.

Worth checking once routing lands: `types/wire_types` is why `ends_with_bare_attribute` was needed, and a whole-file compile never line-classifies. The classifier fix stands on its own for interactive `#[wire]` input — its test is a good one — but confirm it is still load-bearing for the tool, and say which it is.

## Recommended, not blocking

`verify_entry_passes_for_a_file_defining_its_own_main` returns green when its probe fails:

```rust
+        if !require_toolchain() {
+            return;
+        }
```

A test that can pass without executing its assertion reports coverage it may not have. The precedent in `eval::repl::tests::require_toolchain` is real and `make playground-verify` covers the behaviour in CI regardless, so this is not a blocker — but a probe failure should be visible rather than silent.

## Verified clean at `7c6b77fa`

- `make playground-verify` is wired into `ci-shard-1` and genuinely ran: `Linux gates (shard 1/3)` on run `33653234091` logs `target/debug/hew tool playground-verify` and `playground verify: 37 passed, 0 failed, 7 skipped`, the seven skips being pre-existing `capabilities.wasi != "runnable"` manifest entries.
- No allowlist and no expected-failure file touched.
- `item_input_not_defining_main_still_gets_synthetic_main` is a real negative control for the guard as written, and `input_completeness_buffers_bare_attribute_awaiting_its_item` pins both the buffering and the completion case.
- Body is Why/What/Test with no counts; no session markers in the commits, the diff, or the description.
- Every check at `7c6b77fa` passes.

slepp:
Review of `fix/playground-verify-item-defines-main` at `7c6b77f`. The root cause is named correctly and the gate is real: `Linux gates (shard 1/3)` on run `33653234091` logs `playground verify: 37 passed, 0 failed, 7 skipped`, so `make playground-verify` genuinely executed in CI rather than only locally. Three things block the merge.

### 1. `defines_main` only looks at the new input, not at the session it is appended to

```diff
+                if !defines_main(input) {
                     source.push_str("fn main() {}\n");
+                }
```

but the source this decides for is built from `self.items` first:

```rust
        for item in &self.items {
            source.push_str(&item.source);
            source.push('\n');
        }
```

and `record_success` (`repl.rs:1297`) remembers every successful `InputKind::Item`. Before this change, `fn main() { ... }` typed at the REPL failed with `main is defined multiple times` and was therefore never recorded. After it, that eval succeeds and the item is remembered, so the next item eval emits remembered-`main` + the new item + the synthetic `fn main() {}` and hits the same duplicate the PR is fixing. The predicate has to consider the remembered items as well as the input.

The covering test is the second half the current pair is missing: `add_item("fn main() { println(\"hi\"); }")`, then `build_program("fn foo() -> i32 { 42 }")`, asserting exactly one `fn main` in `prog.source`. `item_input_not_defining_main_still_gets_synthetic_main` is a good negative control for the input clause and does not reach this one.

While you are there: `program_defines_main` already exists at `hew-cli/src/eval/repl.rs:376` and does exactly what the new `defines_main` does, one parse later.

```rust
+fn defines_main(input: &str) -> bool {
+    let parse_result = hew_parser::parse(input);
```

Two predicates for one fact is the thing to avoid. Lift the existing one to somewhere both modules can call and delete the copy.

### 2. The new gate passes when nothing was verified

`cmd_playground_verify` exits non-zero only on a failure count:

```rust
    if entries.is_empty() {
        eprintln!("Warning: manifest contains no entries.");
        return;
    }
...
    if failed > 0 {
        std::process::exit(1);
    }
```

An empty manifest returns 0, and so does a run where every entry took the `capabilities.wasi != "runnable"` skip. That was tolerable for a hand-run tool. As of

```diff
-	o2-differential-selftest
+	o2-differential-selftest playground-verify
```

it is a CI gate, and the day a manifest regeneration or a capabilities flip empties the runnable set, the gate goes green having checked nothing. Please make a run with no `Pass` exit non-zero, and keep the failing message specific enough to distinguish "no entries" from "all skipped".

### 3. The branch is behind main

`mergeStateStatus` is `DIRTY`, `mergeable` is `CONFLICTING`. The one conflict is the `.PHONY` line, where main has since added `test-install-version-resolution`:

```
<<<<<<< .our
.PHONY: all build bootstrap install-hooks help shell-script-lint test-install-version-resolution actionlint ...
=======
.PHONY: all build bootstrap install-hooks help shell-script-lint actionlint ... playground-check playground-wasi-check playground-verify ...
>>>>>>> .their
```

Rebase onto `origin/main` keeping both additions.

### Not blocking

- `playground::tests::require_toolchain` is a verbatim copy of `eval::repl::tests::require_toolchain`. The doc comment says so, which is the honest version, but a shared test helper would be better than a second copy of the probe.
- `ci-shard-1` now does a native compile, link and run for every runnable playground entry. Shard 1 was the shortest of the three at about 20 minutes; worth watching that this does not make it the long pole.

### Verified clean at `7c6b77f`

- The root cause is the real one: `build_program_with_kind_and_auto_print`'s `InputKind::Item` arm unconditionally appended `fn main() {}`, and `eval_file_captured` sends whole files through it.
- The `#[wire]` buffering fix is in scope, not drive-by: it is the second failure the tool hits once the duplicate main is gone, and `input_completeness_buffers_bare_attribute_awaiting_its_item` pins both the buffering and the completion once the item is appended.
- Both new tests actually ran in CI rather than skipping past the toolchain probe: `verify_entry_passes_for_a_file_defining_its_own_main` reports 0.221s on `Linux gates (shard 2/3)`, alongside 0.254s to 0.390s for the existing `eval_file_multiline` / `eval_one_expression` compiles, so the probe resolved and the test did the work.
- No allowlist, no expected-failure file touched, and every check on run `33653234091` passes.
- No session markers in the commits, the diff or the body, and the description carries no counts.

PR #3234, opened 2026-09-02
fix(runtime): publish an actor's crash code before its terminal state

### LINKED ISSUE #3222: Linked crash exit reason races on Windows CI
## Problem

`hew-runtime::links_monitors_integration::test_link_after_crash_delivers_exit_without_stale_registration`
failed on Windows CI on PR #3216 (run 33603873366, job
https://github.com/hew-lang/hew/actions/runs/33603873366/job/100163813726):

```
thread 'test_link_after_crash_delivers_exit_without_stale_registration' panicked at hew-runtime\tests\links_monitors_integration.rs:285:5:
assertion `left == right` failed
  left: Some(ExitMessageView { crashed_actor_id: 2, reason: -1 })
 right: Some(ExitMessageView { crashed_actor_id: 2, reason: 0 })
```

The same test passed on the same commit depth on Linux, macOS, and Windows in
main's latest run (33602799686). PR #3216 touches only
`docs/specs/HEW-SPEC-2026.md`, `docs/syntax-data.json`, `editors/*`, a lexer
test assertion, and a spec-fence checksum — nothing in `hew-runtime`. There is
no code difference between the passing and failing runs that could explain
this, so the failure is a genuine ordering race in the runtime/test, exposed
by Windows CI's scheduler timing.

## Expected

`hew_actor_get_error` read after `wait_for_state(Crashed)` returns should
observe the same `reason` the crashed actor's own EXIT propagation delivers to
its links — deterministically, on every platform.

## Notes

Reading `hew-runtime/tests/links_monitors_integration.rs:253-291` and the
crash-teardown path in `hew-runtime/src/actor.rs` (around line 6830 onward):

The test does, in order:
1. `hew_fault_inject_crash` + `send_empty` to crash `target`.
2. `target.wait_for_state(HewActorState::Crashed, ...)` — polls
   `actor.actor_state` (a separate `AtomicI32`) with `Ordering::Acquire` at
   10ms intervals (`hew-runtime-testkit/src/lib.rs:390-411`).
3. Once that returns true, `hew_actor_get_error(target.as_ptr())` — reads
   `actor.error_code` (`hew-runtime/src/actor.rs:6980-6984`,
   `Ordering::Acquire`) — into `exit_reason`, which becomes the *expected*
   `reason` in the later `assert_eq!` against the EXIT message actually
   delivered to the newly-linked `survivor`.

On the writer side, the crash-teardown function (`actor.rs`, terminal-state
CAS around line ~6820-6830) does, in order:
1. `compare_exchange(current, terminal, Ordering::AcqRel, Ordering::Acquire)`
   on `actor_state` — this is what `wait_for_state` observes.
2. `fault_close_registered_gen_sink`, mailbox close, mailbox drain
   (`mailbox_reclaim_queued_terminal[_if]`), crash-teardown hooks,
   `link::record_link_probe_terminal` — real work, non-trivial wall-clock
   time.
3. Only then: `a.error_code.store(error_code, Ordering::Release)`
   (`actor.rs:6886`).
4. `crate::link::propagate_exit_to_links(actor_id, error_code)`
   (`actor.rs:6910`) — delivers the EXIT using the writer's own local
   `error_code` variable, not a re-read of the atomic.

`actor_state` and `error_code` are independent atomics with no synchronizing
relationship between them beyond program order on the writer thread. A reader
that observes `actor_state == Crashed` via its own `Acquire` load on
`actor_state` gets no happens-before guarantee about the *later*
`error_code.store` — it only synchronizes-with whatever was released at the
CAS in step 1. If the polling test thread's 10ms-interval wakeup lands between
steps 1 and 3, `hew_actor_get_error` returns the field's still-live default
(`AtomicI32::new(0)`, "no error") instead of the real crash code, while the
EXIT message the link actually receives moments later (step 4) carries the
correct, non-zero `error_code` the writer held locally the whole time. That
matches the observed failure exactly: `left.reason == -1` (real crash code,
delivered via the link) vs `right.reason == 0` (stale default, read by the
test before the writer's store executed).

This window is apparently wide enough, or Windows CI's thread scheduling
coarse enough, for the test's 10ms poll to occasionally win the race there but
not (yet observed) on Linux/macOS. Either the test needs a synchronization
point that only proceeds once `error_code` is actually published (not just
`actor_state`), or the runtime should publish `error_code` before/together
with the `actor_state` CAS so any observer of `Crashed` is guaranteed a
consistent `error_code` read.

slepp:
## Why

`test_link_after_crash_delivers_exit_without_stale_registration` failed on Windows CI reading `reason: 0` where the EXIT delivered to the link carried `reason: -1`, on a commit that touched no runtime code. The crash teardown claimed the terminal state first and stored the crash code much later, after the gen-sink fault-close, the mailbox close, and the queued-terminal reclaim. `actor_state` and `error_code` are independent atomics, so an acquire-observer of `Crashed` got no guarantee about a store that had not happened yet and read the field's `0` default.

That pair is not just the test's reading order - it is the runtime's. `link::terminal_exit_reason` (link.rs:259), the monitor terminal-reason lookup (monitor.rs:898), and `hew_actor_await` (actor_group.rs:320) all load `actor_state` and then `error_code` for a link, monitor, or waiter that arrives after the crash. Each of them could report "no error" for a crashed actor.

## What

- **runtime**: `hew_actor_trap_inner` stores `error_code` before the terminal CAS instead of after the teardown. The CAS's release then publishes the code to every acquire-observer of the terminal state, so `hew_actor_get_error` and the EXIT propagation agree by construction. Publishing early is not a new condition for the field: a dispatch that trapped through `hew_panic` has already stamped its code there before the trap runs.
- The WASM twin already had this order - `scheduler_wasm.rs` reads the stamped code and only then stores `Crashed` - so no parity change was needed.
- A trap that loses the CAS now also stores its code, where before it stored nothing. Only one trap per actor is reachable from an activation and the external caller (the failed-DOWN trap in `monitor.rs`) passes a fixed code, so the codes agree today; the comment records the condition that ends that and names packing state and code into one claimed atomic as the real fix.

## Test

`a_crash_code_is_published_before_the_terminal_state_is_visible` (hew-runtime/tests/links_monitors_integration.rs) reads the error through `hew_actor_get_error` from the crash-teardown first-wake hook - the first instant the terminal state is visible to another thread - and asserts it is already the injected crash code. It fails deterministically without the fix with the issue's exact values:

```
assertion `left == right` failed: the crash code must already be readable at the first point the terminal state is visible
  left: 0
 right: -1
```

It also asserts the actor reports `0` while running, so the post-fix assertion cannot be satisfied by a field that is simply always set, and re-checks the same reason against the EXIT delivered to a link registered after the crash.

`native_self_stop_then_crash_publishes_crashed_and_notifies_supervisor` (actor.rs) had a comment documenting the old ordering as intentional and polled state and error together to work around it. It now waits on the state alone and reads the error in a separate step, which is the invariant rather than an accommodation of it.

Fixes #3222


slepp:
Review: NO-GO for now. The ordering fix and its memory ordering are right, the new test is a real deterministic reproduction, and CI is fully green. Two things block the merge: the branch is CONFLICTING against main, and moving the store ahead of the loop changes behaviour for a case the new comment does not cover.

### 1. Rebase — LESSONS.md only

`gh pr view` reports `mergeStateStatus: DIRTY` / `mergeable: CONFLICTING`. `git merge-tree --write-tree origin/main FETCH_HEAD` names exactly one conflicted path:

```
100644 554c162b6d8dfb736e5bcad0078ae88c7a3ceae3 1  LESSONS.md
100644 a9fb02934ec61ee7332224f4b88592d3e3f1072e 2  LESSONS.md
100644 60107c5510657414dd8457b2427617ff576989ef 3  LESSONS.md
```

Both sides appended a new section at the end. `hew-runtime/src/actor.rs` and the integration test merge clean.

### 2. The store now also lands on an actor that is already terminal

```diff
+    a.error_code.store(error_code, Ordering::Release);
     loop {
         let current = a.actor_state.load(Ordering::Acquire);
         if current == HewActorState::Stopped as i32 || current == HewActorState::Crashed as i32 {
             return;
         }
```

Before this change the store sat after the CAS, so the early `return;` on an already-settled actor skipped it. Now the store happens first and the early return leaves it in place, so a late trap rewrites the exit reason of an actor that already has one. Two reachable shapes:

- A stop-trap (`error_code == 0`, the graceful-terminal path this function accepts) arriving on an already-`Crashed` actor zeroes the crash code. `link::terminal_exit_reason` (link.rs:259) and `hew_actor_await` (actor_group.rs:329) then hand a late link or waiter reason `0` for a crashed actor - the same symptom as #3222, from the other direction, and a dropped error rather than a stale one.
- `monitor.rs:845` traps the *monitoring* actor with `HEW_TRAP_ACTOR_SEND_FAILED` when the DOWN send fails, which is the mailbox-closed case, so that actor is usually already terminal and often cleanly `Stopped` with `error_code == 0`. Both readers above return `error_code` for `Stopped` without checking for `Crashed`, so a clean stop starts reporting a send-failure reason.

The new comment's "WHY this is a store and not a claim" paragraph reasons only about two traps racing the CAS. The settled-terminal case is the common one and is not covered.

Suggested fix, keeping the publication guarantee intact - move the store inside the loop, after the terminal check and before the exchange:

```rust
loop {
    let current = a.actor_state.load(Ordering::Acquire);
    if current == HewActorState::Stopped as i32 || current == HewActorState::Crashed as i32 {
        return;
    }
    a.error_code.store(error_code, Ordering::Release);
    if a.actor_state
        .compare_exchange(current, terminal, Ordering::AcqRel, Ordering::Acquire)
        .is_ok()
    {
        break;
    }
}
```

The store still precedes the CAS that publishes the terminal state, so an acquire-observer of `Crashed` still sees the code and #3222 stays fixed. The only remaining window is a store followed by a lost CAS, which is exactly the racing-writers case the comment already documents; please have that comment name the settled-terminal case as the reason the store sits below the early return.

### 3. A negative test for the case above

`a_crash_code_is_published_before_the_terminal_state_is_visible` cannot catch this - it only ever traps a live actor. Please add one in the same fixture: trap an already-terminal actor with a different code (a `Stopped` actor with a non-zero code, or a `Crashed` actor with `0`), assert `hew_actor_get_error` is unchanged, and assert a link registered afterwards receives the original reason.

### What checks out

- `Ordering::Release` on the store, `AcqRel` on the terminal CAS, `Acquire` on both loads in all three readers (link.rs:259, monitor.rs:898, actor_group.rs:329). The pairing is correct.
- The new test is deterministic rather than timing-based: the hook runs on the trapping thread right after the CAS, the `HOOK_NEVER_FIRED` sentinel keeps it from passing vacuously, the running-actor `assert_eq!(error_while_running, 0)` rules out an always-set field, and `FirstWakeHookGuard` clears the hook before the actor handle drops.
- `test_link_after_crash_delivers_exit_without_stale_registration` is untouched.
- The WASM-twin claim holds: `scheduler_wasm.rs:2161` loads the stamped code and only stores `Crashed` at 2177.
- Every check on the PR passes, including the Windows leg the issue came from.
- The actor.rs test change is the right direction: waiting on the state alone and reading the error separately tests the invariant instead of accommodating it.

slepp:
Merge evidence: every Linux, Windows, clippy, licence and compiled-suite check passes at 73e61a2. The only red leg is macOS arm64, whose single unexpected ID is hew-cli::dwarf_debugger_locals_e2e::debugger_renders_only_active_enum_variant_payload, already tracked as #3236 (lldb-side nondeterminism, also seen on PR 3220 which touched no codegen). This PR touches hew-runtime actor state and its integration test only, so the ID is outside its blast radius; the remaining macOS failures are in main's reference set. Merging under the ID-diff protocol.

PR #3235, opened 2026-09-02

slepp:
## Why

`hew tool playground-verify` reported 0 passed / 37 failed against a clean
tree, every entry failing with "`main` is defined multiple times" even
though each source file defines exactly one `fn main`.

## What

- **eval session**: `hew-cli/src/eval/session.rs` chunks a whole file through
  the REPL evaluator and always appended a synthetic `fn main() {}` after any
  top-level item chunk. When the chunk was itself the file's own `fn main`,
  that produced a real duplicate. Skip the synthetic driver when the item
  chunk already declares `main`.
- **eval classify**: fixing the above surfaced a second, independent bug —
  `hew-cli/src/eval/classify.rs`'s completeness check saw a standalone
  attribute line (`#[wire]`) as syntactically balanced and evaluated it on
  its own, producing "expected expression, found `#[`". Buffer one or more
  bare attribute lines until the attributed item follows.
- **CI**: added `make playground-verify` beside `playground-manifest-check`
  and wired it into `ci-shard-3`, next to `sandbox-parity`, which already
  builds the native `hew` driver — the added step only costs a native
  link+run per curated example.

## Test

- `hew-cli/src/eval/session.rs`: unit tests for an item chunk that declares
  its own `main` (no synthetic driver), one that does not (driver still
  appended — negative control), and an accumulated helper item followed by
  its own `main` (the multi-chunk shape).
- `hew-cli/src/eval/classify.rs`: unit tests that a bare attribute line
  buffers as incomplete, stacked attributes buffer, the attributed item
  completes the buffer, and a balanced-bracket non-attribute line that is
  otherwise malformed still reports `Invalid` (negative control).
- `hew-cli/tests/playground_verify_e2e.rs` (new): spawns the real built
  `hew` binary against a temporary manifest — a source file with its own
  `main` verifies, one with a helper and a `#[wire]`-attributed type before
  its own `main` verifies, and a source file with a genuinely duplicated
  `main` still fails with the real "defined multiple times" diagnostic
  (negative control proving the fix does not suppress real duplicate-main
  detection).
- `make playground-verify` now passes 37/0/7 (passed/failed/skipped)
  against the real curated manifest.

### Notes

Reproducing the native-link integration test in-process (calling
`verify_entry` directly from a `#[cfg(test)]` unit test) hit a separate,
pre-existing gap: `hew-cli/src/link.rs`'s `libhew.a` search-path candidates
assume the running exe sits at `target/<profile>/<exe>`, but a
`cargo test`/`nextest` unit-test binary for a bin crate sits one level
deeper at `target/<profile>/deps/<exe>`, so every candidate resolves one
`target/` segment too deep. Every native-compile REPL test in
`hew-cli/src/eval/repl.rs` is gated behind a `require_toolchain()` probe
that hits the same gap and silently skips under `cargo nextest run` even
with `libhew.a` already built. Left untouched here — out of scope for this
fix — and noted in `.tmp/TODO.md`; the real end-to-end proof was moved to a
process-spawning integration test instead, which does not sit at the
affected depth.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -49,6 +49,7 @@
 #   make sandbox-parity            — native hew run ↔ sandbox VM parity harness
 #   make playground-check          — manifest freshness + full hew-wasm test suite + build hew-wasm
 #   make playground-wasi-check     — focused curated manifest WASI runtime preflight
+#   make playground-verify         — focused curated manifest native eval preflight
 #   make licenses-check            — verify THIRD-PARTY-LICENSES is current (used in CI)
 #   make preflight                 — run every unconditional Linux gate, fail-fast
 #   make ci-preflight              — compatibility alias for make preflight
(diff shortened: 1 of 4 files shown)


slepp:
Duplicate of #3225 (the same fix opened three times by interrupted lanes); #3225 is the lane of record.

PR #3237, opened 2026-09-02

slepp:
## Why

`hew tool playground-verify` failed on every runnable curated example
on a clean main tree, with `main is defined multiple times` for
playground files that define exactly one `fn main` (e.g.
`examples/playground/types/record_clone.hew`). Nothing in the build
graph or CI ever ran this tool, so the regression went unnoticed.

## What

- **cli/eval**: `eval_file_captured` (the machinery behind
  `playground-verify` and `hew eval -f`) reuses the REPL's fragment
  builder, which always appended a synthetic `fn main() {}` after any
  new top-level item. A whole file that already defines its own `main`
  got a second one appended and failed to compile. Skip the synthetic
  main when the item, or a prior session item, already defines one.
- **cli/eval**: a second, related bug surfaced once files with their
  own `main` could compile: the file chunker flushed a bare attribute
  line (e.g. `#[wire]`) on its own instead of buffering it with the
  item it decorates, because none of the classifier's continuation
  probes was item-shaped. Add one that is.
- **build/ci**: add a `playground-verify` make target and run it as
  part of `playground-check`, wired into the `playground-wasm-build`
  CI job (which now also installs LLVM, since building the native
  `hew` binary requires it).

## Test

- `hew-cli/src/eval/session.rs`: `item_input_with_own_main_is_not_duplicated`
  (own-main item is not duplicated) and
  `item_input_without_main_still_gets_synthetic_main` (negative control:
  an item without `main` still gets the wrapper).
- `hew-cli/src/eval/classify.rs`: `bare_attribute_line_is_incomplete`
  and negative control `genuinely_invalid_input_stays_invalid`.
- `hew-cli/src/playground.rs`: `playground_file_with_own_main_verifies`,
  `playground_file_without_main_still_gets_wrapper`, and negative
  control `playground_file_with_genuinely_duplicate_main_fails` (a file
  that genuinely defines `main` twice must still fail). These three
  compile and run real programs, so they skip cleanly (mirroring
  `eval::repl::tests::require_toolchain`) in an environment where
  `libhew.a` can't be resolved from the test binary's own directory.
- `make playground-verify` passes locally against every runnable
  curated example.


slepp:
Note for reviewers / follow-up: the three `verify_entry` tests
(`playground_file_with_own_main_verifies`,
`playground_file_without_main_still_gets_wrapper`,
`playground_file_with_genuinely_duplicate_main_fails`) skip locally
because the `cargo test` binary lives under `target/debug/deps/`, a
directory name `hew_lib_candidates` (in `hew-cli/src/link.rs`) does not
recognize as a Cargo profile, so `find_hew_lib` can't resolve the
prebuilt `libhew.a` from there. This is a pre-existing gap, the same
one `eval::repl::tests::require_toolchain` already guards against; not
something to fix in this PR. The real gate for this fix is the
`playground-verify` make target added here, run as a step in the
`playground-wasm-build` CI job — that job builds and runs the actual
`target/debug/hew` binary, whose `exe_dir` name (`debug`) the profile
detection does recognize.

Whether CI's shard layout resolves the `libhew.a` lookup for the
in-process tests is visible in the job logs: grep for `playground
verify_entry tests skipped`. Will report back once the run lands.

slepp:
Review of a1ed927: two required fixes, plus notes.

## Required: the branch conflicts with main

`mergeStateStatus` is `DIRTY` (`mergeable: CONFLICTING`) against base
7a24874. Rebase onto current main before this can be queued; LESSONS.md,
the Makefile `.PHONY` line, and `.github/workflows/ci.yml` are all
high-churn files this branch touches.

## Required: the `self.items` half of `already_has_main` is not covered anywhere

`hew-cli/src/eval/session.rs`:

```rust
                let already_has_main = item_defines_main(input)
                    || self
                        .items
                        .iter()
                        .any(|item| item_defines_main(&item.source));
```

The comment above it says the second operand is load-bearing for a file whose
`main` is not the last top-level item. Nothing exercises it:

- `item_input_with_own_main_is_not_duplicated` and
  `item_input_without_main_still_gets_synthetic_main` build on a fresh
  `Session::new()`, so `self.items` is empty and the first operand decides.
- `playground_file_with_own_main_verifies` chunks `#[wire] type ...` and then
  `fn main`, so the `main` chunk hits the first operand.
- `playground_file_with_genuinely_duplicate_main_fails` short-circuits on the
  first operand too.
- The new `playground-verify` target does not reach it either. All 37 runnable
  manifest entries declare `fn main`, and in every one of them `main` is the
  last top-level item, so no chunk after a recorded `main` is ever built.

Either drop the clause or test it. A session-level unit test is enough:

```rust
let mut session = Session::new();
session.add_item("fn main() { println(\"a\"); }");
let prog = session.build_program("fn helper() -> i32 { 42 }");
assert_eq!(prog.source.matches("fn main").count(), 1);
```

## Note: main-not-last files now fail quietly instead of loudly

`add_item` records `fn main` into `self.items`, and
`build_program_with_kind_and_auto_print` replays every prior item ahead of the
new chunk. For a file whose `main` precedes another item, chunk 1 runs `main`
and every later chunk replays and re-runs it, so `hew eval -f` returns that
output repeated. Before this branch the same file failed with "main is defined
multiple times". `playground-verify` would catch it as an output mismatch;
`hew eval -f` just returns a wrong string. Worth an issue at merge time.

Related: `InputKind::Statement` and `InputKind::Expression` still emit an
unconditional `fn main()`, so a file that mixes its own `main` with a
top-level statement still collides. No curated example has that shape.

## Note: the new native gate skips seven examples on a WASI capability

`verify_entry` returns `Skipped` when `capabilities.wasi != "runnable"`, but it
compiles and runs natively, not under WASI. That excludes
`concurrency/{actor_pipeline,async_await,counter_actor,supervisor}`,
`machines/traffic_light`, `language/string_slicing`, and `types/method_clone`.
All seven run natively, and all are multi-item files of the shape the chunker
bugs here affected. Pre-existing in the tool, but this PR is what puts the gate
in CI, so the blind spot is now worth an issue.

## Note on the CI wiring

`playground-wasm-build` is `timeout-minutes: 20` and now builds the native
compiler, and unlike every other LLVM-linking job it installs no
`libssl-dev pkg-config zlib1g-dev libzstd-dev` (ci.yml lines 173, 222, 426).
If the job goes red on the timeout or on an llvm-sys link, those become
required fixes as well. `release-gate.yml` already provisions LLVM before its
`make playground-check`, so that leg is fine.

The rest holds up: the target is wired through `playground-check` into a job
that runs it, no allowlist was added, `examples/playground/manifest.json` is
untouched, and each fix carries a negative control.

slepp:
Duplicate of #3225 (the same fix opened three times by interrupted lanes); #3225 is the lane of record.

PR #3246, opened 2026-09-02
Close a composite's resource leaf exactly once

### LINKED ISSUE #3070: Close a composite's #[resource] field exactly once
## Problem

A composite that carries a `#[resource]`-kind field — a record, tuple or enum
payload holding a file, socket, child process or other affine handle — can have
that handle closed twice: once through the composite's own drop recipe as it
walks its fields, and once through the resource's close path. Resource kinds
are registered through `TypeRegistry::is_resource`
(`hew-types/src/traits.rs:592`) and the affine walls are in
`hew-types/src/check/admissibility.rs`; the composite-side release is emitted
by MIR drop elaboration, and the two do not agree on who owns the close.

There is no minimal `.hew` reduction yet. Producing one is the first task here.

## Expected

Exactly one close per resource handle on every exit edge, as the
leak-never-double-free order in `docs/v05/ownership.md` requires: a composite
holding a resource either transfers the close obligation to its field walk or
keeps it, never both.

## Notes

D279 places this in its own lane rather than behind the leak-fix branches, and
it gates the machine desugar (Q341) because a machine's state payload is a
composite that can carry a handle.

- `hew-types/src/traits.rs:592` (`is_resource`),
  `hew-types/src/check/admissibility.rs:1071,1267,1512`
- `hew-mir/src/lower/composite_own.rs`, `hew-mir/src/drop_obligation.rs`
- `docs/specs/handle-safety-and-resource-lifetime.md`, `docs/v05/ownership.md`
- Related but distinct: #3068 (WASI encoding `Value` teardown recurses into
  `close` until it traps)
- Absorbs #3113 (its minimal reduction).

## Outcome

A `#[resource]` value or a resource-kind runtime handle (Sink, Stream) reached
through a record, tuple or enum/Result payload closes exactly once on every
exit path, whether a match arm reads a bit-copy field off the binder, the
caller calls `.close()` explicitly on a returned record's fields, or the
binder is untouched: either the composite's drop owns the close or the
explicit close does, never both. This ships fixed at v0.6.0 on the current
lowerer with the minimum change that closes the hole (a silent double free
from an ordinary match arm is not a documented limit under D332), and the
exactly-once guarantee is re-proved as a P2-VALUES and P3-CALLABLES
acceptance case.

## Acceptance

- `hew run` on the Result<Handle, string> field-read repro prints exactly one
  close carrying the real field value, never a zeroed one
- The Pipe{sink,input} reduction exits 0 under `hew run` and as a built
  binary, and the no-explicit-close control still exits 0, so the fix trades
  no double close for a leak
- `cargo test -p hew-cli --test run_e2e run_record_of_handles_return_drops_each_field_once`
  passes with its status and stdout assertions unweakened

slepp:
## Why

A composite carrying an affine handle could close it twice: once through the
composite's own in-place field walk, once through the close the program asked
for. `Pipe { sink, input }` returned from a function and closed field by field
printed its output and then segfaulted at teardown, and `Result<Handle, string>`
matched with `.Ok(h)` ran the user `close(self)` a second time over the zeroed
variant slot, printing a close that carried no field value.

## What

Three producers failed to record that the leaf had left the composite.

- **mir (runtime receiver)**: a runtime symbol that consumes its receiver
  (`hew_sink_close`, `hew_stream_close`, the `Duplex` closes) carries no
  `ItemId`, so a receiver projected out of a live composite took the borrow path
  and never discharged the root-relative neutralize `note_carrier_projection`
  had already recorded. It now runs through the transfer funnel, which nulls the
  composite's slot so its field walk skips the leaf.
- **mir (record roots)**: the scope-exit projection fallback admitted only tuple
  roots, so a record root recorded nothing at all. It now admits a record root
  for a leaf whose release is a close — the leaf its own `RecordInPlace` walk
  would close again. A `Vec` field keeps the send and copy-on-write snapshot
  discipline that already owns it.
- **mir (enum shells)**: a `#[resource]` record payload moved into a match binder
  left the shell's generation live, and a record close has no null representation
  for the shell's walk to skip. The shell now retires where its payload leaves
  it, which retires the discharge-time rule that only covered an explicit
  consume, and the same retirement applies to a record that hands a resource
  field to that field's close.

The record retirement is whole-root, so a sibling resource field the program
does not close itself leaks rather than double-freeing; the marker at the site
carries the reduction and what per-field retirement would take.

## Test

`run_e2e` gains the no-close control for
`run_record_of_handles_return_drops_each_field_once`, the tuple field-projection
spelling, five `Result<Handle, string>` arms (field read, untouched binder,
explicit close, wildcard, and the declining producer that must close nothing),
and three record-of-resource-fields shapes including the documented partial-close
leak. `move_value` gains a record-root admission unit test and its unregistered-root
counterfactual. `declared_release_adoption` pinned the previous shape, where the
shell retired in the close's normal successor and its unwind edge still walked
the handed-over payload; it now pins the hand-off as the retirement point.
`tests/vertical-slice/accept/composite_resource_close_once_asan.hew` is registered
in `scripts/asan-fixture-check.sh`; on the pre-fix compiler it reports
`free(): double free detected in tcache 2`.

Fixes #3070


slepp:
Local evidence, with `hew-mir/src` reverted to the branch base (`322167de6`) for each reference run.

**ID diff, `cargo nextest run -p hew-cli --no-fail-fast`**

| | tests | passed | failed |
| --- | --- | --- | --- |
| base `322167de6` | 1492 | 1432 | 60 |
| this branch | 1492 | 1439 | 53 |

`comm -23 pr_fails base_fails` is empty: every failing id here also fails on the
base. Seven ids fail on the base and pass here, including the acceptance test the
issue names:

```
hew-cli::run_e2e run_record_of_handles_return_drops_each_field_once
hew-cli::run_e2e run_bound_tuple_field_close_drops_each_handle_once
hew-cli::run_e2e run_record_resource_field_partial_close_leaks_the_sibling
hew-cli::run_e2e run_record_resource_fields_both_closed_close_once_each
hew-cli::run_e2e run_result_resource_payload_field_read_closes_once
hew-cli::run_e2e run_result_resource_payload_untouched_binder_closes_once
hew-cli::run_e2e run_result_resource_payload_wildcard_arm_closes_once
```

**`make asan-fixtures`**: base `17 passed, 3 failed`, branch `18 passed, 2 failed`.
The fixture that flipped is the new one; on the base it reports

```
FAIL composite resource close-exactly-once (#3070): ASan/LSan reported findings:
==94944==ERROR: AddressSanitizer: attempting double-free on 0x6e5c859e0020 in thread T0:
    #1 0x5b78b413be36 in Slot::close composite_resource_close_once_asan
```

`string retain-on-share` and `owned-Vec retained param-embed temp (push/set/Arena)`
fail on both sides; both are `hew_string_length` / `hew_string_equals`
use-after-free in string and Vec refcounting, outside what this touches.

**Other gates**: `cargo nextest run -p hew-mir` 1922 passed; `make
test-vertical-slice` 616 fixtures, no failures; `cargo clippy --workspace --tests
--message-format=json -- -D warnings` reports nothing.

**Reductions**, `hew run` and as built binaries, base then branch:

```
base:   id=7 / close 7 / done / close 0          branch: id=7 / close 7 / done
base:   record-ok, exit 1 (139 built)            branch: record-ok, exit 0
base:   close true / done / close true / free(): double free detected in tcache 2
branch: close true / done, exit 0
```

slepp:
Per the D342 family decision: v0.6.0 fixes inside `hew-mir/src/lower` or
`hew-codegen-rs` ownership paths are admitted only when proportional (roughly
under 100 lines of lowerer change), one named root, and no sibling regression
under refutation; otherwise the defect is a documented fail-closed limit for
v0.6.0 and its fixtures and sanitizer oracles are registered as expected
failures for the P1/P2 coverage ratchet to flip. This diff exceeds that bar,
so it does not land for v0.6.0.

The test corpus is salvaged in #3280: the `run_e2e` fixtures, the
`composite_resource_close_once_asan.hew` sanitizer oracle, and the LESSONS.md
notes are carried over and registered as expected failures citing #3070. None
of the `hew-mir`/`hew-codegen-rs` diff is carried forward.

Closing in favour of #3280; the fix itself stays open against #3070 for the
v0.7.0 ladder.

Issue #3168, opened 2026-08-30
A nested aggregate payload matched out of a record field is released twice


slepp:
## Problem

The #3090 handoff covers a direct handle payload (`Option<Vec<T>>`). When the
payload is itself an aggregate that owns a handle, both the arm binder and the
parent record still release it.

```hew
type Inner { v: Vec<i64>; }

type Outer {
    other: Vec<i64>;
    value: Option<Inner>;
}

fn main() {
    let p = Outer { other: [7], value: Some(Inner { v: [1, 2] }) };
    match p.value {
        .Some(inner) => println(inner.v[0] + inner.v[1]),
        .None => println("none"),
    }
    println("ok");
}
```

```console
hew run repro.hew
3
ok
free(): invalid pointer
```

## Expected

The arm binder is the sole owner of the matched field's heap, whatever the
payload's shape, so the program prints `3` and `ok` and exits cleanly.

## Notes

`projected_enum_payload_is_handle_transfer` in `hew-mir/src/lower/ownership.rs`
proves only a direct handle leaf; a nested aggregate payload classifies
`ByteCopyAlias` and answers `false`. The marker on that helper names the real
fix: recurse the classification through `ByteCopyAlias` payloads and publish
the transfer for the exact nested leaf rather than the root generation.

slepp:
Absorbed into #3226, which carries this nested-aggregate payload match (`Outer`/`Inner` field match, released twice) as its third acceptance row and redirects the ownership.rs:3742-3750 shortcut marker to that issue.

slepp:
Absorbed into #3226 ("Release an owned field projected or matched out of a held aggregate exactly once"), the new aggregate-projection issue covering this class along with #3124 and #3156. The Outer/Inner nested match repro from this issue carries forward as that issue's third acceptance row.

Issue #3156, opened 2026-08-30
Matching an enum field of a consumed record parameter double-frees the payload


slepp:
## Problem

A payload bound out of an inline enum field of a by-value record **parameter**
is released twice: once by the arm binder and once by the parameter's scope
exit.

```hew
type Pair {
    other: Vec<i64>;
    value: Option<Vec<i64>>;
}

fn take(p: Pair) -> Vec<i64> {
    match p.value {
        .Some(v) => v,
        .None => [],
    }
}

fn main() {
    let p = Pair { other: [7, 8], value: Some([1, 2, 3]) };
    println(p.other[0] + p.other[1]);
    let taken = take(p);
    println(taken.len());
}
```

```console
hew run repro.hew
15
3
free(): invalid pointer
```

## Expected

The arm binder is the sole owner of the matched field's heap, so the program
prints `15` and `3` and exits cleanly.

## Notes

The same shape over a record **local** is fixed by #3090's
`publish_consuming_match_projection` restore - the parameter path does not
reach that publication, so its root generation still runs a terminal recursive
drop. `tests/vertical-slice/accept/match_record_enum_field_payload_owner.hew`
covers the local shape; this is the parameter twin.

slepp:
Absorbed into #3226 ("Release an owned field projected or matched out of a held aggregate exactly once"), which carries this repro (a match on an inline enum field of a by-value record parameter) as its second acceptance row, alongside #3124 and #3168.

slepp:
Absorbed into #3226, which carries this by-value-parameter match repro as its second acceptance row (alongside #3124 and #3168). Closing here; further discussion should go on #3226.

PR #3248, opened 2026-09-02
Release an owned field projected or matched out of a held aggregate exactly once

### LINKED ISSUE #3226: Release an owned field projected or matched out of a held aggregate exactly once
Problem

Three shapes of ordinary record code compile, print the right values, and then corrupt the heap at exit because an owned field moved out of an aggregate is released once by its new owner and again by the parent's drop. Reading `r.pair.0` out of a returned record into a new binding exits 139 (SIGSEGV) at teardown (#3124). Matching an inline `Option<Vec<i64>>` field of a record parameter taken by value aborts with `free(): invalid pointer` (#3156). Matching a nested aggregate payload (`Option<Inner>` where `Inner` owns a Vec) out of a record field aborts the same way (#3168). All three reproduce on hew 0.6.0-rc3-dev.146+1ded4d0c8 and `hew check` admits every one of them.

Expected

Projecting or transferring an owned field out of a record, tuple or enum the program still holds releases that allocation exactly once on every exit path, and the parent's remaining owned fields are still released once each. A read that leaves the source live does not neutralize the slot.

Notes

- D332 lets v0.6.0 ship ownership-generation ICEs as documented fail-closed limits but not miscompiles. These are miscompiles in first-week code, so they are fixed on the current lowerer under U371 before the v0.6.0 tag rather than walled off or documented.
- PR #3162 (fixes #3090) fixed the local-binding sibling and left the shortcut marker at hew-mir/src/lower/ownership.rs:3742-3750 naming #3168; that marker now points here. scripts/nextest-expected-failures.tsv:18 carries the #3124 row.
- The durable proof is P2-VALUES: SIR `destructure` yields owned parts from a consumed aggregate with no sibling authority, and plan §1.1 says this class cannot be written under it. The P2 exit gate names these three repros.

Outcome

All three repros print their expected output and exit 0 at O0 and O2, a leak/ASan run shows one release of each transferred allocation, and the same guarantee is re-proved as a P2-VALUES acceptance row through SIR.

Acceptance

- `hew run` on the #3124 reduction (`type T { pair: (Vec<string>, i64), tag: string }`, `mk()` returning T, main reading `r.pair.0` into a binding) prints the tag line and `1` and exits 0; the direct-field-overwrite sibling (`current = first.balancer`) exits 0.
- `hew run` on the #3156 repro (Pair with `other: Vec<i64>` and `value: Option<Vec<i64>>`, matched inside `fn take(p: Pair)`) prints `15` then `3` and exits 0.
- `hew run` on the #3168 repro (`Outer { other: Vec<i64>; value: Option<Inner> }`, `Inner { v: Vec<i64> }`, `match p.value { .Some(inner) => ... }`) prints `3` and `ok` and exits 0.
- `valgrind --leak-check=full --error-exitcode=1` over all three reports zero definitely-lost bytes and no invalid free, with a sibling-retention control proving the parent's other owned fields are released exactly once.
- `cargo test -p hew-cli --test funcupdate_consume_semantics accept_carry_nested_and_collection_bearing_tuples -- --exact` passes and the matching scripts/nextest-expected-failures.tsv row is deleted in the same commit.
- The direct-handle case PR #3162 fixed (`Option<Vec<T>>` matched straight out of a field) still passes as the negative control against a blanket suppression.
- The three repros are in the P2-VALUES corpus and pass through SIR when P2 lands.

Absorbs #3124, #3156, #3168.

slepp:
## Why

Three shapes of ordinary record code compiled, printed the right values, and
then corrupted the heap at exit, because an owned field moved out of an
aggregate was released once by its new owner and again by the parent's drop.
Reading `r.pair.0` out of a returned record exited 139 at teardown; matching an
inline `Option<Vec<i64>>` field of a by-value parameter and matching a nested
aggregate payload out of a record field both aborted on `free(): invalid
pointer`. `hew check` admitted every one of them.

They turned out to be three different root causes, in three different
authorities, and the third one is not the cause the issue names. `--dump-mir
raw` of #3156's callee shows the parameter has no owner generation at all
because the callee does not own it (`mode: BorrowReadOnly`); the competing
release is the CALLER's `RecordInPlace` drop, so there is no callee generation
to end.

## What

- **mir (payload classification)**: an aggregate enum payload is byte-copied
  into the arm binder with no per-leaf retain, so a handle reached through it
  is handed over exactly as a direct handle is. The handoff classification now
  recurses through the byte-copy class and stops at the first transferred leaf,
  which closes the shortcut marker that named #3168. A `string` payload still
  classifies retained and keeps the parent's cleanup.
- **mir (neutralize path)**: a tuple member of a record owner is now a hop in
  the root-relative neutralize path. It clears nothing itself; the handle one
  level deeper names `root.pair.0`, so the record keeps its composite drop and
  releases every sibling it still owns. Ending the owner's generation instead
  would have retired that composite drop, which is the posture the
  cancellation-token cursor test already forbids.
- **mir (body-escape summary)**: destructuring a projection of a parameter is
  positive owned-carrier evidence in its own right. The summary previously read
  a projection's `Consume` intent stamp as its only evidence, and a match
  scrutinee is stamped `Read` whatever its arms then bind, so the parameter was
  classified BORROW and the caller handed the callee its live record to move a
  leaf out of. The caller now hands over an owned copy and the callee discharges
  what it did not move; a match whose arms only inspect the scrutinee carries
  nothing out and is unaffected.

## Test

`match_nested_aggregate_payload_owner`, `projected_tuple_element_owner`, and
`match_param_field_payload_carrier` are new vertical-slice accept fixtures, each
carrying its own control: a retained `string` payload that must keep the
parent's cleanup, a read that leaves the source live and must not neutralize the
slot, and a `peek` callee that reads through the payload without carrying it
out. `projected_owned_field_transfer_asan` is a new ASan/LSan oracle for the
double-free axis. `accept_carry_nested_and_collection_bearing_tuples` greens, so
its `scripts/nextest-expected-failures.tsv` row is deleted here. Three new unit
tests in `move_value.rs` pin the hop's contract: a tuple member extends the path
to `fields=[0, 0]`, an inline enum member does not, and a non-null-safe handle
leaf still clears nothing.

All three reproductions plus their siblings exit 0 optimized and at `-g`, and
`valgrind --leak-check=full --error-exitcode=1` reports zero definitely-lost
bytes and no invalid free over each of them.

Fixes #3226


slepp:
Per D342: "v0.6.0 fixes inside hew-mir/src/lower or hew-codegen-rs ownership paths are admitted only when proportional (roughly under 100 lines of lowerer change), one named root, and no sibling regression under refutation. Otherwise the defect is a documented fail-closed limit for v0.6.0 and its fixtures and sanitizer oracles are registered as expected failures for the P1/P2 coverage ratchet to flip." This PR's own description names three different root causes in three different authorities (facts.rs consumption scanning, move_value.rs neutralize-chain seeding, ownership.rs payload-transfer classification), and the lowerer change is well over the proportionality bound even excluding its own new tests. Nothing here suggests the fix is unsound; it is deferred to the P1/P2 ladder work under a decision about scope, not correctness. The test material (the three vertical-slice fixtures with their negative controls, the ASan oracle, and the two LESSONS.md entries recording why the alternate fix routes each fail a counterexample) is salvaged without the lowerer changes in #3278, registered as known v0.6.0 limits. Closing in favour of #3278.

PR #3249, opened 2026-09-02
fix(mir): keep the payload a match over a channel recv produces

### LINKED ISSUE #3127: A match on an intercepted recv drops the payload it just produced
## Problem

Passing a tuple that contains both `channel.Receiver<string>` and `string`
through an actor message preserves the receiver and loses the string field. The
handler receives an empty string and otherwise runs successfully.

```sh
RUSTC_WRAPPER= make test-vertical-slice
```

The `actor_nested_handle_tuple_transfer` fixture prints:

```text
worker got payload: 
done
```

instead of:

```text
worker got payload: hello
done
```

The fixture sends `tx.send("hello")`, transfers `(rx, "payload")` to
`Worker.deliver`, destructures the tuple, and receives from `rx`. The empty
payload reproduces on every run with exit status 0.

## Expected

Actor-message lowering and dispatch preserve every owned tuple field exactly
once. The receiver yields `hello`, the tuple tag remains `payload`, and
teardown neither leaks nor double-releases either field.

## Notes

Actor message argument packing and unpacking for tuples with mixed
opaque-handle and owned string fields; the fixture is
`tests/vertical-slice/accept/actor_nested_handle_tuple_transfer.hew`.

## Outcome

A match whose scrutinee is a channel `recv`/`try_recv` call keeps the arm-bound payload alive through the match result: the program prints the received string, with no leak and no double release on either arm, and the downstream actor symptom (a handle tuple sent to a worker) prints its payload. A silent wrong answer from ordinary code is not a documented limit under D332, so this is fixed at v0.6.0 on the current lowerer under U371 (publishing an owner for the synthesized Option carrier), and P4-CONCURRENCY's SSA def/use for call scrutinees is the durable proof.

## Acceptance

- `hew run` on the minimal repro prints `got: [hello]`
- tests/vertical-slice/accept/channel_recv_match_result_survives.hew lands on main and passes under `make test-vertical-slice`, covering the direct-match and let-bound-carrier control
- tests/vertical-slice/accept/actor_nested_handle_tuple_transfer.hew passes, printing `worker got payload: hello` before `done`

slepp:
## Why

A `match` whose scrutinee is a channel or stream `recv` released the payload it
had just produced, so the program read an empty string where the sender wrote a
real one. `hew run` on the report's repro printed `got: []`, and the downstream
actor fixture printed `worker got payload:` with nothing after it.

A channel or stream `recv` is a codegen-intercepted layout-witness call. The
intercept deactivated its destination's crash cleanup before writing the
`Option<T>` carrier and never rearmed it, so the destination had no lifecycle
hook and MIR could not publish an owner for the carrier. Without a carrier
owner the arm binder became the sole owner of the payload, took a body-end
release, and that release canonicalized onto the match result place - nulling
the value the match had just produced.

## What

- **codegen**: the layout-witness recv intercept rearms its destination once
  `store_recv_option_via_layout` has written the payload slot and selected the
  tag, which is the post-write rearm hook the crash-cleanup admission catalog
  asks for; the four recv families join that catalog. One shared predicate
  drives the emitter dispatch and the admission, so the set that gets the
  intercept and the set that gets its rearm cannot drift.
- **mir**: match lowering mints the carrier owner for a recv scrutinee like any
  other call scrutinee, and the second, body-end release discipline that the
  suppression existed to protect is deleted with the symbol-keyed
  `is_recv_next_scrutinee` detector that selected it. A recv payload now has
  one release authority: the carrier owns the frame, the arm binder takes it by
  transfer, and the drop lands where the binder's generation ends.
- **tests**: the for-await release tests counted raw `Instr::Drop` on a return
  edge; that release is now scheduled by drop elaboration, so they ask the
  elaborated drop plans instead. The identity-forward case changed answer as
  well as mechanism - the binder owns its own count and the callee retains on
  return (`fn wrap` lowers to `string.retain` before its return move), so
  releasing the binder's place on the break edge is exactly-once.

## Test

- `tests/vertical-slice/accept/channel_recv_match_result_survives.hew` pins the
  direct-match shape for both `recv` and `try_recv` against the let-bound
  carrier control that already worked. The known-failure ratchet on
  `actor_nested_handle_tuple_transfer` is replaced by the ordinary stdout
  assertion; it prints `worker got payload: hello` before `done`.
- `tests/vertical-slice/accept/recv_frame_release_balance.hew` is the leak and
  underflow oracle for the discipline this change deletes, wired into
  `scripts/asan-fixture-check.sh` as well as the vertical slice: a drain, an
  early-return edge, an identity-callee forward, a heap-owning record payload,
  and the direct match shape, each repeated enough that one leaked count per
  iteration is unmissable.
- `forawait_early_return_edge_releases_current_iteration`,
  `forawait_break_edge_release_survives_sibling_return_path`,
  `forawait_break_forwarded_via_call_releases_the_binder_count_once` and
  `generator_pump_releases_record_yield_value_in_place` cover the same
  behaviour against the current release authority.
- The duplex recv families lost only the suppression: a duplex recv over an
  owned payload is refused upstream (`call-scrutinee ownership is unresolved`,
  identical on `origin/main`), so `duplex_split_roundtrip`'s `i64` shape is the
  only duplex match the carrier mint reaches.
- `make lint`, `make test-vertical-slice`, `make test-ownership-balance-corpus`
  and `make test-core-matrix` pass. `make asan-fixtures` reports the new
  fixture passing and the same two failures `origin/main` has at `e242af355`
  (`string retain-on-share`, `owned-Vec retained param-embed temp`). The
  `nextest -p hew-mir -p hew-codegen-rs -p hew-cli` failing-ID set is identical
  to that baseline's, with nothing added and nothing removed.

Fixes #3127


slepp:
Per the D342 decision on this ledger: v0.6.0 fixes inside `hew-mir/src/lower` or `hew-codegen-rs` ownership paths are admitted only when proportional (roughly under 100 lines of lowerer change), one named root, and no sibling regression under refutation — this PR's fix spans a codegen crash-cleanup rearm plus a match-lowering carrier-owner change and does not clear that bar for a v0.6.0 patch, so #3127 is being documented as a fail-closed limit for v0.6.0 and tracked for the v0.7.0 P1/P2 ladder instead. The test coverage here (the direct-match fixture, the release-balance fixture, and the LESSONS.md entries) has been salvaged as known-failure ratchets pinning the exact wrong output in #3277, so the shape stays covered and trips the moment the real fix lands. Closing this PR without its `hew-mir`/`hew-codegen-rs` changes.

PR #3251, opened 2026-09-02
fix(mir): keep a borrowed match payload owned by its scrutinee

### LINKED ISSUE #2523: Payload-projection-then-reassign on an enum/machine value drops a still-live alias (wrong result)
When a value is projected out of an enum or machine payload, used, and then the same variable is reassigned, the store side releases a value that is still live-aliased. The observable symptom is a wrong result: a program that should accumulate a log to length 13 instead reports length 1.

This is a store-side release-vs-live-alias defect, distinct from the actor-state load path — it reproduces on current main and is unaffected by the actor-state-field retain classification. The load side is byte-for-byte identical; the fault is that the reassignment store releases the prior binding while an alias to it is still live.

Next steps:
1. Reduce to a minimal `.hew` repro (payload projection, a use of the projected binding, then a reassign of the same variable).
2. Add a wrong-result / leak oracle that distinguishes the correct accumulation (length 13) from the corrupted one (length 1).
3. Fix the store-side release so a still-live alias is preserved.

## Outcome

A `match` arm that only reads a projected enum or machine payload keeps the scrutinee's composite drop and the projected binding's lifetime coherent: the four borrow-only shapes run under MallocScribble/MallocGuardEdges/MallocPreScribble to their asserted status with no double free and no null dereference. Those four are memory corruption from ordinary code and do not ship at v0.6.0; they are fixed on the current lowerer under U371 with the minimum change that closes the hole. `field_place_borrow_only` already fails closed as an ICE and may ship on the documented fail-closed list with this number if it cannot be fixed in the same change.

## Acceptance

- `cargo test -p hew-cli --test match_payload_reassign_uaf_oracle borrow_only -- --test-threads=1` passes every borrow_only test
- borrow_only_projection_keeps_scrutinee_live, captured_binding_borrow_only_is_valid, nested_enum_borrow_only_is_valid and borrow_only_guard_fallthrough_runs_safely each exit with their asserted status under the poisoned-allocator triple
- field_place_borrow_only_keeps_scrutinee_live either compiles and runs or refuses with a user-facing diagnostic naming the borrowed payload; if it stays an ICE at the tag it is on the release notes' fail-closed list with this number

slepp:
## Why

A `match` arm that only reads an enum payload corrupts memory. The arm binder was minted a release authority of its own even when the matched storage still scheduled the payload's drop, so two owners covered one buffer; the projection-adoption pass then resolved that by nulling the scrutinee's payload slot. The payload is freed at the end of the first arm and the scrutinee is left holding null, so a loop that re-matches the same value reads freed memory, and the shapes the neutralize cannot reach - a nested pattern, a closure-captured binding - double-free instead.

The five borrow-only controls on the issue reproduce it on `e242af355`: `borrow_only` and `guard_fallthrough` segfault, `nested_enum_borrow` and `captured_borrow` abort with `free(): invalid pointer`, and `field_place_borrow` fails closed as an ownership-generation ICE.

## What

- **mir (lowering)**: a payload binder is aliased onto the matched storage wherever the scrutinee origin proves that storage retains the payload - an owning binding, a re-readable place, a closure-environment copy, or the outer value behind a nested transient. Only an ephemeral producer hands a sole owner over. A genuine move-out is unaffected: the consume hook still transfers the authority at the move, or refuses it fail-closed. Applies at all four destructure sites (`match`, `let else`, `if let`, `while let`) and to their nested binders.
- **mir (model)**: `OwnershipEvent::DemoteToAlias` carries an `AliasDemotionReason`. An owned call carrier's snapshot drop expects the aliased slot cleared on the path where the binder escapes; a scrutinee that keeps reading its own payload must keep that slot populated. Both were one untyped event, and the projection-adoption pass needs to tell them apart.
- **mir (adoption)**: the pass no longer reads `Move(projection); Mint(dest)` as an ownership handoff when the demotion says the scrutinee retains the payload.
- **mir (field projection)**: the field-projection root generation ends only where the destructure really hands the payload over, so a borrow of a re-readable place keeps its release authority and a loop no longer re-ends a generation that already ended.
- **mir (overwrite)**: the enum-overwrite refusal counts live payload aliases rather than scope-exit owners, so a record or vector payload alias still refuses instead of falling through into a silent leak.
- **mir**: `ProjectedPayloadOrigin` and its provenance record move to `lower/projected_payload.rs`.

A `string` payload binder is the one projected binder that keeps an owner - its release is armed by a `Guard { ProjectedPayload }` flag the enum-overwrite authority sets when it hands the old generation over - so it still takes the unconditional adoption, and the borrow-only `string` read in a loop still returns the first iteration's length. Withholding the adoption for guarded owners is not the fix, because that neutralize is what keeps a genuine move-out from double-freeing; the neutralize has to be emitted under the same flag. That is #3252, marked as a shortcut where it lives.

## Test

- `hew-cli/tests/match_payload_reassign_uaf_oracle.rs`: the five borrow-only controls the issue names now run to their asserted status under the poisoned-allocator triple, including `field_place_borrow_only_keeps_scrutinee_live`, which no longer needs the fail-closed list. Two new fixtures, `if_let_borrow_only_keeps_scrutinee_live` and `while_let_borrow_only_keeps_scrutinee_live`, cover the two other destructure sites; both segfault without the `control_flow.rs` half of the change. The move-out shapes in the same file are the negative controls and are unchanged.
- `hew-cli/tests/support/payload_handoff_mir.rs`, `string_payload_handoff_leak_oracle.rs`, `bytes_payload_handoff_leak_oracle.rs`: the handoff oracles locate the payload binder by its projection move and expect a release only where the stream still grants one.
- `hew-mir/tests/lowering_expr/owner_mint_warrant_seams.rs`: the two domestic controls now assert the alias demotion and the absent adoption alongside the shell's unchanged release.
- `hew-codegen-rs/src/llvm_tests.rs`: the projected-resource escrow check counts escrowed owners rather than naming one local.

Verified on macOS as well as Linux, since the poisoned-allocator and `leaks(1)` oracles only assert there. Against an `origin/main` baseline built on the same host, the branch fixes the five borrow-only controls and leaves a strict subset of the baseline's failures, every one of them already on the ratchet ledger.

Fixes #2523


slepp:
Per the D342 admission decision: a v0.6.0 fix inside `hew-mir/src/lower` or `hew-codegen-rs` ownership paths is admitted only when it is proportional (roughly under 100 lines of lowerer change), has one named root, and causes no sibling regression under refutation. This PR's lowerer change exceeds that threshold, so the defect stays a documented v0.6.0 fail-closed limit tracked under #2523 rather than landing now. The pure test material (the `if let`/`while let` borrow-only fixtures and the LESSONS.md analysis) has been salvaged into #3279, which registers all seven borrow-only shapes as known failures against #2523 for the P1/P2 ownership-ladder coverage ratchet to flip when the fix lands.

Closing in favour of #3279.

PR #3260, opened 2026-09-03
feat(sir): carry ownership kinds and operations in SIR

slepp:
## Why

We need SIR to be the single authority for ownership so lowering, verification, and MIR cannot classify the same value differently.

## What

- Publish structural class facts and canonical recursive/imported declaration identities from `hew-types`.
- Carry per-binding provenance, ownership kinds, explicit ownership operations, places, traps, and suspension shapes in SIR.
- Make verification and SIR-to-MIR lowering consume those facts without fallback classifiers or compatibility paths.
- Use Hew dotted construction syntax in the Result-returning `std.os` environment helper.

## Test

- Behavioural class-fact and SIR verifier tests cover aggregate, generic, recursive, imported, opaque-handle, provenance, ownership-operation, trap, and suspension cases, including rejection controls.
- `channel_auto_close_scope` and `channel_recv_select` raw/elaborated MIR goldens record the intentional per-binding shape: the real tuple binding owns the result directly, with no synthetic `__hew_produced_value` owner, transfer, neutralization, or drop.
- Focused checked-MIR redump verification for both affected fixtures.
- `make structural-lint`.


slepp:
Final review against `plans/briefs/P1-L1.md`. **NO-GO.** Five required fixes below, then this is ready to re-review.

What is not in question: every **Files/Create** item exists, every **Interfaces** signature landed as specified, all nine `repros/ladder/p1/` fixtures are present, all fourteen commits are signed by the author, and no session marker, `Claude-Session` line or claude.ai URL appears in any commit or in the diff. The acceptance greps pass - `UseMode`, `EffectSummary|effects_in`, `fn is_copy` in `ty.rs` and `.is_copy()` under `hew-types/src/check` are all empty; `fn of_ty` and `pub fn mangle_resolved_ty` are one hit each; `ValueClass::Unknown` is still 21 sites and `is_copy_layout` is unchanged, both carrying their `LEGACY (P5)` marker; `hew-hir/src/monomorph.rs:346` is a `pub use`. The `hew-hir/src/value_class.rs` carve-out is comment-only, and every new `SemOpKind`/`SemTerminator` arm added to `hew-mir/src/sir.rs` refuses rather than lowers.

---

**1. The branch does not merge, and the required checks have no run.**

Problem: `git merge-tree --write-tree origin/main HEAD` exits 1 with `CONFLICT (content): Merge conflict in LESSONS.md`; `gh pr view` reports `mergeable: CONFLICTING`, `mergeStateStatus: DIRTY`. `gh pr checks 3260` lists only `Analyze (javascript-typescript)` and `Analyze (python)` - neither `Clippy & format` nor `Build & test (Linux)` has a run on `900e803c9`.

Expected: rebase onto `origin/main` (13 commits ahead of the base `7c1490375`), resolve `LESSONS.md`, push, and let the two required legs run. `--auto` cannot be armed against a conflicting branch.

**2. Four of the six D359 repair items are not implemented as decided, and the PR body re-argues two of them.**

The brief frames those items as "decided by the architect from the variance report so the repair does not become a second round of guessing".

- `SemModule.type_facts` still ships empty: `hew-sir/src/lower.rs:798 type_facts: BTreeMap::new()`, behind the same marked shortcut the repair section quotes as the thing to remove. Brief: "**`SemModule.type_facts` is populated by `hew_sir::lower_module` from `TypeCheckOutput`.** ... This is L1's projection to own, not L3's". PR body defect 12 argues it is L3's; that reading was already overruled. Consequence in this diff: `OwnKind::of_ty` (`hew-sir/src/ownership.rs:59`) reads `ClassContext::empty()`, so every user `Named` type refuses.
- The indirect enum refuses instead of classing. `hew-types/src/value_class.rs:568-584` returns `ClassError::RecursiveDeclaration`. Brief: "A recursive occurrence is an owning edge: an `indirect` enum classes as `CowValue` or `AffineResource` by its payload, the same as any other aggregate". A refusal is also against **Must NOT** - "§1.1 is total; a type you cannot class is a bug in the table, not a case for an escape hatch." The brief also asks for the indirect-enum case in the per-variant table test, positive and negative; `class_indirect_enum_refused.hew` asserts the refusal instead.
- No `Guaranteed` constructor. `grep -rn Guaranteed hew-sir/src` finds comments, one `assert_ne!` and the dump arm - no producer. `hew-sir/src/lower.rs:1113-1126` is a marked shortcut instead. Brief: "Add the constructor so a `Borrow` header slot always produces `OwnKind::Guaranteed`." The shortcut's stated reason (`SemParamPassing` has one variant, `ReadOnly`) is worth putting to the architect, but the item is unmet.
- `ValueDef.provenance` is `Some` for parameters only. `hew-sir/src/lower.rs:1244` is a marked shortcut on the `let` path. Brief: "**`ValueDef.provenance` is `Some` for every `let`/`var` binding**, not only for function parameters."

**3. The one repair item the brief forbids in L1 is implemented.**

Problem: `hew-sir/src/verify.rs:1506 verify_own_kind` cross-checks `ValueDef.own`/`BlockArg.own` against the class table, and `verify.rs:1428-1435` raises `InvalidTerminator` for `Trap` and `Suspend`.

Expected: both reverted. Brief: "**Own-field cross-check and `Suspend`/`Trap` verification are L2's**, not L1's ... Do not implement these in L1; the field and shape stay as L1 lands them, L2 adds the checks." **Must NOT** repeats it - "Do not add verifier rules. Rules 1-6 are L2's; `hew-sir/src/verify.rs` keeps exactly the checks it has today, adjusted only for the shape change." The `OwnershipKind` diagnostic and the tests named for it in the Test section (`verifier_rejects_a_suspend_no_relation_row_admits`, `verifier_rejects_a_trap_endpoint_it_states_no_rule_for`, `verifier_refuses_an_own_kind_the_class_table_contradicts`, `verifier_refuses_a_value_whose_type_the_class_rule_cannot_decide`, and their two admitting counterfactuals) go with them.

**4. The size arithmetic in the PR body is wrong.**

Problem: the body's **Size** paragraph says "52 files changed, 4,130 insertions, 907 deletions". `git diff --shortstat 7c1490375 HEAD` on the pushed head gives `52 files changed, 4154 insertions(+), 907 deletions(-)`, and `gh pr view` reports the same 4,154.

Expected: re-measure against the merge base after the rebase and state the measured pair. The brief budgets ~2,600 additions for L1; say plainly that the figure is over it and why, rather than leaving the reader to notice.

**5. The `ResolvedTy::from_ty` Err sites are reported by error kind, not by site and program.**

Problem: the body says `from_ty` returns `Err` at the thirteen copy sites "on real programs - `TaintedError` for a leaked error type, plus `UnresolvedAssocProjection` and `UnresolvedInference`", and the fail-closed arm is already written.

Expected: the brief asks for something narrower and earlier - "**Where `from_ty` returns `Err` on a program that reaches the site, stop and report that site with the program before writing its replacement** ... Whether an unresolved `Ty` fails closed or keeps a narrowed predicate is a checker decision the architect makes on your report". Give the file:line of each site that observes an `Err` and the program that reaches it, so the architect can rule on it.

---

Two smaller things, not blocking:

- The **Test** section carries counts - "62 fixtures x 2 stages", "59 fixtures executed". The voice rule for `## Test` is named tests, no pass counts.
- Clause 3 asks for "a compile-fail or type-level demonstration that a consumer cannot read a `bool` out of" `SendFact`, and "assert that shape holds". What landed is a doc comment at `hew-types/src/type_facts.rs:39` plus a hand-written match inside `a_closure_row_defers_its_send_fact`. A `bool` cannot in fact be read out today, so this is a wording gap rather than a hole.

Recorded for the architect, not counted against the lane: the `LocalPid`/`HewActor` and BitCopy-scalar marker corrections did not land on `BuiltinType::marker()`. The body names the failing tests that force it, and it is the same parity-oracle argument the brief itself accepts for `ValueClass::Unknown` and `is_copy_layout`. The other ladder defects the body records - the §9 phase split, the two §1.5 mode gaps, the `Yield`/`Select` questions - are reported rather than hidden, which is what the brief asks for.

Issue #3265, opened 2026-09-03
The gen fn return spelling is refused as a yield type mismatch with no named code


slepp:
## Problem

`gen fn f() -> Generator<Y, R>` is refused, but as a plain type mismatch on
the `yield` operand:

```hew
gen fn counter() -> Generator<i64, ()> {
    yield 1;
}
```

```
error: type mismatch: expected `Generator<i64, ()>`, found `i64`
```

The diagnostic blames the `yield` for the annotation's mistake, names no
code, and offers no fix-it. The return type of a `gen fn` names the yield
type, so the annotation is what is wrong.

## Expected

`gen fn f() -> Generator<Y, R>` is `E_GEN_RETURN_SPELLING` (User) reported at
the annotation, with a fix-it that replaces it with `Y`.

## Notes

HEW-SPEC-2026 §4.12 states the rule.

The pattern half of the variant-spelling rule (`E_BARE_VARIANT_PATTERN` as an
error rather than a warning) is #3244 and is not part of this issue.

Issue #3264, opened 2026-09-03
A body-less machine transition to a bare state name is refused as a bare variant


slepp:
## Problem

A body-less machine transition to a bare state name is refused as a bare
variant expression, although the name after `=>` is a state name and not a
variant:

```hew
machine Toggle {
    events { Flip; }
    state Off;
    state On;
    on Flip: Off => On;
    on Flip: On => Off;
    default { state }
}
```

```
error: E_BARE_VARIANT_EXPR: bare variant `On` is not an expression; use `.On` ...
```

The parser accepts the form. The refusal comes from the checker treating a
transition target as an expression, and the fix-it it offers teaches that a
state is a variant at the surface.

## Expected

`on Flip: Off => On;` type-checks. The name after `=>` resolves against the
machine's `state` declarations, and the variant-spelling rule does not reach
it. A `;` body is legal in every transition form, guarded ones included.

## Notes

HEW-SPEC-2026 §3.11.3 states the rule. `E_BARE_VARIANT_PATTERN` is separately
tracked and unaffected; this is only about the transition-target position.

Issue #3263, opened 2026-09-03
break accepts a discarded operand and scope binds () in value position


slepp:
## Problem

Two surfaces accept a value where none can go.

`break` accepts an operand and evaluates it for side effects only:

```hew
fn main() {
    var total: i64 = 0;
    loop { total = 1; break total; }
    println(f"{total}");
}
```

`hew check` reports `OK`. A reader writing `break total` reasonably expects a
loop-as-expression, which the language does not have.

`scope { .. }` is Unit-typed, so `let r = scope { .. }` binds `()` silently
instead of refusing the value position.

## Expected

- `break expr;` is `E_BREAK_VALUE` (User) with the hint "assign to a `var`
  before `break`". Bare `break`, `break @label`, and `break` in expression
  position as a diverging `!` value are unaffected.
- `scope` leaves `Primary`. `let r = scope { .. }` is `E_SCOPE_IS_STATEMENT`
  (User). The value-producing fan-out stays `join { .. }`.

## Notes

HEW-SPEC-2026 §12.4 and §4.2 carry the rules and the `BreakStmt` fragment
without its operand.

Issue #3262, opened 2026-09-03
Retire the seven inert keywords and make default, emit, and pool contextual


slepp:
## Problem

Seven words are lexer keywords whose whole behaviour is a refusal message or
a mark that marks nothing: `try`, `catch`, `race`, `cooperate`, `foreign`,
`super`, and `async`. `catch` has no parser reference at all. Three more
(`default`, `emit`, `pool`) are unconditional keywords although each is only
meaningful in one position, so an actor field named `pool` or a variable
named `default` is a parse error for no reason. `let mut x = 0` produces a
cascade of parse errors rather than one diagnostic naming `var`.

Reserving a word costs nothing before 1.0 and cannot be undone after it.

## Expected

- `try`, `catch`, `race`, `cooperate`, `foreign`, `super`, and `async` are
  ordinary identifiers. `try`, `catch`, and `foreign` keep an
  identifier-position hint naming the replacement.
- `async fn` is `E_NO_ASYNC_FN` and `async gen fn` is `E_NO_ASYNC_GEN`, both
  User channel, both with a fix-it that deletes the word.
- `for` over a `Stream` is `E_FOR_STREAM_NEEDS_AWAIT` (fix-it adds `await`)
  and `for await` over a generator or a collection is `E_AWAIT_NOT_STREAM`
  (fix-it drops it).
- `default`, `emit`, and `pool` become contextual, recognised only in a
  machine's fallthrough arm, a machine transition body, and a supervisor
  body.
- `mut` stays reserved for `*mut T`, and `let mut <ident>` is one error
  carrying the `var` fix-it.
- `docs/syntax-data.json` is regenerated from the lexer and its
  `reserved_unused` group shrinks to what actually remains.

## Notes

HEW-SPEC-2026 §12 carries the keyword table this describes, and §4.12 the
generator half. `budget` is also a reserved token with no parser use; the
surface decision does not name it, so it stays reserved unless a follow-up
decision retires it.

Downstream: `docs/syntax-data.json` feeds `tools/downstream/`, `editors/`,
and the sibling highlighter repos, so this change carries a grammar sync.

slepp:
Two things the spec text for this rule (HEW-SPEC-2026 §12, §4.12, landed in
#3268) leaves for the enforcement change to settle:

- `AsyncGenerator<Y>` outlives its only producer. `std/builtins.hew` declares
  the builtin and `impl<Y> Iterator for AsyncGenerator<Y>`, and the checker
[...]

PR #3268, opened 2026-09-03
docs(spec): name the parser as grammar authority and close the keyword set

slepp:
## Why

The specification named two standalone reference grammars as the authority
over the language surface. Nothing generated them, no gate compared them to
the compiler, and they asserted productions the parser refuses while omitting
productions that ship — a second authority that drifts silently is worse than
none. Alongside that, seven keywords were reserved for constructs that either
never reached the parser or marked nothing, the attribute set was open enough
that a misspelled `#[test]` dropped a test and exited 0, and several surfaces
accepted a value where none can go. These are surface decisions: free before
the edition stabilizes, unaffordable after.

## What

- **§12**: `hew-parser` is the grammar authority, `tree-sitter-hew/grammar.js`
  its mirror. `docs/specs/grammar.ebnf` and `docs/specs/Hew.g4` are deleted,
  and the remaining `ebnf` blocks are labelled as illustrations of surface
  spelling rather than a grammar. A closed keyword table lands: `try`,
  `catch`, `race`, `cooperate`, `foreign`, `super`, and `async` are ordinary
  identifiers, each with the reason it was reserved and what replaced it;
  `default`, `emit`, and `pool` join the contextual list; `mut` stays
  reserved for `*mut T` with `let mut x` a single error carrying the `var`
  fix-it. An enforcement note under the table records that the lexer has not
  closed to it yet, so `docs/syntax-data.json` still carries the retired
  words — a defect against the table rather than a second reading of it,
  stated the way §3.1 states the bare-variant expression gap.
- **new §12.6**: the attribute set is closed behind `E_UNKNOWN_ATTRIBUTE`,
  with a table of every attribute and its legal positions and a second list
  of the substrate attributes gated to `std/`. `#[noncancellable]` and the
  `#[on(upgrade)]` hook kind are recorded as removed, and §9.1.2 and §4.5
  stop holding a place for them.
- **§3.1**: one variant spelling in both positions and every enum.
  `E_BARE_VARIANT_PATTERN` is enforced at v0.6.0; the prelude constructors'
  bare expression spelling is named as a defect against the rule rather than
  a permitted form (#3240).
- **§4.2, §12.4**: `scope` is a statement (`E_SCOPE_IS_STATEMENT`) and `break`
  carries no operand (`E_BREAK_VALUE`); the `BreakStmt` fragment loses its
  `Expr?`.
- **§3.11.3**: a machine transition target is a state name, not a variant, so
  the §3.1 rule does not reach it and a `;` body is legal in every transition
  form.
- **§4.12**: `gen fn f() -> Y` names the yield type, with
  `E_GEN_RETURN_SPELLING` for the handle spelling; `async gen fn` is gone,
  and the `for`/`for await` mismatch codes are stated.
- **HEW-FUTURE**: §1.2 cancellation tokens move to `[REJECTED]`, §1.4 records
  that first-completion-wins is the task-await arm and `race` is not
  reserved, §1.6 drops `async gen fn`.
- Supporting updates: the guide's `let mut` note, the `grammar-mirror` lesson
  row, and the three `grammar.ebnf:N` citations in the Sublime grammar and
  the tree-sitter keyword patcher.

Decisions landed: D3 (pattern half only, per ledger D341), D5, D6, D7, D16,
D19.

## Test

- `make test-doc-examples` — passes. `spec-0134` and `spec-0135` moved to
  dotted patterns, so both checksums in
  `scripts/doc-test-expected-failures.txt` were re-verified against the
  binary and updated; both still fail for the recorded SNIPPET reason (a bare
  `let`/`var` at file scope).
- `make hew-check-all` — corpus sweep passes.
- `cargo test -p hew-lexer -- syntax_data` — `syntax_data_json_covers_all_keywords`
  passes; `docs/syntax-data.json` is exported from the lexer and is not
  touched here.
- `make grammar-parity` does not exist as a target, and no other gate
  compares the tree-sitter mirror to the parser. Building it is a build
  lane, not this PR, and it is now tracked as #3287 rather than left to the
  exit-gate list.

Enforcement of the rules this PR makes normative is tracked in #3261
(unknown attributes), #3262 (the keyword diet), #3263 (`break` and `scope`),
#3264 (the machine transition target), and #3265 (the `gen fn` return
spelling). The pattern half of the variant rule is already in flight as
#3244.

Two comments that reach the built ast-grep Hew dialect still cite
`grammar.ebnf` by line. They are context lines in
`tools/tree-sitter-hew-consume-parameter.patch`, reproducing text that lives
in `tree-sitter-hew`'s `grammar.js`, so rewriting them here alone makes the
patch fail to apply; the fix belongs upstream and is #3283.


slepp:
Required before merge (D348): PR 3267 is the authority for HEW-FUTURE.md section 1.2 (cancellation tokens REJECTED). Drop this PR's hunk on that section so the two do not race; keep everything else.

Issue #3256, opened 2026-09-03
Node setup is one typed call and the registry knows the actor type


slepp:
## Problem

Node setup is a sequence of untyped calls, and a refused start does not stop
the program. `Node.set_transport`, `Node.load_keys`, `Node.allow_peer`, and
`Node.start(addr)` must be called in order, each carries one fact that belongs
in configuration, and a `Node.start` that fails prints to stderr and returns
while the program runs on. That is a fail-open surface in a fail-closed
language.

The registry has the matching hole. `Node.register(name, pid)` returns `i32`
and the runtime entry is name to location; `Node.lookup<T>(name)` is registered
with `T` free, so the value type is whatever the reader wishes for. A lookup
that names the wrong actor type is an unchecked cast.

## Expected

One call, typed:

```hew
let config = NodeConfig.at("127.0.0.1:9101");
match Node.start(config) {
    .Ok(_) => println("node up"),
    .Err(e) => println(f"node refused: {e}"),
}
```

`Node.start(config: NodeConfig) -> Result<(), NodeError>` is the only start.
`NodeConfig { bind, transport, key, trust, peers, seeds }` is a prelude record;
`NodeConfig.at(addr)` fills the defaults. No field is inert: `Node.start` pins
every `peers` entry, the slot a peer occupies being its index, dials every
`seeds` entry once while skipping its own bind address, and admits
`trust = "pinned"` only, returning `NodeError.Config` otherwise. Re-dial with
backoff and `NodeConfig.load()` from `[cluster]` plus `HEW_NODE_*` are v0.8.0.

`Node.set_transport`, `Node.load_keys`, and `Node.allow_peer` are deleted:
their facts are config fields. `Node.connect(addr) -> Result<(), NodeError>`
stays for explicit dials, `Node.unregister` becomes a registered builtin, and
`Node.shutdown()` stays `()`.

The registry knows the actor's type.
`Node.register(name, pid: LocalPid<A>) -> Result<(), RegisterError>` records
`A`'s declaration identity beside the location, and `Node.lookup<A>(name)`
compares it, returning `Err(LookupError.TypeMismatch)` on disagreement.
`Node.register` is the one registration verb: it registers locally whether or
not a node has started, and publishes cluster-wide once one has.
`whereis<A>(name) -> Result<LocalPid<A>, LookupError>` is the local view of the
same registry. At v0.6.0 the identity is recorded and compared locally; the
gossiped identity rides the wire-version bump, and the cluster-wide compare is
[...]

Issue #3255, opened 2026-09-03
link and monitor are actor-context only and share one typed error


slepp:
## Problem

`link` and `monitor` subscribe the caller to a `DOWN` or a linked exit, which
only an actor can receive. Called from `main` they succeed anyway: the runtime
call returns and the program prints `linked` (`adv3r2/link_free_fn2.hew`). The
subscription has no reader, so the failure is invisible at both compile time
and run time.

The typed surface is also two surfaces. `link` returns
`Result<(), LinkError>` with ten variants, `monitor` returns
`Result<MonitorRef, MonitorError>`, and neither error set names the missing
context.

## Expected

`link(pid) -> Result<(), LinkError>` and
`monitor(pid) -> Result<MonitorRef, LinkError>`, one error type for both, with
three inhabitants: `Dead`, `Partition`, and `NoContext`. Local and remote pids
use the same set, because a cross-node link carries a `PartitionPolicy` and
needs the typed failure as much as a local one does. `monitor` on an
already-dead pid is not an error; it delivers `DOWN` at once, so `NoContext` is
its only `Err`.

A direct `link` or `monitor` in `main` is `E_ACTOR_CONTEXT_REQUIRED` at compile
time. The same call reached through a free function called from `main` is the
same fact decided at run time, `Err(LinkError.NoContext)`, because D9 makes the
execution context dynamic and a static "actor-only function" marker would
colour every function that might one day link. Neither form succeeds silently.

The `Dead` arm is produced once dead-target resolution lands (#3254); until
then it is never returned, and the fixture says so.

## Notes

Decision D12 of the v0.6.0 surface review. The rule is normative in
`docs/specs/HEW-SPEC-2026.md` §5.7.1 with an implementation-status note
pointing here. The "interim window" named at
`tests/vertical-slice/accept/link_monitor_value_result.hew:1-17` closes with
this change.

Issue #3254, opened 2026-09-03
A dead send target reports Dead instead of returning normally


slepp:
## Problem

A dead target has several shapes, and one of them is silence. A send to a
supervised child whose restart budget is exhausted returns normally, both from
`main` and from an actor holding a `ChildRef` (`sup_exhaust.hew`,
`sup_exhaust_actor.hew`). An unbounded send is unit-typed, so there is nowhere
for the fact to go, and the caller carries on as though the message landed.
`SendError` spells the closed case `Closed`, which does not distinguish a
mailbox that was closed under a live actor from a target that is gone.

This is a fail-open in a fail-closed language: liveness is guessed, and a lost
delivery is dropped silently.

## Expected

Resolution never hands back a dead address. A supervised role whose occupant is
gone resolves closed: a `ChildRef` send reports `Dead`, `whereis(name)` is
`None`, and `Node.lookup(name)` is `Err(LookupError.NotFound)`.

A held handle reports the same fact at the send. Every send expression has type
`Result<(), SendError>`. For an unbounded mailbox the only inhabitant of `Err`
is `SendError.Dead`; for a bounded or policy-sensitive mailbox `Dead` joins the
policy variants. A send to a dead target never traps and never returns `Ok(())`.

Discarding the result stays the existing `must_use` split: a discarded
unbounded-mailbox send is a warning, a discarded policy-sensitive send is an
error. Statement-position sends keep compiling.

`AskError` gains `Dead` and `StaleRef`, so a remote ask's failure kinds are one
enum.

## Notes

Decision D11 of the v0.6.0 surface review, frozen at v0.6.0 with the mechanism
at v0.7.0. The rule is normative in `docs/specs/HEW-SPEC-2026.md` §5.6 and §9.3
with an implementation-status note pointing here.

Issue #3253, opened 2026-09-03
Pool arity is a child clause, not an init field


slepp:
## Problem

Pool arity is spelled as an init field. `pool ws: Worker(count: 2, value: 7);`
puts `count` in the same parenthesised list as the actor's own fields, so the
two namespaces collide. An actor that declares a field named `count` cannot be
pooled, and the diagnostic blames the user's field rather than the collision.

Today the clause form is refused:

```
pool workers: Worker(value: 7) count: 2;
                                ^^^^^ unknown supervisor field: Some(Identifier("count"))
```

## Expected

`count:` is a per-child clause beside `restart:` and `shutdown:`, and the
parenthesised list stays the actor's field namespace:

```hew
supervisor Farm {
    strategy: simple_one_for_one;
    intensity: 3 within 60s;
    pool workers: Worker(value: 7) count: 2;
}
```

`count:` is required on a `pool` child and rejected on a `child` declaration.
An actor with a field named `count` pools like any other, and its `count` field
is set the way every other field is.

The old spelling migrates with a fix-it: `pool ws: A(count: N, ..)` becomes
`pool ws: A(..) count: N;`.

## Notes

Decision D13 of the v0.6.0 surface review. The rule is now normative in
`docs/specs/HEW-SPEC-2026.md` §5.1 with an implementation-status note pointing
here. Related: #2245 (config-derived pool counts).

PR #3271, opened 2026-09-03
docs(spec): make the actor, runtime, and node surfaces fail closed

slepp:
## Why

Six of the ratified v0.6.0 surface decisions govern actors, the runtime, and
the distributed node, and none of them was written down. The spec still said an
actor enters `Stopping` through a bare `stop()`, that a send to a terminal actor
is a silent no-op, that pool arity is an init field, and that starting a node is
a sequence of untyped calls whose refusal prints to stderr. Those are the
fail-open readings the decisions replace, and a reader who follows the spec today
writes the surface that is going away.

## What

- **§2.1, §9.1 (D10, A379)**: stop is a method with one signature. `self.stop()`
  inside the actor, `pid.stop()` from outside returning `()` and idempotent on
  an actor that has already stopped or crashed. `stop` becomes a reserved
  handler name (`E_RESERVED_HANDLER_NAME`). Messages queued behind a stop are
  dropped and the drop is disclosed through the `DOWN` record's `dropped: n`
  and `mailbox.dropped_on_stop_total{actor}`. The §9.1 transition line names
  `self.stop()` as the second source of the `Running` to `Stopping` edge.
- **§4.2, §4.3, §12.3 (D9, A372)**: every function may suspend and none is
  marked as suspending; what differs is the execution context the call runs on.
  §4.3 gains a closed suspension-point set, and the v0.6.0 refusal from `main`
  is `E_LIMIT_MAIN_CONTEXT` with the two shapes that work today named in it.
- **§2.1.1, §5.6, §9.3 (D11, A365)**: resolution never hands back a dead
  address, and a held handle reports the fact at the send as
  `Err(SendError.Dead)` under the existing must_use split. `Dead` joins the
  mailbox send state machine as a terminal state the policy table is never
  consulted for. §2.1.1's typing lines now say which half is the v0.6.0 typing
  and which is the frozen rule, so the two sections agree.
- **§5.7.1 (D12)**: `link` and `monitor` share one typed error with three
  inhabitants and are actor-context only. Direct from `main` is
  `E_ACTOR_CONTEXT_REQUIRED`; through a free function it is
  `Err(LinkError.NoContext)` decided at run time, because the execution context
  is dynamic and a static actor-only marker would be colouring.
- **§5.1 (D13)**: pool arity is a `count:` child clause beside `restart:` and
  `shutdown:`, so the parenthesised list stays the actor's field namespace and
  an actor with a field named `count` pools like any other.
- **§11, HEW-DIST-SPEC §2.2, §3, §7, §13 (D23)**: node setup is
  `Node.start(NodeConfig) -> Result<(), NodeError>`, `set_transport` /
  `load_keys` / `allow_peer` are gone into `NodeConfig` fields, a peer's route
  slot is its index in `NodeConfig.peers`, and the registry records the actor's
  declaration identity so `Node.lookup<A>` compares rather than casts
  (`RegisterError`, `LookupError.TypeMismatch`).
- **guide and `docs/language/`**: the cancellable-worker example renames its
  handler off the reserved `stop`, the actors and supervision modules gain the
  stop, dead-target, and pool-clause rules, the concurrency module names
  `E_LIMIT_MAIN_CONTEXT` instead of describing `main` as lacking a parameter,
  and the distributed chapter builds a `NodeConfig`.

Every rule the shipped compiler does not accept carries an
implementation-status note naming its tracking issue: #3193 (D10, with the
A379 scope recorded on it), #3195 and #3196 (D9), #3254 (D11), #3255 (D12),
#3253 (D13), #3256 (D23). No compiler code changes.

## Test

`make test-doc-examples` and `make hew-check-all` pass with the same extracted
fence set and the same expected-failure set as the branch point, so no fence
was added, removed, or newly broken and no expected-failures entry was needed:
the new surface appears only in fences already marked `<!-- doctest: skip -->`
and in `HEW-DIST-SPEC.md`, which the harness does not scan.
`make doc-ratchet-selftest` passes. The three refusals the status notes claim
were reproduced against the branch's own build before the notes were written:
`self.stop()` / `pid.stop()`, `pool ws: A(..) count: N;`, and
`NodeConfig.at(..)`.

Decisions D9, D10, D11, D12, D13, D23; ledger A365, A372, A379.

### DIFF
--- a/docs/hew-language-guide.md
+++ b/docs/hew-language-guide.md
@@ -1651,7 +1651,7 @@ actor Worker {
         ticks = ticks + 1;
     }
 
-    receive fn stop() {
+    receive fn halt() {
         running = false;
     }
 
@@ -1663,7 +1663,7 @@ actor Worker {
 fn main() {
     let worker = spawn Worker(running: true, ticks: 0);
     sleep(150ms);
-    worker.stop();
+    worker.halt();
     sleep(150ms);
     let r = await worker.total();
     match r {
@@ -1673,11 +1673,20 @@ fn main() {
 }
 ```
 
-Here `receive fn stop()` is a user-defined actor message. It is unrelated to the
-`#[on(stop)]` lifecycle hook, which is a plain `fn` invoked when the actor is
-tearing down. The `sleep_loop_blocks_mailbox` lint warns on the mailbox
-starvation shape where a receive handler loops around `sleep`/`sleep_until`
-without an in-loop exit path.
+Here `receive fn halt()` is a user-defined actor message, and it is the flag
+flip, not the actor's death. It is unrelated to the `#[on(stop)]` lifecycle
+hook, which is a plain `fn` invoked when the actor is tearing down. The
+`sleep_loop_blocks_mailbox` lint warns on the mailbox starvation shape where a
+receive handler loops around `sleep`/`sleep_until` without an in-loop exit path.
+
+`stop` is a reserved handler name, so the handler above is `halt`, not `stop`.
+Stopping is a method with one signature: inside the actor, `self.stop()`
+finishes the current handler, runs `#[on(stop)]`, and stops; from outside,
+`pid.stop()` requests the same and returns `()`, doing nothing if the actor has
+already stopped or crashed. A `receive fn stop()` is `E_RESERVED_HANDLER_NAME`,
+whose fix-it is to rename the handler or to call `self.stop()`. The shipped
+compiler still exposes the older free-function `stop(actor)` builtin
+(hew-lang/hew#3193).
 
 ### Accepting connections and reading in a receive handler: use `await`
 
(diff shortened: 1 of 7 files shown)


slepp:
Required before merge (D348): PR 3267 owns the spec section 9.1 actor-lifecycle line (`self.stop()`). Drop this PR's duplicate hunk on that line; keep the D10 section and the guide rename.

slepp:
Required before merge, in addition to dropping the section 9.1 hunk 3267 owns: A380 amends D11. A statement-position send or ask whose Result is discarded is an error, `E_SEND_RESULT_DROPPED`, with the fix-it `_ = pid.m()`; there is no must-use lint tier. Replace the 'unbounded-mailbox send stays a warning and a policy-sensitive send stays an error' prose with that rule (spec 5.6 and 9.3).

PR #3272, opened 2026-09-03
fix(types): restrict is to actor handle identity

### LINKED ISSUE #3134: is on enums, machines, and bytes reaches the span-less codegen-front error
## Problem

`a is b` type-checks for enum, machine, and `bytes` operands and then fails in the codegen front with the span-less `E_CODEGEN_FRONT_FAIL_CLOSED: … IdentityCompare lhs must be a pointer or integer value`. The working identity set is narrower than the checker admits - only actor handles, `Vec`, `HashMap`, and `HashSet` compare cleanly. Grounded on the #3133 branch: `enum Colour { Red; Green; }`, a payload enum, `machine Tank`, and `bytes.new()` all reproduce it.

## Expected

The checker and the codegen front agree on one identity set. Every operand shape the codegen front cannot identity-compare is rejected at check time with the `E_IS_VALUE_TYPE` guidance #3133 added for records, derived from the same predicate codegen consults rather than a second list.

## Notes

Extends ledger A351, which authorized the record answer only. The spec's claim that enums are identity-bearing needs the matching correction.

## Outcome

No shape of `is` reachable from user source produces E_CODEGEN_FRONT_FAIL_CLOSED. `is` on any enum value, `indirect` included, or on a machine value is rejected at the use site with E_IS_VALUE_TYPE naming the type and pointing at `==`, the answer #3133 gave records: `indirect` is a layout annotation (HEW-SPEC-2026 §3.7.4) and exposing its heap allocation through `is` would make it a semantic one, so there is no carve-out. `is` on `bytes` type-checks and compiles end to end, matching Vec/HashMap/HashSet.

## Acceptance

- `hew check` on `Colour.Red is Colour.Green`, on a payload enum, on an `indirect` enum and on a two-state machine value each report E_IS_VALUE_TYPE at the `is` expression's span
- `hew run` on `bytes.new() is bytes.new()` compiles and executes, returning a boolean consistent with pointer identity
- hew-cli/tests/is_value_type_reject_e2e.rs gains the new operand shapes plus a control asserting E_CODEGEN_FRONT_FAIL_CLOSED never appears for any `is` expression

slepp:
## Why

I allowed `is` on several heap-backed value types that the backend could not lower, so valid-looking programs reached a spanless compiler invariant. Hew defines `is` as handle identity; copy-on-write and structural values use `==`.

## What

- I make the checker admit actor handles and reject `bytes`, collections, enums, machines, records, trait objects, and closures with `E_IS_VALUE_TYPE`.
- I fail closed when either operand remains unresolved after inference.
- I remove the now-unreachable bytes identity lowering and its MIR support.
- I keep sandbox compilation fail-closed for `is`; actor identity in the sandbox requires a future VM capability rather than another type classifier.

## Test

- `is_allowance` covers the accepted and rejected type categories.
- `is_value_type_reject_e2e` covers user diagnostics.
- Native lowering retains an actor-handle identity control.
- Sandbox coverage proves accepted actor identity stops as a reserved runtime feature before bytecode emission.

Fixes #3134


slepp:
Required before merge (D2, ratified in plans/v060-surface-review.md and ledger D340): the `is` admission set is handle identity only. This PR narrows it for enums and machines but leaves `Vec`, `HashMap`, `HashSet` and `bytes` identity-capable, and D2's table puts collections and `bytes` in the value row: `a is b` on them must be `E_IS_VALUE_TYPE` with the `==` fix-it, exactly like enums. Add the four to the refusal with a reject fixture per type and the negative control that `LocalPid`, `Rc`, `dyn Trait` and closures still pass. The corpus has no `is` on a collection (review file, D2 row), so no migration.

PR #3277, opened 2026-09-03
test(e2e): pin the recv-match ownership limits from #3127

slepp:
## Why

A `match` over a channel or stream `recv()` result silently loses the value
it produces: `match rx.recv() { .Some(s) => s, ... }` returns an empty
string where the sender wrote a real one (#3127). PR #3249 diagnosed the
root cause and proposed a fix in `hew-mir`/`hew-codegen-rs`, but under the
D342 decision that fix is out of proportion for a v0.6.0 patch (a codegen
crash-cleanup rearm plus a match-lowering carrier-owner change, with no
single-root, sub-100-line shape). #3127 stays a documented fail-closed
limit for v0.6.0, tracked for the v0.7.0 P1/P2 ladder work, and this PR
salvages #3249's test coverage against that decision instead of its fix.

## What

- **fixtures**: `channel_recv_match_result_survives.hew` pins the exact
  direct-match shape for both `recv` and `try_recv` against a let-bound
  carrier negative control (binding the carrier to a local first already
  works today). `recv_frame_release_balance.hew` covers the for-await
  drain/early-return/identity-forward/record-payload release shapes plus
  the same direct-match case — the for-await shapes already release
  correctly on the current lowerer; only the fixture's direct-match line
  hits #3127.
- **run.sh**: both fixtures are registered as known-failure ratchets using
  the existing `actor_nested_handle_tuple_transfer` pattern — asserting the
  CURRENT wrong stdout exactly, so the ratchet trips (and must be deleted)
  the moment the real fix lands, and trips differently if the failure mode
  ever changes.
- **LESSONS.md**: records the ratchet pattern and the exact wrong-output
  values, verified by running the built compiler against both fixtures
  before writing the assertions.

Not carried over from #3249: the `hew-mir`/`hew-codegen-rs` fix itself, and
the fixtures were not wired into `scripts/asan-fixture-check.sh` — linking
any `receive gen fn` / `for await` program (including a pre-existing,
currently-passing fixture) against the ASan-instrumented `libhew.a` aborts
on an unrelated coroutine-stack/ASan unmap incompatibility before the
process reaches its own exit path. Noted in `.tmp/TODO.md` as separate
follow-up work, not part of this salvage.

## Test

`make test-vertical-slice` passes, including the two new
`KNOWN channel_recv_match_result_survives` / `KNOWN recv_frame_release_balance`
ratchet lines.


slepp:
Reviewed against the D342 salvage conditions. The file scope is clean - LESSONS.md, the two fixtures with their `.expected`, and `tests/vertical-slice/run.sh`, nothing under `hew-mir/src`, `hew-codegen-rs/src`, `hew-hir/src`, `hew-types/src`, or `hew-runtime/src`. Both new fixtures are registered as run.sh known-failure ratchets citing #3127 (open), run.sh enumerates accept fixtures explicitly rather than globbing, so the ratchets are the only path that runs them, and `Linux gates (shard 1/3)` - the shard carrying `test-vertical-slice` - is green.

Three defects to clear before this can merge.

**1. Session markers on the commit and the PR body**

Problem: commit `916f4af` ends with `Claude-Session: https://claude.ai/code/session_...`, and the PR body ends with the same URL. On a single-commit squash the body defaults to that commit message, so the marker lands on main.

Expected: the footer removed from the commit message and from the PR body. The squash body still needs `Fixes` / issue context for #3127 as the ratchet's citation.

**2. `recv_frame_release_balance.hew` claims a gate that does not run it**

Problem: the fixture header says "this fixture exists for the sanitizer gate: `scripts/asan-fixture-check.sh` runs it under LSan". The fixture appears nowhere in `scripts/asan-fixture-check.sh` on this branch, and the PR body states it was deliberately left out because of the coroutine-stack/ASan link abort. The header is a false statement about which gate exercises the fixture, and it is the kind of claim a later reader will trust when deciding the fixture is covered.

Expected: the header says what actually runs it today - the run.sh ratchet - and records the ASan wiring as blocked, with the blocker's issue number.

**3. The ASan blocker is deferred only to a gitignored file**

Problem: the ASan/coroutine-stack incompatibility is tracked in `.tmp/TODO.md`, which is gitignored, so nothing outside this branch carries it. No open or closed issue mentions the `for await` / ASan `libhew.a` abort.

Expected: an issue filed for the coroutine-stack/ASan unmap abort, cited in the PR body and in the fixture header from defect 2, so the deferred half of the salvage is trackable after this branch is gone.

Not counted against the PR: `Build & test (macOS arm64)` is red, but macOS is informational under the current merge policy and its run log was not yet retrievable to diff its failing IDs against main. Since fixing the defects above rewrites the branch anyway, the next run is the one to judge.

PR #3292, opened 2026-09-03
build: re-cut the ast-grep dialect patch against the parser-cited grammar

### LINKED ISSUE #3283: The ast-grep dialect patch still cites the deleted grammar.ebnf
## Problem

`docs/specs/grammar.ebnf` was deleted at the v0.6.0 surface freeze, and
HEW-SPEC-2026 §12 now names `hew-parser` as the grammar authority with
`tree-sitter-hew/grammar.js` as its mirror. Two comments that survive into
the built ast-grep Hew dialect still cite the deleted file by line number:

```
tools/tree-sitter-hew-consume-parameter.patch:27:     // Import (spec grammar.ebnf:63-68):
tools/tree-sitter-hew-consume-parameter.patch:188:     // YieldExpr (grammar.ebnf:322): "yield" Expr? — the operand is optional, so
```

Both are context lines. The text lives upstream in `tree-sitter-hew`'s
`grammar.js`; the patch reproduces it verbatim so `git apply` finds its
anchor. Rewriting the citations in the patch alone makes the patch fail to
apply and breaks `scripts/build-ast-grep-lang.sh`, so the reference cannot be
fixed from this repo on its own.

## Expected

The two comments name the parser rather than a file that no longer exists.
The rewrite lands in `tree-sitter-hew/grammar.js` first, then
`tools/tree-sitter-hew-consume-parameter.patch` is re-cut against the new
upstream so its context lines match again.

## Notes

The citations on the patch's deletion lines need nothing: the patch removes
those comments from `grammar.js`, so they never reach the built dialect.
Found while landing D19 in the spec (PR #3268), which deleted
`docs/specs/grammar.ebnf` and `docs/specs/Hew.g4`.

slepp:
Why: tree-sitter-hew's grammar.js now cites hew-parser instead of the deleted docs/specs/grammar.ebnf, and the consume-parameter patch reproduced the old citation as unchanged context, so `scripts/build-ast-grep-lang.sh` broke.

What: bump tools/ast-grep.lock to the new tree-sitter-hew commit and re-cut tools/tree-sitter-hew-consume-parameter.patch down to the one dialect gap still needed (the parameter rule's independent `consume`/`var` ordering) — upstream's dotted-path migration already covers what the rest of the old patch used to carry.

Fixes #3283


slepp:
Superseded by #3306, which upstreamed the remaining grammar fix, updated the pinned revision, and deleted this local patch rather than carrying another re-cut.

PR #3349, opened 2026-09-04
fix(stdlib): handle Result and Option values explicitly

slepp:
## Why

Fallible standard-library APIs now return Result or Option, but two behavioural test suites still consumed compression bytes and CSV tables directly.

## What

- Match compression Results before round-trip assertions.
- Match CSV parse Results before accessing Table methods.
- Keep malformed-input, channel-constructor, and template-parse checks on explicit Result branches.

## Test

- `compress_test.hew`
- `csv_test.hew`
- `template_compiled_free_function_p0`
- `affine_resource_carrier_boundaries`


slepp:
CI evidence:

- Compared Linux gates (shard 1/3) from run 33912133974 with main run 33907845901.
- Main stops earlier at `template_compiled_free_function_p0`; this PR reaches `std_panic_wrapper_regex_new_invalid`, where the `regex.new: invalid pattern` diagnostic is absent after status 101.
- Commands: `make hew-fmt-check`; `make test-vertical-slice`.
- Formatter correction: `5ae3066dd`.

Verdict: not a new cause of failure.

slepp:
Focused CI evidence:

- `make hew-native` plus `std_panic_wrapper_regex_new_invalid`: exit `101` with `regex.new: invalid pattern: [`.
- `std_panic_wrappers_success` exits `0`.
- Current `main` gives the same invalid-fixture result through the same compile-and-run path.

The historical missing diagnostic is not reproducible from either revision. No source or oracle change is justified.

Verdict: not a new cause of failure.

slepp:
Superseded by the final-core integration branch's Result-only std convention; 116 of these 184 files were reworked there and the examples exercise the new surface.

PR #3350, opened 2026-09-04
fix(runtime): bind port-only profiler addresses to loopback

slepp:
## Why

Setting `HEW_PPROF` to a port number listened on every interface. A port-only setting should be local unless an address is supplied.

## What

- Bind port-only profiler addresses to loopback.
- Warn when the configured address listens on every interface.

## Test

- `port_only_profiler_address_binds_to_loopback`
- `explicit_wildcard_profiler_address_requests_warning`


slepp:
Integrated into refactor/final-core as 3b6a935b6 (bind port-only profiler addresses to loopback). Closing in favour of the native cutover branch.

PR #3351, opened 2026-09-04
feat(codegen): handle main return values consistently

slepp:
## Why

Native and WASI builds interpreted `main` return values separately. The same program could therefore produce different process exits on different targets.

## What

- Decide during type checking how `main` maps its return value to a process exit.
- Carry that decision through HIR, SIR, and MIR.
- Use it in native and WASI code generation for unit, integer, and Result returns.
- Format Result errors through the selected Display implementation.

## Test

- Native and WASI `main` returning unit
- Native and WASI `main` returning an integer
- Successful and failed Result returns


slepp:
Integrated into refactor/final-core through fe054358f (unified checked process and test entry selection). Closing in favour of the native cutover branch.

PR #3355, opened 2026-09-04
feat(parser): reduce reserved keyword set

slepp:
Why

Retired spellings unnecessarily reserve valid names, and deprecated async function prefixes need a direct migration path.

What

- Make retired and contextual spellings identifier tokens.
- Diagnose deprecated async item prefixes and incompatible stream-loop forms with direct corrections.
- Regenerate the nano grammar from syntax data.

Test

- `keyword_diet_words_are_identifiers`
- `async_fn`, `async_gen_fn`, `for_over_stream`, `for_await_over_vec`, `keyword_diet_identifiers`
- `make grammar-parity`
- `make downstream-check`


slepp:
The keyword diet lands on the final-core integration branch, re-cut against its restructured lexer (async already retired there); this branch's code is stale against it.

PR #3356, opened 2026-09-04
fix(fuzz): ratchet main-less imported helpers

slepp:
## Why
The fuzz oracle recursively executes every accepted Hew source file, so imported library helpers without `main` drift into unexpected rejections.

## What
- Add the two imported equality and hash helper fixtures to the exact rejection ratchet.
- Add a focused oracle self-test proving an importer plus a main-less helper requires manifest membership.

## Test
- `make fuzz-oracle-selftest`


slepp:
Re-cut onto the final-core integration branch as 91e65af33, where the two fixtures already live; the self-test's new case passes there. The branch's pre-existing 2a self-test failure is tracked separately.

PR #3357, opened 2026-09-04
feat(runtime): make node setup and registry typed

slepp:
## Why

Node setup failures were reported outside the type system, configuration was split across mutable setters, and registry lookup could return an actor under the wrong type.

## What

- Replace staged setters with `Node.start(NodeConfig)` and typed configuration, startup, connection, and registration errors.
- Keep registration available before startup and preserve compiler-minted actor identity through local registry entries and distributed gossip.
- Reject wrong-type lookups with `LookupError.TypeMismatch` and migrate the distributed examples to the new surface.
- Retain `RemotePid<A>` as the current return spelling; the final public `RemotePid<A>` to `Pid<A>` unification follows in S-4. This PR does not complete #3256.

## Test

- `make test-node-api`
- `make test-node-api-runtime`
- `make test-node-api NODE_API_TEST=remote_ask_round_trip_returns_exact_values`
- `emitted_node_builtins_are_catalogued_for_wasm_rejection`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -686,8 +686,13 @@ ci-preflight-smoke:
 	$(TEST_RUN_ENV) cargo nextest run --workspace --profile smoke
 
 NODE_API_TEST ?= node_api
+NODE_API_EXAMPLE_INVENTORY = --label "node API" \
+	  --source examples/distributed/kv_server.hew \
+	  --source examples/distributed/kv_client.hew
 test-node-api: hew-native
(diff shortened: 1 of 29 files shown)


slepp:
Superseded by the final-core integration branch, which carries NodeConfig, typed startup and connection errors and typed registry lookup (D453). The distributed examples move to that surface there.

PR #3364, opened 2026-09-08
feat(types): require close(consume self) on resource/opaque types

slepp:
## Why

A `#[resource]` / `#[opaque]` type's `close` ran the implicit scope-exit release a second time when called explicitly, because a borrowing `fn close(self)` left the receiver live. HEW-SPEC-2026 already spells `close(consume self)`.

## What

- Checker rule (D442): `close` on a `#[resource]`/`#[opaque]` type must be `fn close(consume self)`; a borrowing form is a checker error. A borrowing `close` on an unmarked type stays legal.
- Codegen fix: the record-resource release path re-released the record's members after calling a consuming `close`, on the stale assumption that `close` only borrowed — a genuine double-free reproduced with ASan on a `#[resource]` record holding a `string` field. Deleted the redundant release; the callee's own body already destroys what it doesn't move out.
- Migrated every remaining `close(self)` site on a marked type across std tests, hew-cli fixtures, hew-types tests, core-acceptance cases and the core-matrix generator.

## Test

- New unit tests + negative control in `hew-types/src/check/tests/opaque_resources.rs`.
- New core-acceptance case `resource-close-consumes` and a sibling `hew-cli` reject test for the use-after-close consume wall.
- `make core-safety` clean (ASan) for the touched cases, where it previously use-after-freed.
- `make test-core-matrix` re-recorded, counts unchanged (294 pass / 14 crash / 66 nyi-clean / 4 nyi-internal).
- `hew-types` test-strict: 19 known failures before and after (all pre-existing, unrelated).


slepp:
Integrated onto refactor/final-core by cherry-pick (8e5229923); the final-core work lands through the native cutover PR rather than per-lane PRs to main.

PR #3409, opened 2026-09-13
stdlib(semver): enforce SemVer 2.0.0 parsing and caret bounds

gertybotbot:
# stdlib(semver): enforce SemVer 2.0.0 parsing and caret bounds

## Summary

Tightens `std::text::semver` to reject invalid SemVer 2.0.0 inputs, fixes the pre-1.0 caret upper-bound rule, and adds small version increment helpers.

This is a focused stdlib-only patch with four commits and no compiler changes. It is based on upstream `v0.5.6` (`2762fadc`), matching the Hew 0.5.6 compiler used to verify it.

## Current delivery status

**Current follow-up (2026-09-19):** strict grammar validation and unbounded numeric
comparison already exist on main. The remaining pre-1.0 caret-bound defect is
ported to the current API in #3529 and passes the native SemVer suite. The old
integer-limited increment helpers are not carried into the unbounded API.

**Draft as of 2026-09-13.** Current upstream `main` has redesigned the SemVer
module around unbounded string components and a `parse -> Result` API. The
four-patch v0.5.6 series conflicts with that redesign and must not be rebased
mechanically. This PR preserves the reviewed historical series; a successor
must port only the still-missing behaviour to the current API and validate it
with a current Hew compiler.

### Strict `try_parse`

`try_parse` now rejects:

- leading-zero core identifiers (`01.2.3`) and numeric pre-release identifiers (`1.2.3-01`);
- empty pre-release/build tails (`1.2.3-`, `1.2.3+`) and empty dot identifiers;
- non-ASCII pre-release/build identifiers;
- signed, non-numeric, partial, or extra core text (`-1.2.3`, `1.2.3x`, `1.2.3.4`);
- core or numeric pre-release identifiers outside the `i64` range supported by this implementation.

It preserves valid canonical versions, including build numeric identifiers with leading zeroes (`1.2.3-rc.1+build.001`), as SemVer permits those in build metadata.

### Caret ranges

Caret matching now advances the left-most non-zero component:

- `^1.2.3` → `>=1.2.3 <2.0.0`
- `^0.2.3` → `>=0.2.3 <0.3.0`
- `^0.0.3` → `>=0.0.3 <0.0.4`

In particular, `0.3.0` no longer matches `^0.2.3`.

### Overflow hardening

The `Version` core fields are `i64`. `try_parse` now uses checked conversion for core identifiers, rejecting values above `i64::MAX` instead of silently converting them to zero. It also rejects out-of-range numeric pre-release identifiers at parse time; that preserves SemVer §11 numeric precedence without an incorrect overflow comparison. Build identifiers remain strings and are not subject to this numeric limit.

Increment helpers now return `Result<Version, string>` and report a component-specific error at `i64::MAX`, rather than overflowing/trapping.

### Increment helpers

Adds:

```hew
semver.increment_major(version) // Ok(2.0.0), or Err at i64::MAX
semver.increment_minor(version) // Ok(1.3.0), or Err at i64::MAX
semver.increment_patch(version) // Ok(1.2.4), or Err at i64::MAX
```

On success these return new values, discard pre-release/build metadata, and reset lower core components for major/minor increments.

## Tests

Adds 13 `#[test]` cases in `tests/hew/semver_test.hew`, covering every corrected grammar class, valid build metadata, both pre-1.0 caret bounds, reset/drop semantics, overflow parse rejection, and defined increment-at-maximum errors.

Verified with the release compiler:

```bash
cd hew
HEW_STD="$PWD/std" ~/local/hew-0.5.6/bin/hew test tests/hew/semver_test.hew
# 40 passed; 0 failed (27 existing + 13 new)
```

**Delivery verification (2026-09-13):** the canonical four-patch export was
replayed from `2762fadc` in a fresh local clone and reproduced the fork branch
tree exactly. The Hew 0.5.6 binary used for the recorded test run is not
installed on this delivery host, so no new compiler-test claim is made here.

## Compatibility

This deliberately makes `try_parse` fail for inputs that were previously accepted despite violating SemVer 2.0.0. `parse` retains its documented behavior of panicking when `try_parse` fails. Valid SemVer API behavior, pre-release precedence, build-metadata precedence exclusion, existing comparators, and tilde ranges are unchanged.

## Commits

1. `stdlib(semver): validate strict SemVer grammar and add increments`
2. `test(semver): cover strict parsing and caret bounds`
3. `stdlib(semver): harden i64 overflow handling`

## Provenance

The behavior matrix and initial executable prototype were developed in the sealed RFC-022 audit: `projects/hew-stdlib-semver-rfc/patch-candidate/`. This PR ports that proposal into a clean upstream 0.5.6 patch series.

### DIFF
--- a/std/text/semver/semver.hew
+++ b/std/text/semver/semver.hew
@@ -142,6 +142,14 @@ pub fn try_parse(s: string) -> Result<Version, string> {
         || !valid_identifier_list(build, true) {
         return Err(f"semver: invalid version string: {s}");
     }
+    // `string.to_int` intentionally maps malformed and out-of-range input to
+    // zero. SemVer components must retain their exact numeric value, so reject
+    // values outside the module's i64 representation before conversion.
+    if !is_i64_numeric_identifier(major_str)
+        || !is_i64_numeric_identifier(minor_str)
+        || !is_i64_numeric_identifier(patch_str) {
+        return Err(f"semver: invalid version string: {s}");
+    }
     let major = string.to_int(major_str);
     let minor = string.to_int(minor_str);
     let patch = string.to_int(patch_str);
@@ -154,6 +162,14 @@ fn is_strict_numeric_identifier(s: string) -> bool {
         return false;
     }
     s.len() == 1 || !s.starts_with("0")
+}
+
+/// Return true when a numeric identifier fits the i64 representation.
+fn is_i64_numeric_identifier(s: string) -> bool {
+    match string.try_to_int(s) {
+        Ok(_) => true,
+        Err(_) => false,
+    }
 }
 
 /// Return true when an identifier contains only ASCII letters, digits, or `-`.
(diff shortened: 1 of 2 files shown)


slepp:
The strict parsing changes are already on main, and the remaining pre-one caret correction landed in #3529. This older branch is superseded.

Issue #3499, opened 2026-09-18
A supervisor declared in a module is never lowered into a HIR item


slepp:
## Problem

A supervisor declared inside a module is never lowered into a `HirItem`. The
root item loop lowers `Item::Supervisor` (`hew-hir/src/lower.rs:4361`), but both
module item loops list it among the kinds they ignore
(`hew-hir/src/lower.rs:2673` and `:2997`). `collect_supervisor_registry`
(`:30507`) walks module supervisors, but only to gate spawn arguments; it emits
no item.

So `spawn module.Rack` and a bare `spawn Rack` on an imported module supervisor
both reach SIR with a handle whose declaration matches no lowered supervisor,
and lowering falls through to the actor arm:

```
compiler limitation: E_SIR_UNSUPPORTED: semantic lowering of `__hew_fn_main` is
not implemented: local actor handle lacks its exact declaration
```

A module *actor* spawns either way, so this is specific to supervisors.

## Expected

A supervisor declared in a module spawns exactly as one declared at the root or
in a file import does.

## Notes

The checker half is done: #3461's registration change gives a module supervisor
its `{module}.Name` identity in `type_defs` and `supervisor_children`, and
`resolve_spawn_target` admits it through both the bare binding and
`spawn module.Name`. Both spellings now reach the same honest limitation instead
of one reporting "module has no exported actor". A file-imported supervisor
works today because file imports are flattened into `program.items` after
typecheck, so its declaration reaches the root loop.

Until this lands, two modules that each declare a supervisor of the same name
cannot collide in a user-visible way, so the collision half of #3461 has no
behavioural oracle yet.

Issue #3388, opened 2026-09-10
A reserved type name in a machine cascades instead of failing closed once


slepp:
## Problem

```hew
machine tuple {
    events { Toggle, }
    state Closed,
    state Open,
    on Toggle: Closed => Open { .Open }
    on Toggle: Open => Closed { .Closed }
}
fn main() {}
```

Expected exactly one `E_RESERVED_TYPE_NAME`. The checker reports 13 errors: the reserved-name refusal, then a cascade of `ContextVariantNoType` ("expected type `tuple` is not an enum or machine"), `TypeUsedAsValue` and `NonExhaustiveMatch` ("missing _"). The other eight declaration kinds in the same test — `type`, `actor`, `trait`, alias, generic `type` — all stop at one error.

## Expected

A machine whose name is reserved stops after the refusal, like every other declaration kind, so the programmer sees the one thing that is wrong.

## Notes

`hew-types check::tests::basic::reserved_type_names_fail_closed_across_declaration_kinds` is held in the nextest expected-failure ledger against this issue. The test's own transition syntax was stale and has been corrected in the same lane; that removed one confounding error but not the cascade.

Issue #3386, opened 2026-09-10
std.net.http.http_client is gated as the HTTP server on wasm


slepp:
## Problem

Two related failures at the wasm capability boundary.

A program importing `std.net.http.http_client` and calling `http_client.set_timeout` / `request_string` is refused on wasm — but with the server's reason:

```
PlatformLimitation  "HTTP server operations are not supported on WASM32 — the std.net.http server
is backed by native sockets and tiny_http; no cooperative wasm32 server implementation exists yet"
```

The expected row is `http-client` in `wasm-capability-manifest.toml:322-333`, label `std.net.http.http_client operations`. `wasm_native_only_module_feature` (`hew-types/src/check/diagnostics.rs:366-378`) takes the leaf of `canonical_module_import_owner(surface)`, so that owner is resolving to `std.net.http`, leaf `http`.

Worse, `http_client.Response` methods are not gated at all. A program holding a `Response` and calling `status()` and `close()` type-checks on wasm with zero errors, although the same manifest row covers `http_client.Response.*` explicitly.

## Expected

An `http_client` import and every `http_client.Response` method are refused on wasm against the `http-client` capability row, naming the client, not the server.

## Notes

The `Response` half is fail-open, which is the more serious of the two. `wasm_rejects_http_client_request_value_position` passes only because it accepts any `PlatformLimitation`, so it does not catch either. Lead on the root: `canonical_module_import_owner` / `module_import_bindings` at `hew-types/src/check/methods.rs:1732`. `e2e_typecheck http_client_module_helpers_rejected_on_wasm` and `http_client_response_methods_rejected_on_wasm` are held in the nextest expected-failure ledger against this issue.

PR #3500, opened 2026-09-18
fix(types): register a supervisor under its declaration identity

### LINKED ISSUE #3461: A module supervisor is registered under its bare leaf while the resolver mints a qualified identity
## Problem

The checker registers every `supervisor` declaration in the root namespace under
its bare leaf name, even when the declaration belongs to a module
(`hew-types/src/check/registration.rs:2626-2648`: `register_type_namespace_name(None, &sd.name, span)`,
`type_defs.insert(sd.name.clone(), ..)`, `supervisor_children.insert(sd.name.clone(), ..)`).
The resolver mints the supervisor's `DefId` as `{module}.Name`, so one
declaration carries two identities.

A file-imported supervisor is refused at SIR because of it. Given
`tests/hew/imported_same_file_supervisor_support/worker.hew` declaring
`supervisor Inner`, `spawn Inner` in the importing file types as
`Named { name: "Inner", builtin: ActorHandle }`, `hew-sir/src/supervisor.rs::declared_handle`
mints `DefId("Inner")` from that name, and the module's `HirSupervisorDecl.declaration`
is `imported_same_file_supervisor_support.worker.Inner`. Lowering falls through
to the actor arm and fails closed:

```
E_SIR_UNSUPPORTED: semantic lowering of `__hew_fn_imported_supervisor_runs_its_same_file_actor`
is not implemented: local actor handle lacks its exact declaration
```

Two modules that each declare a supervisor of the same name also collide on the
one bare key today.

## Expected

A supervisor registers under `declaration_identity(&sd.name)`, as an actor
already does, so the checker and the resolver name it the same way. A bare
reference resolves through the published bare binding rather than a bare
declaration row, and `spawn Inner` on a file-imported supervisor lowers.

## Notes

The lookup sites that read `supervisor_children` by name move onto the same
identity: `items.rs:149,210`, `calls.rs:1941,2101`, `mod.rs:243,743`,
`resolution.rs:2728,2804,3066`, `expressions.rs:7571,9178`, `methods.rs:2336`.
Root supervisors are unaffected, since their declaration identity is the bare
name; the behaviour change is confined to module and file-imported supervisors.

`spawn module.Supervisor(..)` is refused for a related reason:
`resolve_spawn_target`'s `Expr::FieldAccess` arm filters on `TypeDefKind::Actor`
alone.

Ledger row: `tests/hew/imported_same_file_supervisor_test.hew::imported_supervisor_runs_its_same_file_actor`
in `scripts/hew-suite-expected-failures.txt`.

slepp:
## Why

A supervisor declared in a module or a file import reached no checker table at
all. Only the root item loop wrote `type_defs` and `supervisor_children`, and it
keyed them by the bare leaf; neither import loop had a `Item::Supervisor` arm.
The resolver meanwhile minted `{module}.Name` for the same declaration, so
`spawn Inner` on a file-imported supervisor typed as a bare handle that SIR
could not match to its declaration:

```
E_SIR_UNSUPPORTED: semantic lowering of
`__hew_fn_imported_supervisor_runs_its_same_file_actor` is not implemented:
local actor handle lacks its exact declaration
```

Two modules each declaring a supervisor of the same name would also have shared
the one bare key. (The issue describes this as a bare-leaf registration to
re-key; it is in fact a missing registration path, which is why the change is
larger than a key rename.)

## What

- **checker** (`hew-types/src/check/registration.rs`): the root arm's body is
  extracted as `register_supervisor_decl_as(sd, identity)`, mirroring
  `register_actor_decl_as`, and is now called from three places - the root loop
  under the bare declaration identity, the user-module loop under
  `{module_full_path}.{name}` with the module's export and import bindings
  recorded, and the flat-file loop under `{owner}.{name}` with the bare binding
  published, exactly as the actor arm beside it does. `known_types` gains the
  identity, so a supervisor no longer depends on the `supervisor_children`
  fallbacks to be a resolvable type spelling.
- **checker** (`hew-types/src/check/expressions.rs`): `resolve_spawn_target`'s
  module-qualified arm admits a supervisor as well as an actor - they spawn the
  same way - and its diagnostic says so. The bare arm already resolved both
  through `resolve_bare_spawn_target_identity`.
- A file-imported supervisor must be `pub` to be spawned from the importer, as a
  file-imported actor already must; `tests/hew/imported_same_file_supervisor_support/worker.hew`
  says so.

Root and flat-file supervisors keep their existing identities, which
`hew-hir/tests/declaration_identity.rs` pins.

## Test

`tests/hew/imported_same_file_supervisor_test.hew::imported_supervisor_runs_its_same_file_actor`
passes and its row leaves `scripts/hew-suite-expected-failures.txt`;
`make test-hew-ratchet` exits 0 against the shortened ledger.

`make lint` and `make core-acceptance` (2099 cases) exit 0. The 78 supervisor
and declaration-identity tests across `hew-types`, `hew-hir` and `hew-sir` pass,
including `root_supervisor_name_does_not_false_positive_in_module` and the
`declaration_identity` pins for a root supervisor. The three failures in a full
`hew-types`/`hew-hir`/`hew-sir` run are the ledgered #3388 and #3386 rows.

## What this does not fix

A supervisor declared in a **module** still cannot be spawned: HIR never lowers
one into an item (`hew-hir/src/lower.rs:2673` and `:2997` ignore
`Item::Supervisor`, only the root loop at `:4361` lowers it). Both spellings now
reach the same honest `E_SIR_UNSUPPORTED` limitation rather than one of them
claiming the module exports no such actor. Filed as #3499 with the located
cause. A file import works because its items are flattened into `program.items`
after typecheck and so reach the root loop.

Fixes #3461


slepp:
Pushed `e20e8968f`: four capability-boundary tests pinned the substring `has no exported actor \`Account\``, which the widened spawn-target diagnostic no longer produces verbatim - it now reads `has no exported actor or supervisor \`Account\``, because after this change a module can export a spawnable supervisor and the old wording would be incomplete. The tests follow the message; all four pass locally.

PR #3528, opened 2026-09-19
fix(runtime): drain services after signal shutdown

slepp:
## Why

A service that polls an actor from its main loop could keep submitting requests after SIGTERM, exhaust the shutdown drain budget and remain alive. Separately, closing listener admission also rejected writes on established connections, preventing an accepted TCP handler from finishing during implicit shutdown.

## What

Explicit shutdown closes root and external admission while accepted actor turns can finish peer calls and established I/O. An admission permit covers publication into the scheduler, so drain cannot overlook a producer between its admission check and mailbox handoff. Attachment termination remains completion of accepted work. Implicit shutdown after main returns continues to drain descendants.

## Test

- Linux and macOS source regression at O0 and O2: signal shutdown rejects subsequent root calls, finishes an accepted peer call and closes root and actor resources exactly once.
- Existing TCP drain completion and timeout tests on Linux and macOS, runtime shutdown tests and cross-runtime rejection coverage on Linux.
- Unsuppressed ASan/LSan check for an admitted producer crossing the shutdown boundary; full `make lint`.
- Rebuilt the blog CMS example, verified HTTP POST/GET with Unicode and HTML escaping, then observed a clean exit after SIGTERM.

macOS validation used arm64 macOS 26.6.2, Rust 1.96, LLVM 22.1.8 and the macOS 15.4 SDK. The initial concurrent Nextest run reported a process/stdio leak marker for the admission unit test; an isolated run passed cleanly, and direct process observation saw exit and both output pipes close within 3 ms. No runtime source was changed to suppress the marker.


slepp:
Superseded by #3533, which includes signal-shutdown admission and draining together with the shared consuming-cleanup implementation. Validation and integration continue on that coherent change.

